PR #26, opened 2026-05-10
ci: dependabot and bump gh actions

milosveljkovic:
Some of the actions depend on node.js20, which will be removed from the runner on September 16th, 2026.

Also added:
- CODEOWNERS
- ISSUE_TEMPLATE (should be set since repo is getting a lot of attention)
- dependabot (so you do not think about the gh actions anymore)


zechengz:
Thanks for the contribution.

PR #58, opened 2026-05-16
fix(server): validate Host header on the daemon to defend against DNS rebinding

aaronjmars:
## Summary

The Mirage daemon binds to `127.0.0.1:8765` by default and ships with **no Host-header validation**, **no CORS allowlist**, and **no server-side authentication** (`MIRAGE_TOKEN` is sent client-side as a `Bearer` header but the server never reads it — see [`cli/client.py:48–51`](https://github.com/strukto-ai/mirage/blob/main/python/mirage/cli/client.py#L48-L51) and the lack of any auth middleware in [`server/app.py`](https://github.com/strukto-ai/mirage/blob/main/python/mirage/server/app.py)).

Combined, this makes the daemon reachable via **DNS rebinding** from any web page the user visits: the page resolves an attacker-controlled hostname to `127.0.0.1`, then issues same-origin `fetch()` calls against the daemon's REST API. From there an attacker can drive every privileged endpoint — `POST /v1/workspaces/{id}/execute`, `GET /v1/workspaces/{id}/snapshot`, `POST /v1/workspaces/load`, `DELETE /v1/workspaces/{id}` — against the user's mounted accessors (Slack, Gmail, Discord, GitHub, S3, Postgres, MongoDB, SSH, Redis, Linear, Notion, etc.).

This PR adds Starlette's `TrustedHostMiddleware` to `build_app()` with a loopback-only default allowlist and an env-var / kwarg escape hatch for reverse-proxy deployments.

## Impact

When a user is running the daemon (`mirage workspace --create CONFIG.yaml` auto-spawns it) and visits a page the attacker controls:

1. The page serves JS that periodically issues `fetch('http://attacker.example:8765/...')`.
2. The attacker's authoritative DNS for `attacker.example` flips from a real IP to `127.0.0.1` (TTL=0 rebinding).
3. From the browser's perspective the request is same-origin (origin = `attacker.example`), so the response body is readable.
4. The page can now:
   - `GET /v1/workspaces` → enumerate every mounted workspace.
   - `GET /v1/workspaces/{id}/snapshot` → download a full tar of workspace state, including any tokens stashed in the workspace.
   - `POST /v1/workspaces/{id}/execute` → run arbitrary shell commands across all mounted accessors (read/write Gmail, post to Slack, exfil S3, query Postgres, SSH operations, etc.).

The daemon does not need to be exposed beyond loopback for this to succeed — DNS rebinding turns a loopback service into a browser-reachable service.

## Location

- `python/mirage/server/app.py` — `build_app()` had no middleware installed.
- Daemon spawn defaults: [`cli/settings.py:20`](https://github.com/strukto-ai/mirage/blob/main/python/mirage/cli/settings.py#L20) (`http://127.0.0.1:8765`) + [`cli/client.py:105–116`](https://github.com/strukto-ai/mirage/blob/main/python/mirage/cli/client.py#L105-L116) (`uvicorn --host 127.0.0.1`).

## Fix

- New `python/mirage/server/host_validation.py` — small helper that resolves the allowed-host list from an explicit kwarg, then `MIRAGE_ALLOWED_HOSTS` (CSV), then a hard-coded loopback default `("127.0.0.1", "localhost", "::1")`.
- `build_app()` gains an `allowed_hosts: list[str] | None = None` kwarg. Unless the resolved list contains `"*"`, the app installs `starlette.middleware.trustedhost.TrustedHostMiddleware`.
- `["*"]` is the documented opt-out for reverse-proxy deployments.

`TrustedHostMiddleware` strips port before comparison, so `127.0.0.1:8765` matches `127.0.0.1` automatically — verified by `test_explicit_wildcard_disables_enforcement` and an in-process port-stripping check.

## Tests

- `python/tests/server/test_host_validation.py` — 9 new tests covering CSV parsing, env precedence, default rejection of unknown hosts, default acceptance of loopback hosts, env-extended allowlist, and explicit `["*"]` disable.
- `python/tests/server/conftest.py` — autouse fixture that sets `MIRAGE_ALLOWED_HOSTS=*` for the existing server-test directory (which drives the app via `httpx.ASGITransport(base_url="http://test")`). New host-validation tests opt out of the override with the `@pytest.mark.no_host_override` marker registered in `pyproject.toml`.

All assertions verified locally via direct module load (`smoke-test.js`) and via an in-process FastAPI integration harness exercising the middleware against `attacker.example`, `127.0.0.1`, `localhost`, `daemon.mirage.local` (env-extended), and `127.0.0.1:8765` (port-stripping). I couldn't run the full pytest suite locally — Mirage's full dep tree (`jq`, `mfusepy`, `pypdfium2`, `tree-sitter-bash`, etc.) needs the `uv sync --all-extras --no-extra camel` setup from `CLAUDE.md`, and `uv` isn't in this scanning sandbox. CI will exercise the full suite.

## Detected by

Aeon + semgrep p/security-audit, p/owasp-top-ten, p/secrets + trufflehog (fs + git history) + osv-scanner + manual review of `server/`, `cli/`, and every router.

- CWE-346 (Origin Validation Error) — primary
- CWE-350 (Reliance on Reverse DNS Resolution for a Security-Critical Action) — secondary, since the attack relies on the daemon trusting any client that can reach `127.0.0.1`
- Severity: **high** — pre-auth, no user click required beyond visiting a page; the daemon must be running, which is the normal CLI flow

## Out of scope / separate follow-ups noticed during review

(Flagging here so the team has the context; not bundled into this PR to keep the diff narrow and single-concern.)

1. **`MIRAGE_AUTH_TOKEN` is never enforced on the server.** [`cli/client.py:103–104`](https://github.com/strukto-ai/mirage/blob/main/python/mirage/cli/client.py#L103-L104) propagates the token into the spawned daemon's env, and [`cli/client.py:48–51`](https://github.com/strukto-ai/mirage/blob/main/python/mirage/cli/client.py#L48-L51) sends `Authorization: Bearer …` on every request, but nothing on the server side reads either the env var or the header. The design intent is clearly there — wiring this up would be a clean defense-in-depth complement to this PR (Host validation defends the cross-origin browser path; bearer auth defends the same-host malicious-process path).
2. **Dependency CVEs in `python/uv.lock`** — osv-scanner flagged 13 packages with advisories, including litellm (1 CRITICAL + 3 HIGH, but Proxy-mode only — likely out-of-scope if Mirage uses litellm SDK-only), urllib3 2.6.3 (2× HIGH), python-multipart 0.0.22 (1× HIGH), pydantic-ai-slim (2× HIGH), langsmith (1× HIGH), langchain-core (1× HIGH), aiohttp 3.13.3, cryptography 46.0.5, basic-ftp, fast-uri (npm-side), protobufjs (npm-side). Worth a separate lockfile-only PR.

The `npm` side of the monorepo (`typescript/`) wasn't deeply reviewed for the same DNS-rebinding pattern — happy to scan and file separately if there's interest.

## Verification commands

```bash
# Per CLAUDE.md PR flow
./python/.venv/bin/pre-commit run --all-files
cd python && uv run pytest tests/server/test_host_validation.py -v
cd python && uv run pytest tests/server/  # full server suite under autouse override
```

---

Filed by [Aeon](https://github.com/aaronjmars/aeon-aaron). Happy to iterate on the API surface (env var name, default allowlist, opt-out sentinel) if you'd prefer a different shape. The diff is intentionally narrow and additive — single concern, single new module, single `build_app()` kwarg.

### DIFF
--- a/python/mirage/server/app.py
+++ b/python/mirage/server/app.py
@@ -21,9 +21,9 @@ from contextlib import asynccontextmanager
 from pathlib import Path
 
 from fastapi import FastAPI
-from starlette.middleware.trustedhost import TrustedHostMiddleware
 
-from mirage.server.host_validation import resolve_allowed_hosts
+from mirage.server.host_validation import (HostHeaderMiddleware,
+                                           resolve_allowed_hosts)
 from mirage.server.jobs import JobTable
 from mirage.server.persist import restore_all, snapshot_all
 from mirage.server.registry import WorkspaceRegistry
@@ -125,7 +125,7 @@ def build_app(idle_grace_seconds: float = 30.0,
     app = FastAPI(title="Mirage daemon", version="0.1", lifespan=_lifespan)
     hosts = resolve_allowed_hosts(allowed_hosts)
     if "*" not in hosts:
-        app.add_middleware(TrustedHostMiddleware, allowed_hosts=hosts)
+        app.add_middleware(HostHeaderMiddleware, allowed_hosts=hosts)
     app.state.allowed_hosts = hosts
     app.state.started_at = time.time()
     app.state.exit_event = exit_event or asyncio.Event()
--- a/python/mirage/server/app.py
+++ b/python/mirage/server/app.py
@@ -21,7 +21,9 @@ from contextlib import asynccontextmanager
 from pathlib import Path
 
 from fastapi import FastAPI
+from starlette.middleware.trustedhost import TrustedHostMiddleware
 
+from mirage.server.host_validation import resolve_allowed_hosts
 from mirage.server.jobs import JobTable
 from mirage.server.persist import restore_all, snapshot_all
 from mirage.server.registry import WorkspaceRegistry
@@ -91,7 +93,8 @@ async def _lifespan(app: FastAPI):
 
 def build_app(idle_grace_seconds: float = 30.0,
               exit_event: asyncio.Event | None = None,
-              persist_dir: str | Path | None = None) -> FastAPI:
+              persist_dir: str | Path | None = None,
+              allowed_hosts: list[str] | None = None) -> FastAPI:
     """Construct a daemon FastAPI app.
 
     The workspace registry is created eagerly so the app is usable
@@ -109,11 +112,21 @@ def build_app(idle_grace_seconds: float = 30.0,
         persist_dir (str | Path | None): directory for snapshot /
             restore. When set, the daemon snapshots every workspace
             on graceful shutdown and rehydrates them on next start.
+        allowed_hosts (list[str] | None): host allowlist for the
+            ``Host`` header. ``None`` (default) reads
+            ``$MIRAGE_ALLOWED_HOSTS`` (CSV) or falls back to
+            loopback-only (``127.0.0.1``, ``localhost``, ``::1``).
+            Pass ``["*"]`` to disable enforcement (only safe behind
+            a trusted reverse proxy).
 
     Returns:
         FastAPI: configured app with all routers mounted.
     """
     app = FastAPI(title="Mirage daemon", version="0.1", lifespan=_lifespan)
+    hosts = resolve_allowed_hosts(allowed_hosts)
+    if "*" not in hosts:
+        app.add_middleware(TrustedHostMiddleware, allowed_hosts=hosts)
+    app.state.allowed_hosts = hosts
     app.state.started_at = time.time()
     app.state.exit_event = exit_event or asyncio.Event()
     app.state.registry = WorkspaceRegistry(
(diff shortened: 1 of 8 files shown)


zechengz:

Updated with

1. Log on rejection. Custom HostHeaderMiddleware replaces TrustedHostMiddleware so we get a logger.warning(...) when a request is dropped. Same 400 + port stripping otherwise.
2. Narrowed conftest. MIRAGE_ALLOWED_HOSTS=test,127.0.0.1,localhost,::1 instead of "*", so the middleware stays active during the existing server tests (they all use base_url="http://test").
3. TS mirror. Ported the validation to typescript/packages/server (Fastify onRequest hook + module + 12 tests). End-to-end tested against the cross-resource example for both daemons.

zechengz (approved):

Issue #59, opened 2026-05-17
[Issue]: add key_prefix to S3Config for bucket subpath scoping


Goktug:
### What type of report is this?

Feature request

### Area

TypeScript packages

### Summary

I'd like to propose adding an opt-in `key_prefix` field to `S3Config` so an `S3Resource` can be scoped to a bucket subpath. The prefix would be applied transparently to operations and stripped from results, keeping agent-facing paths clean.
 
A common pattern is keeping multiple logical datasets in one bucket, separated by key prefix (`s3://app-data/scope-a/...`, `s3://app-data/scope-b/...`). Today, mounting one of these requires either subclassing `S3Resource` to rewrite paths (which depends on internal API stability at v0.0.1), one bucket per scope (limited by AWS bucket quotas), or living with the prefix in every agent-visible path.
 
A native `key_prefix` solves all three. Backwards compatible: when unset, behavior is byte-identical to today. Happy to do both Python and TypeScript for parity if accepted. Filing this as RFC before sending the PR.

### Steps to reproduce

This is a feature request, not a bug, so there isn't a literal reproduction. To demonstrate the friction, here's the current vs. desired behavior:
 
**Current behavior — no `key_prefix`:**
 
```python
s3 = S3Resource(S3Config(bucket='app-data'))
ws = Workspace({'/data': s3})
 
# Agent must write paths like:
await ws.execute("cat /data/scope-a/notes/foo.md")
# Mount path leaks the scope prefix into every agent-visible path.
```
 
**Desired behavior — opt-in `key_prefix`:**
 
```python
s3 = S3Resource(S3Config(bucket='app-data', key_prefix='scope-a/'))
ws = Workspace({'/data': s3})
 
# Agent sees clean paths:
await ws.execute("cat /data/notes/foo.md")
# Under the hood, S3 GET is called with Key='scope-a/notes/foo.md'.
# List operations strip 'scope-a/' from results before returning.
```

### Expected behavior

`S3Config` accepts an opt-in `key_prefix: str | None = None` (Python) / `keyPrefix?: string` (TypeScript). When set:
 
1. The accessor transparently prepends `key_prefix` to bucket keys in every operation: `read_bytes`, `write_bytes`, `readdir`, `stat`, `copy`, `rename`, `unlink`, `rmdir`, `range_read`, `read_stream`, `find_flat`, `truncate`, `create`, `exists`, `du`, `du_all`, `mkdir`.
[...]

PR #60, opened 2026-05-17
feat(s3): add key_prefix for multi-tenant resource scoping

Goktug:
## Summary

Adds an opt-in `key_prefix` (Python) / `keyPrefix` (TypeScript) config field to `S3Resource` so multi-tenant systems can transparently scope each tenant to a bucket subpath without leaking the prefix into agent-visible paths.

Refs #59

## Motivation

Building a multi-tenant product on Mirage today requires either:

- one bucket per tenant (hits AWS bucket limits at scale — 100 default, 1000 hard cap), or
- each tenant gets full bucket access via STS, but then the agent has to know about its own tenant ID in every path.

Neither is ideal. A native `key_prefix` lets the agent see clean paths like `/data/notes.md` while operations under the hood touch `s3://bucket/scope-a/notes.md`. STS policies still enforce isolation at the AWS layer.

## Changes

- `S3Config` gains `key_prefix: str | None` (Python) / `keyPrefix?: string` (TypeScript).
- Normalization at config construction: leading slashes stripped, trailing slash enforced, empty string treated as unset.
- Field threaded through the existing `_key` / `_prefix` (Python) and `s3Key` / `s3Prefix` (TypeScript) utility helpers. `S3Accessor` stays a thin config holder; no architectural refactor.
- `readdir` and `find` strip the prefix from S3-returned keys before building user-facing entries, preventing leakage into the agent's view.
- `get_state` / `getState` includes `key_prefix` (identity, not credential — intentionally not in `redacted_fields`).
- TypeScript YAML configs accept the snake-case `key_prefix` alias, auto-mapped to `keyPrefix`.
- Tests for every op (read, write, readdir, glob, stat, exists, copy, rename, unlink, get_state) with prefix set and unset, plus normalization tests.
- Docs updated in `docs/python/resource/s3.mdx`, `docs/typescript/setup/s3.mdx`, and `docs/home/setup/s3.mdx`.

## Browser variant

`mirage-browser`'s `S3Resource` (presigned-URL flow) inherits the field through the shared core config. The docs add a security note that for browser flows, prefix scoping must be enforced server-side in the `presignedUrlProvider` implementation and the backing IAM/STS session policy — the client-side `keyPrefix` is ergonomics only, not a security boundary.

## Backwards compatibility

`key_prefix` defaults to `None` / `undefined`. With it unset, behavior is byte-identical to before this PR. All existing tests pass without modification.

## Security note

`key_prefix` is a path-scoping convenience, not a security boundary. For real isolation, pair with STS AssumeRole session policies that constrain the AWS-side credentials to the same prefix. The docs section explains this pairing.

## Test plan

- [x] Python: `cd python && uv run pytest tests/core/s3` — 22 passed
- [x] TypeScript: `pnpm --filter @struktoai/mirage-node test -- --run src/resource/s3` — 1286 passed
- [x] TypeScript: `pnpm typecheck` — clean across all packages
- [x] Lint/format: `pre-commit run --all-files` — clean
- [ ] Maintainer review

### DIFF
--- a/docs/home/setup/s3.mdx
+++ b/docs/home/setup/s3.mdx
@@ -31,3 +31,29 @@ AWS_SECRET_ACCESS_KEY=wJal...
 If you have `~/.aws/credentials` configured, you can use a profile name instead of explicit access keys.
 
 For the Python resource API, see the [S3 resource doc](/python/resource/s3).
+
+## Scoping a resource to a key prefix
+
+Both runtimes support a key prefix option that transparently scopes every operation to a subpath of the bucket, so agents see clean paths while the underlying S3 keys carry the full prefix.
+
+<CodeGroup>
+
+```ts TypeScript
+const s3 = new S3Resource({
+  bucket: 'app-data',
+  region: 'eu-west-1',
+  keyPrefix: `users/${userId}/`,
+})
+```
+
+```python Python
+S3Resource(S3Config(
+    bucket="app-data",
+    region="eu-west-1",
+    key_prefix=f"users/{user_id}/",
+))
+```
+
+</CodeGroup>
+
+Leading slashes are stripped and a trailing slash is added automatically. `None` / `undefined` and empty strings both mean "no prefix." For browser security considerations, see the [TypeScript S3 setup doc](/typescript/setup/s3).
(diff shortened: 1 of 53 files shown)


zechengz:

zechengz (approved):

zechengz:
Some additional changes:

1. Fixed two rg bugs — Python and TS rg double-prepended the mount prefix to results (/deep/deep/foo), and TS also had a cache bug producing /deep//foo. Both fixed.
2. Cleaned up the prefix code — moved S3Config to accessor/s3.py to kill a circular import, dropped redundant normalization in _client.py, and extracted the four prefix functions (normalize/apply/apply_dir/strip) into shared utils/key_prefix.{py,ts} for future backends to reuse.
3. Expanded the examples — added a /deep mount with multi-segment key_prefix="subdata/subsubdata/" to s3.py, s3_vfs.py, s3_fuse.py, and the TS equivalents (also switched TS examples from MinIO to the real bucket so they run without Docker).

PR #70, opened 2026-05-20
refactor(commands): consolidate cloud backend wrappers (s3/gdrive/google workspaces)

zechengz:
## Summary

Apply the same thin-wrapper + generic-dispatch pattern that landed for `ram`/`disk`/`redis` in #68. Each cloud backend wrapper now does just three things:

1. **Glob-resolve paths** via its own `core/<backend>/glob.py`
2. **Inject backend-specific callables** (`read_bytes`, `read_stream`, `readdir_fn`, `stat_fn`, `mkdir_fn`) into the generic
3. **Pass through** stdin/options unchanged

**Stats:** 87 files changed, +984 / −3,471 (net **−2,487 lines**).

## What got refactored

| Backend | Wrappers | Notes |
|---|---|---|
| s3 | 34 | Full set of single-file flat wrappers |
| gdrive | 33 | Same, plus `read as read_bytes` alias so generic commands see `.gdoc.json` / `.gsheet.json` / `.gslide.json` as bytes |
| gdocs / gsheets / gslides / gmail | 4 each | basename, dirname, realpath, tree |

## What stayed resource-specific

Per the "use resource-specific when generic doesn't fit" guidance:

- **s3/du, s3/find**: custom S3 listing semantics (paginator, signed-prefix iteration)
- **gdrive/du, gdrive/find, gdrive/sed**: no `core/gdrive/du`, `find`, or `write` module
- **gdocs/gsheets/gslides/gmail jq, nl, find**: no `core/.../stream` or `core/.../find` module
- **Folder-based commands** (`cat/`, `head/`, `tail/`, `ls/`, `stat/`, `file/`, `cut/`, `grep/`, `wc/`): dispatch over format variants (parquet/feather/orc/hdf5/etc.)
- **`gws_*` domain-specific Google commands**: no generic equivalent

## Streaming preserved end-to-end

Verified via a side-by-side test on a 10.4 MB JSONL object:

| Pipeline | Bytes downloaded | Time |
|---|---|---|
| `head -n 1` | 8 KB | 0.21s |
| `cat \| head -n 1` (2-stage) | 8 KB | 0.18s |
| `cat \| tr \| grep \| head -n 1` (4-stage) | 8 KB | 0.70s |
| `cat \| tr \| grep \| head -n 1 \| wc -l` (5-stage) | 8 KB | 0.22s |
| `cat \| wc -l` (non-cancellable) | 10.4 MB | 1.50s |

Chunked S3 GetObject reads + backpressure cancellation propagate through arbitrarily long pipes. 1273× data reduction on early-exit pipes — identical to pre-refactor behavior.

## Bug fixes in cloud examples (pre-existing, unrelated to refactor)

While verifying the examples I found two pre-existing bugs:

1. **`examples/python/s3/s3.py`** — `Workspace.load` was supplying only `/s3/` override when the workspace had both `/s3/` and `/deep/` mounts. The snapshot correctly redacted creds for both; `Workspace.load` correctly refused. Fixed by passing both overrides.

2. **`examples/python/gcs/gcs.py`** — called `await ws.reset()` which has never existed on `Workspace` (file unchanged since initial public release `924ec49`). The surrounding code (cache.get() assertions, max_drain_bytes manipulation) makes it clear the intent was cache clearing. Fixed: `ws.reset()` → `ws.cache.clear()`.

Both bugs verified pre-existing by `git stash` + re-running on clean main.

## Other

- `gdrive/__init__.py`: renamed `sort_cmd` import to `sort` for consistency with other backends.
- `.gitignore`: added `docs/learnings/` alongside `docs/plans/`.
- Extended `examples/python/gcs/gcs.py` with streaming + multi-stage pipe backpressure demonstrations.

## Test plan

- [x] `cd python && uv run pytest --ignore=tests/agents` — 6203 passed, 266 skipped
- [x] `pre-commit run --files <staged>` — all hooks pass
- [x] `examples/python/s3/s3.py` — exit 0 (cross-mount, redaction, persistence, drift)
- [x] `examples/python/gdrive/gdrive.py` — exit 0 (`.gdoc.json` handling)
- [x] `examples/python/gdocs/gdocs.py`, `gsheets`, `gslides`, `gmail` — all exit 0
- [x] `examples/python/cross/example.py` + `load_check.py` — exit 0 (6/6 replay match)
- [x] `examples/python/r2/r2.py`, `oci/oci.py`, `supabase/supabase.py` — all exit 0
- [x] `examples/python/gcs/gcs.py` — exit 0 (streaming + chain backpressure verified)


python/mirage/commands/builtin/gdrive/tree.py
```diff
@@ -101,23 +37,18 @@ async def tree(
...
-    warnings: list[str] = []
-    results = await _tree_async(
-        accessor,
-        p0,
-        max_depth=max_depth,
+    paths = await resolve_glob(accessor, paths, index)
+    return await generic_tree(
+        paths[0],
+        readdir=partial(readdir, accessor),
+        stat=partial(stat, accessor),
```
sonhmai, line 47: Will this cause an error in executing `mirage.core.gdrive.stat.stat`? 
- this passes stat=partial(stat, accessor) into generic_tree()
- generic tree calls s = await stat(entry_spec) https://github.com/strukto-ai/mirage/blob/a78c24b3d04d892896759d2be7aacb6c86976ad2/python/mirage/commands/builtin/generic/tree.py#L43-L45
- mirage.core.gdrive.stat.stat() runs with index=None and raise `FileNotFoundError(path)` https://github.com/strukto-ai/mirage/blob/a78c24b3d04d892896759d2be7aacb6c86976ad2/python/mirage/core/gdrive/stat.py#L38-L39

python/mirage/commands/builtin/gdrive/realpath.py
```diff
@@ -44,15 +33,13 @@ async def realpath(
...
-        if e and not await _exists(accessor, resolved, _extra.get("index"),
-                                   prefix):
-            raise FileNotFoundError(
-                f"realpath: '{p.original}': No such file or directory")
-        lines.append(resolved)
-    return ("\n".join(lines) + "\n").encode(), IOResult()
+    if paths:
+        paths = await resolve_glob(accessor, paths, index)
+    return await generic_realpath(paths or [],
+                                  stat_fn=stat_impl,
```
sonhmai: Will this have the same problem as in `tree` mentioned in another comment as it seemingly does not pass the index as param to `stat_impl`? 

Ref old code: https://github.com/strukto-ai/mirage/pull/70/changes#diff-654ede6e1405cf437cac2eff280d28e9656be5c17b14c6f5eae0191543d016c9L32

python/mirage/commands/builtin/gdrive/jq.py
```diff
@@ -34,37 +68,19 @@ async def jq(
...
-        spread = "[]" in expression
-        return format_jq_output(result, r, c, spread), IOResult()
-    raise ValueError("jq: missing input")
+        paths = await resolve_glob(accessor, paths, index)
+    else:
+        paths = []
+    return await generic_jq(paths,
+                            *texts,
+                            read_bytes=read_bytes,
+                            read_stream=_stream,
```
sonhmai: Will these two lose passing index as a param too?

Old code `raw = await gdrive_read(accessor, p, _extra.get("index"))` and index is not passed here

sonhmai: The same in other commands using `read_bytes and read_stream`: awk, md5, sort, etc.

sonhmai:
hey @zechengz, this cleaned up a lot of dups which is good! I left some comments on possible missing index param passing.

python/mirage/commands/builtin/gdrive/realpath.py
zechengz: Thanks for spotting this :)! Will fix that!

PR #72, opened 2026-05-20
[codex] Report read-only resource commands

### LINKED ISSUE #28: Write-flagged resource commands return "command not found" on READ-mode mounts (vs filesystem writes which return a clear "read-only mount" error)
## Summary

Resource-specific write commands (e.g. `slack-post-message`, `slack-reply-to-thread`) are listed by `man <cmd>` and `man` on a READ-mode mount, but invoking them fails with `<cmd>: command not found` instead of a clear "this command requires WRITE mode" error.

Contrast: filesystem-level write commands (`rm`, `mv`, `cp` dst) on a READ-mode mount return the correct `cmd: read-only mount at <prefix>/` error.

## Environment

- `@struktoai/mirage-cli` `0.0.1`
- `@struktoai/mirage-server` `0.0.1`
- bun `1.3.9`, macOS arm64

## Repro

Encountered first with `/slack` in READ mode:

```bash
# /slack mounted with `mode: READ`
mirage execute -w demo -c 'man slack-post-message'
# → renders full help: "# slack-post-message ... OPTIONS: --channel_id, --text"

mirage execute -w demo -c 'slack-post-message --channel_id CXXX --text hi'
# →
# {
#   "exitCode": 127,
#   "stderr": "slack-post-message: command not found"
# }
```

After flipping `/slack` to `mode: WRITE`, `slack-post-message` works.

For comparison, filesystem-level writes on a READ-mode mount are surfaced cleanly:

```yaml
mounts:
  /disk:
    resource: disk
    mode: READ
    config:
      root: /tmp/some-dir
```

```bash
mirage execute -w repro -c 'rm /disk/x.txt'
# →
# {
#   "exitCode": 1,
#   "stderr": "rm: read-only mount at /disk/"
# }
```

## Expected

Resource-specific write commands should fail the same way filesystem writes do, with a clear "read-only mount" error pointing at the offending prefix and what mode it needs. The current `command not found` message:

1. Is technically incorrect — `man` knows the command exists
2. Sends users hunting for typos / version mismatches / missing extras instead of the actual fix (flip the mount mode)
3. Inconsistent with the filesystem-write error path

## Hypothesis

`@command(..., write=true)` commands look to be registered into the per-mount registry conditionally on mount mode (so they're absent from the exec dispatch table on READ mounts, hence "command not found"), while `man` reads from a different / unfiltered source. Filesystem-level writes get registered unconditionally and check mount mode at exec time, producing the better error. Aligning the two paths would close the inconsistency.

## Severity

UX. Cost us ~5 minutes today and one accidentally-misformatted Slack message because the error didn't point at the right thing.

puneetdixit200:
## Summary
- allow write-flagged resource commands on READ mounts to resolve to their owning mount
- preserve the existing mount-level read-only error instead of returning command not found
- update the mount registry regression test to cover no-path write command fallback

Fixes #28.

## Validation
- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm exec vitest run src/workspace/mount/registry.test.ts`
- `pnpm exec vitest run src/workspace/mount/registry.test.ts src/workspace/mount/mount.test.ts src/workspace/executor/command.test.ts`
- `pnpm exec tsc --noEmit`
- `pnpm exec prettier --check src/workspace/mount/registry.ts src/workspace/mount/registry.test.ts`
- `git diff --check`

Note: `pnpm install --frozen-lockfile` without `--ignore-scripts` failed locally because `@zkochan/fuse-native` needs system FUSE headers (`fuse.h`), which are not present on this machine. The changed package tests do not require native FUSE.

## AI Assistance
This PR was prepared with AI assistance. I reviewed the code changes and validated them with the commands listed above.

### DIFF
--- a/typescript/packages/core/src/workspace/mount/registry.test.ts
+++ b/typescript/packages/core/src/workspace/mount/registry.test.ts
@@ -271,7 +271,7 @@ describe('MountRegistry.resolveMount: cross-mount fallback', () => {
     expect(mount).toBe(b)
   })
 
-  it('skips fallback mount when cmd is write-only and target mount is READ', async () => {
+  it('resolves write command on READ fallback mount so execution reports read-only', async () => {
     const reg = new MountRegistry(
       { '/a': new RAMStubResource(), '/b': new RAMStubResource() },
       MountMode.READ,
@@ -288,6 +288,12 @@ describe('MountRegistry.resolveMount: cross-mount fallback', () => {
     if (writeCmd === undefined) throw new Error('missing write cmd')
     b.register(writeCmd)
     const mount = await reg.resolveMount('mutate', [], '/a/x')
-    expect(mount).toBeNull()
+    expect(mount).toBe(b)
+    if (mount === null) throw new Error('missing mutate mount')
+    const [, io] = await mount.executeCmd('mutate', [], [], {})
+    expect(io.exitCode).toBe(1)
+    expect(new TextDecoder().decode(io.stderr as Uint8Array)).toContain(
+      'mutate: read-only mount at /b/',
+    )
   })
 })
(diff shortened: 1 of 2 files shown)


zechengz (approved):
Thanks! :)

PR #104, opened 2026-05-27
fix(node): enforce mount mode for FUSE buffered writes

Aleksion:
## Summary
- route Mirage FUSE buffered writes through Workspace.dispatch instead of Workspace.fs.writeFile
- keep create, direct write, truncate fallback, and flush on the same mount-mode/invalidation path
- add a regression test proving READ-mode FUSE create and buffered flush both fail with EACCES and leave the backing file unchanged

## Why
A hosted Daytona FUSE proof exposed inconsistent READ-mode behavior: MirageFS.create() returned EACCES because it used Workspace.dispatch(), but existing-file overwrite through write()+flush() could still succeed because flush used Workspace.fs.writeFile() directly. This patch makes FUSE writes consistently honor mount mode.

## Test
- pnpm --dir packages/node exec vitest run src/fuse/fs.test.ts

### DIFF
--- a/examples/python/ram/ram_fuse_readonly.py
+++ b/examples/python/ram/ram_fuse_readonly.py
@@ -1,0 +1,60 @@
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


zechengz (approved):
Add some examples. Thanks for the contribution! :)

PR #93, opened 2026-05-25
feat(databricks-volume): add file write

sonhmai:
## Summary
Adds file-level write support for Databricks Volume resources.
Continuing the existing read-only implementation #62. Part of #61

- Adds Databricks Volume core operations for `write`, `create`, and `unlink`, using `PathSpec` paths, `backend_path()` mapping,  parent-directory cache invalidation after mutations.
- Adds Databricks Volume `touch` and file-only `rm` commands for `Workspace.execute`.

Out-scope (will do in future PRs): 
- directory ops e.g. mutation, recursive delete, copy, rename, du, etc.

## Test plan
- [x] `cd python && uv run pytest tests/core/databricks_volume tests/resource/databricks_volume`
- [x] actual integration test run with my databricks volumes

<details>
  <summary>Test Result. Click to expand!</summary>
  
  ```
databricks fs ls dbfs:/Volumes/workspace/default/volume_test --profile default
debug_output.json
debug_output_5_lines.jsonl
pr93_file3.txt

../python/.venv/bin/python test_pr_93.py

============================
STEP 1: Create and write files (Happy Path)

============================
COMMAND: touch /dbx/pr93_file1.txt
EXPECTED RESULT: SUCCESS
EXIT CODE: 0

============================
COMMAND: touch /dbx/pr93_file2.txt
EXPECTED RESULT: SUCCESS
EXIT CODE: 0

============================
COMMAND: touch /dbx/pr93_file3.txt
EXPECTED RESULT: SUCCESS
EXIT CODE: 0

============================
STEP 2: Delete 2 files (file1 and file2)

============================
COMMAND: rm /dbx/pr93_file1.txt
EXPECTED RESULT: SUCCESS
EXIT CODE: 0

============================
COMMAND: rm /dbx/pr93_file2.txt
EXPECTED RESULT: SUCCESS
EXIT CODE: 0

============================
STEP 3: Read operations on leftovers (expect SUCCESS)

============================
COMMAND: ls /dbx/
EXPECTED RESULT: SUCCESS
EXIT CODE: 0
STDOUT:
codex-smoke-20260518
debug_output.json
debug_output_5_lines.jsonl
pr93_file3.txt

============================
COMMAND: find /dbx/ -type f
EXPECTED RESULT: SUCCESS
EXIT CODE: 0
STDOUT:
/dbx/codex-smoke-20260518/nested/alpha/info.txt
/dbx/codex-smoke-20260518/nested/beta/deeper/data.json
[...]

PR #66, opened 2026-05-18
feat(py/databricks): add read-only volume resource docs and examples

sonhmai:
## Summary

Add a read-only Python support for `DatabricksVolumeResource` for Databricks Unity Catalog volumes.

This branch adds the resource/config/registry wiring, SDK-backed auth config (`host`, `token`, `profile`, `timeout`), read-only file ops (`readdir`, `stat`, `exists`, `read_bytes`, `read_stream`, `range_read`, glob), `root_path` handling, token redaction, expanded tests, and docs/examples.

Part of #61. Will add write support and higher-level filesystem ops in follow-up PRs.

Use case: stated in #61, basically projects (mine included), companies building agents and apps on Databricks Apps runtime (their abstration ~ EKS) can use this with Databricks Volume (their storage abstraction for files).

### Supporting

new resource
- add `DatabricksVolumeResource` and `DatabricksVolumeConfig`
- register `databricks_volume` in the Python resource registry

ops
- add `readdir`
- add `stat`
- add `exists`
- add `read_bytes`
- add `read_stream`
- add `range_read`
- add glob resolution

shell commands
- support `ls`
- support `find`
- support `cat`
- support `head`
- support `tail`
- support `stat`
- support `grep`
- support `rg`
- support `tree`

docs and examples
- add Python setup/resource docs
- wire docs navigation and resource matrix
- add a basic workspace example
- add a LangChain deepagents example

### Path handling

- map Mirage paths onto `/Volumes/<catalog>/<schema>/<volume>/...`
- support mounting a subtree with `root_path`
- reject paths that escape above the configured volume root

## Testing

Added tests
- unit tests for path handling, reads, streams, globbing, accessor config, resource wiring, `stat`
- coverage for mounted command dispatch on `databricks_volume`

Ran
- [x] `cd python && uv run pytest tests/core/databricks_volume tests/resource/databricks_volume`
- [x] `./python/.venv/bin/pre-commit run --all-files`
- [x] Databricks volume test
  - mount `DatabricksVolumeResource` at `/dbx/` in a read-only `Workspace`. test
    - `ls /dbx`
    - `find /dbx -maxdepth 2 -type f`
    - `cat /dbx/debug_output.json`
    - `head -n 5 /dbx/debug_output.json`
    - `tail -n 5 /dbx/debug_output.json`
    - `stat /dbx/debug_output.json`
    - `tree /dbx`
  - verify nested traversal / file reads
    - `ls -l /dbx/smoke-20260518`
[...]

Issue #61, opened 2026-05-17
[Feature]: Add Python Databricks Volume resource backend


sonhmai:
### What type of report is this?

Feature request

### Area

Python package

### Summary

I'd like to add a Python `DatabricksVolumeResource` backend that mounts Unity Catalog volumes into Mirage's filesystem abstraction.

Use case: Containers running in Databricks Apps should be able to access Databricks volume files through `Workspace` paths `/Volumes/<catalog>/<schema>/<volume>/<path>`. [Volumes are Unity Catalog](https://docs.databricks.com/aws/en/files/volumes) (Databricks thingy) objects that govern access to non-tabular data.

Example usage:

```python
from mirage import MountMode, Workspace
from mirage.resource.databricks_volume import (
    DatabricksVolumeConfig,
    DatabricksVolumeResource,
)

resource = DatabricksVolumeResource(
    DatabricksVolumeConfig(
        catalog="main",
        schema="default",
        volume="agent_files",
    )
)

ws = Workspace({"/volume": resource}, mode=MountMode.READ)
data = await ws.read("/volume/reports/latest.md")
```

### Proposed API

- New Python resource: `DatabricksVolumeResource`, config: `DatabricksVolumeConfig`
- Resource registry key: `databricks_volume`
- Optional dependency extra: `mirage-ai[databricks]`
- Use Databricks SDK auth by default so Databricks Apps runtime auth works without hardcoded credentials or explicit PAT config as below

```python
# explicit PAT
DatabricksVolumeConfig(
    catalog="main",
    schema="default",
    volume="agent_files",
    host=os.environ["DATABRICKS_HOST"],
    token=os.environ["DATABRICKS_TOKEN"],
)
```

### Tentative plan
1. [x] Scaffolding and read-only DatabricksVolumeResource #66 
2. [x] Write support, other filesystem ops
3. [x] Docs, anything else

sonhmai:
Hi @puneetdixit200, thanks for sending in your version. I created this issue so that I can add the features for my usage of Databricks Volume and therefore would like it to be tested with Databricks (which I'm doing) and for it to be comprehensive - I'm following up to add many other features.

sonhmai:
Added file write in #93

zechengz:
Shipped: resource/databricks_volume in Python, with writes in #93. The auth and client redesign continues in #289. Closing.

PR #109, opened 2026-05-27
feat(databricks-volume): add text commands eg sort, sed, diff, awk, etc.

sonhmai:
## Summary

Adds text-processing commands for the Databricks Volume backend, so an agent can read and analyze file content on a mounted volume (e.g. `wc -l`, `sort | uniq -c`, `sed s/.../.../`, `jq`, `diff`). Continues the Databricks Volume feature (read #66 , write #93) and follows the "delegate to `generic`" pattern in #102. Part of #61.

Commands added (10): `awk`, `cut`, `diff`, `jq`, `nl`, `sed`, `sort`, `tr`, `uniq`, `wc`.

Existing of Databricks volume before this PR `cat`/`grep`/`rg`/`head`/`tail`/`ls`/`stat`/`tree`/`find`/`touch`/`rm`.

## How

- Each command is a thin per-resource wrapper that resolves globs against the backend and hands the bytes/stream to the shared `generic.<verb>` helper.
- **No `core`/`ops` changes.** These are read-only wrappers over the existing `read` op.

## Known limitation

This PR deliberately does not touch the shared `generic/*` helpers. A few of them (`awk`, `cut`, `tr`, `uniq`, and `jq` JSONL streaming) read only the first operand path, so a multi-match glob like `awk '{print}' /dbx/*.csv` processes only the first file today. Proper multi-file support belongs in the generic layer as a separate, cross-backend change.

## Test plan

- [x] `cd python && uv run pytest tests/commands/builtin/databricks_volume tests/core/databricks_volume tests/resource/databricks_volume`
- [x] `uv run python -c "import mirage.commands.builtin.databricks_volume"`
- [x] databricks volume integration test (running against actual volume)

### DIFF
--- a/python/mirage/commands/builtin/databricks_volume/__init__.py
+++ b/python/mirage/commands/builtin/databricks_volume/__init__.py
@@ -12,28 +12,48 @@
 # limitations under the License.
 # ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
 
+from mirage.commands.builtin.databricks_volume.awk import awk
 from mirage.commands.builtin.databricks_volume.cat import cat
+from mirage.commands.builtin.databricks_volume.cut import cut
+from mirage.commands.builtin.databricks_volume.diff import diff
 from mirage.commands.builtin.databricks_volume.find import find
 from mirage.commands.builtin.databricks_volume.grep import grep
 from mirage.commands.builtin.databricks_volume.head import head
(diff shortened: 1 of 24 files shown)


sonhmai:
@zechengz can you review this? Thank you!

zechengz (approved):
Thanks! :)
There are some bugs and missing features from the generic commands. Created them to issues that we can fix them later.

Issue #121, opened 2026-05-28
feat(gdrive): surface Shared Drives as top-level directories


zechengz:
## Problem

The `gdrive` resource lists **My Drive** only. `core/google/drive.py:list_files` builds its query as `'<folder_id>' in parents` with no `corpora` / `driveId`, so files in Shared Drives are invisible.

## Proposal

Surface visible Shared Drives as top-level directories under the mount:

- Thread an optional `drive_id` through `list_files` (set the appropriate `corpora` / `driveId` / `includeItemsFromAllDrives` / `supportsAllDrives` params when present).
- Add a `list_shared_drives` helper that enumerates the Shared Drives the user can see.
- Present each visible Shared Drive as a top-level entry alongside My Drive.

Default behavior (My Drive) stays unchanged when no `drive_id` is given.

Issue #120, opened 2026-05-28
feat(vfs): synthetic mount-parent directories for multi-mount layouts


zechengz:
## Problem

When only deep paths are mounted (e.g. only `/repos/<repo>` exists, with nothing mounted at `/repos`), operations on the *parent* path fail because no resource backs it:

- `ls /repos`
- `cd /repos`
- `stat /repos`
- `find /repos -maxdepth 1`

The parent is a purely structural node implied by its children, but the VFS treats it as nonexistent.

## Proposal

A path that has child mounts but no backing resource of its own should behave as a traversable, listable directory:

- `stat` reports it as a directory
- `ls` enumerates its immediate child mounts
- `cd` can enter it
- `find` traverses it

This is a general VFS-ergonomics fix for any multi-mount layout, not specific to any backend.

sevdeutsch:
Did some empirical digging on current `main` before sending a patch, since this area has clearly evolved. Mounting only `/repos/alpha` + `/repos/beta` (nothing at `/repos`):

| command | `main` today |
|---|---|
| `ls /` (shows `repos`) | ✅ works |
| `find /repos -maxdepth 1` | ✅ works |
| `ls /repos` | ❌ exit 1 |
| `stat /repos` | ❌ exit 1 |
| `cd /repos` | ❌ exit 1 |

So `_inject_child_mounts` (root `ls`) and `_synthesize_find_mount_entries` (`find`) already cover a good chunk of this. The remaining gap is the three operations applied **directly to** a synthetic parent: `ls`, `stat`, `cd`.

Proposed approach (happy to adjust to however you're shaping dispatch):

1. **`workspace.dispatch`** — when `mount_for(path)` has no backing mount but `child_mount_names(path)` is non-empty, answer `readdir` with the child mount names and `stat` with a directory `FileStat`. This is the low-level primitive; it also makes `cd` work, since `handle_cd` stats the target first.
2. **command pipeline (`_dispatch_command_body`)** — `ls`/`stat` resolve a backing mount before reaching dispatch, so add an early branch: if the single target is a synthetic parent, serve `ls` via the existing `_inject_child_mounts` and `stat` via a synthetic directory `FileStat`.
3. `classify` already treats `/repos` as a path (since `find`/root-`ls` work), so likely no change there on `main`.

[...]

PR #123, opened 2026-05-28
feat(github): allow owner/repo/ref in GitHubConfig for YAML-declared repos

### LINKED ISSUE #122: feat(github): allow owner/repo/ref in GitHubConfig for YAML-declared repos
## Problem

A GitHub repo cannot currently be declared purely in workspace YAML the way every other resource can. `owner` / `repo` / `ref` are not part of `GitHubConfig`, so they must be supplied out-of-band.

## Proposal

Let `owner`, `repo`, and `ref` live in `GitHubConfig` so a repo can be declared entirely in workspace YAML, consistent with other resources.

Explicit kwargs still take precedence over config values, so existing call sites are unaffected (backward compatible).

sevdeutsch:
Closes #122.

## Problem
A GitHub repo can't be declared purely in workspace YAML the way other resources can — `owner`/`repo`/`ref` aren't part of `GitHubConfig`, so they must be passed as constructor kwargs out of band.

## Change
- Add optional `owner` / `repo` / `ref` to `GitHubConfig` (`ref` defaults to `"main"`).
- `GitHubResource.__init__` falls back to those config fields when kwargs are absent; explicit kwargs still take precedence (backward compatible). Raises a clear `ValueError` if neither supplies owner/repo.

## Tests
Added to `tests/resource/github/`:
- config exposes `owner`/`repo`/`ref` with correct defaults and values
- constructor falls back to config; kwargs override config; missing owner/repo raises
- `46 passed` (was 41) for `tests/resource/github tests/core/github`.

## Context
One of three generic improvements we carry downstream in a Context Engine built on Mirage; upstreaming so we can drop the local patch. Companion PRs for #120 / #121 to follow.

### DIFF
--- a/python/mirage/core/github/config.py
+++ b/python/mirage/core/github/config.py
@@ -17,3 +17,6 @@ from pydantic import BaseModel, SecretStr
 
 class GitHubConfig(BaseModel):
     token: SecretStr
+    owner: str | None = None
+    repo: str | None = None
+    ref: str = "main"
--- a/python/mirage/resource/github/github.py
+++ b/python/mirage/resource/github/github.py
@@ -47,8 +47,7 @@ class GitHubResource(BaseResource):
         if owner is None or repo is None:
             raise ValueError(
                 "GitHubResource requires owner and repo, either as "
-                "constructor kwargs or in GitHubConfig"
-            )
+                "constructor kwargs or in GitHubConfig")
         default_branch = fetch_default_branch_sync(config, owner, repo)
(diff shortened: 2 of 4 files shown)


python/mirage/core/github/config.py
```diff
@@ -17,3 +17,6 @@
 
 class GitHubConfig(BaseModel):
     token: SecretStr
+    owner: str | None = None
```
zechengz, line 20: Yea we should do this

zechengz (approved):
Thanks!

PR #126, opened 2026-05-29
feat(generic/wc): add format_multi helper, emit POSIX trailing newline

### LINKED ISSUE #114: feat(generic/wc): add multi-file helper and emit POSIX trailing newline
## Problem

`generic.wc` only handles a single source. The multi-file loop is duplicated in every backend wrapper:

```python
# s3/wc.py and databricks_volume/wc.py — identical
outputs: list[str] = []
totals = WCCounts()
for p in paths:
    ...
return "\n".join(outputs).encode(), IOResult()
```

Two issues with this duplication:

1. **POSIX divergence (bug):** `"\n".join(...)` produces output without a trailing newline. Real `wc` ends each record with `\n`. Violates the project's "match BSD/GNU behavior" rule.
2. **Code duplication (~30 lines per backend):** every new backend will copy the same loop.

## Proposed fix

- Add `generic.wc.format_multi(paths, *, read_bytes, accessor, flags, label_fn)` (or similar) that handles the per-file loop, totals row, and POSIX-correct line termination.
- Backend wrappers shrink to a few lines (resolve globs + call the helper).

## Affected wrappers (will simplify)

- `commands/builtin/s3/wc.py`
- `commands/builtin/databricks_volume/wc.py`

## Context

Surfaced in review of #109. Both wrappers carry the bug identically — the fix belongs in `generic`.

sonhmai:
## Summary

Closes #114.

`generic.wc` previously only handled a single source; every backend wrapper duplicated the per-file loop + totals row, and **all of them produced output without a trailing newline** violating POSIX (the `wc` spec terminates every record format string with `\n`, including the totals row).

## POSIX bug

Code in wrapper before the fix:

```python
return "\n".join(outputs).encode(), IOResult()
```

`"\n".join(...)` uses `\n` as a **separator**, not a **terminator** — so the last record (single file, or the `total` row) had no trailing newline. POSIX `wc` (IEEE Std 1003.1-2024) defines the output format with `\n` inside every `printf` string:

```
per-file: "%d %d %d %s\n"
stdin:    "%d %d %d\n"
totals:   "%d %d %d total\n"
```

Without the terminator, downstream tools (`while read`, `tail`, `awk`) treat the last record as incomplete and can drop it.

## Resources wrappers fixed (5)

| Wrapper | Notes |
|---|---|
| `s3/wc.py` | Issue scope. |
| `databricks_volume/wc.py` | Issue scope. |
| `disk/wc.py` | Canonical reference backend, same bug. |
| `hf_buckets/wc.py` | Same pattern as s3, fixed. |
| `nextcloud/wc.py` | Same pattern as s3, fixed. |

Each is now ~10 lines: resolve glob → call helper → return.

## Out of scope (follow-up)

The same buggy pattern still exists in **16 other wrappers**, they all use `"\n".join(outputs).encode()` identically:

```
discord, email, gdocs, gdrive, github, github_ci, gmail, gsheets,
gslides, langfuse, linear, postgres, ram, redis, slack, trello
```

created #131 for this 

## Test plan

- [x] **New helper coverage** in `tests/commands/builtin/generic/test_wc.py`
- [x] **Existing tests updated** to assert the POSIX trailing newline (they were asserting the *bug*):
  - `disk/test_wc.py::test_wc_default`
  - `databricks_volume/test_wc.py::test_workspace_execute_databricks_volume_wc`
- [x] `databricks_volume + disk + generic + native` test suites — **1552 passed**.
- [x] `yapf` / `isort` / `flake8` clean on all changed files.

### DIFF
--- a/python/mirage/commands/builtin/databricks_volume/wc.py
+++ b/python/mirage/commands/builtin/databricks_volume/wc.py
@@ -18,7 +18,7 @@ from mirage.accessor.databricks_volume import DatabricksVolumeAccessor
 from mirage.cache.index import IndexCacheStore
 from mirage.commands.builtin.databricks_volume._helpers import (
     path_prefix, read_bytes_with_index)
-from mirage.commands.builtin.generic.wc import WCCounts, format_wc
+from mirage.commands.builtin.generic.wc import format_multi, format_wc
 from mirage.commands.builtin.generic.wc import wc as generic_wc
 from mirage.commands.builtin.utils.stream import _resolve_source
 from mirage.commands.registry import command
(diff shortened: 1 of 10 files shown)


sonhmai:
this is ready for review.

zechengz (approved):
Fixed some minor issues

PR #133, opened 2026-05-30
fix(databricks): fix volume stream downloads

sonhmai:
## Summary

Part of #61 roadmap.

Fix Databricks Volume `read_stream` so sequential streaming uses a single Databricks file download response instead of one Range GET per yielded chunk.

Before this change, streaming a 100 MB file with the default 8192-byte chunk size caused roughly 12k network GETs. `read_stream` now opens one `files.download()` response, reads the returned body incrementally with `contents.read(chunk_size)`, and closes the body when the stream finishes or is closed early.

`range_read` is unchanged and still uses Range GETs for bounded/random byte reads.

## Tests

- `cd python && uv run pytest tests/core/databricks_volume/test_stream.py`
- `cd python && uv run pytest tests/core/databricks_volume/test_stream.py tests/core/databricks_volume/test_read.py tests/core/databricks_volume/test_head.py`
- `cd python && uv run pytest tests/core/databricks_volume tests/resource/databricks_volume`

### DIFF
--- a/python/mirage/core/databricks_volume/stream.py
+++ b/python/mirage/core/databricks_volume/stream.py
@@ -12,13 +12,34 @@
 # limitations under the License.
 # ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
 
+import asyncio
 from collections.abc import AsyncIterator
+from typing import BinaryIO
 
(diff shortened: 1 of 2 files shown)


sonhmai:
this is ready

PR #135, opened 2026-05-30
fix(databricks_volume): move blocking readdir/stat off the event loop

sonhmai:
## Summary

In the `databricks_volume` backend, `readdir` and `stat` are `async` but called the
**synchronous** Databricks SDK directly inside the coroutine, blocking the event loop
for the duration of each network round-trip:

- `readdir` → `accessor.files.list_directory_contents(...)`
- `stat` → `accessor.files.get_metadata(...)` and `accessor.files.get_directory_metadata(...)`

Fix: Move the blocking SDK calls onto a worker thread via `asyncio.to_thread`.
This also unblocks the generic `copy` path.

## Tests

- Added off-event-loop assertions for `readdir` and `stat` (monkeypatch `asyncio.to_thread`
  with a recorder and assert it's invoked).
- Re ran databricks volume tests

### DIFF
--- a/python/mirage/core/databricks_volume/readdir.py
+++ b/python/mirage/core/databricks_volume/readdir.py
@@ -12,6 +12,7 @@
 # limitations under the License.
 # ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
 
+import asyncio
 import logging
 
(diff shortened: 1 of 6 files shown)


zechengz (approved):
Nice :)

PR #139, opened 2026-05-30
feat(onedrive): OneDrive/SharePoint backend (versioned, snapshot-capable)

zechengz:
## What

Adds a `onedrive` resource that mounts a OneDrive / SharePoint drive as a mirage filesystem over the Microsoft Graph API. Graph is a distinct protocol (not S3-compatible), so this is a full backend with its own `core/onedrive` and command surface, mirroring the S3 layout: full command set, snapshots with version-pinned reads + cTag drift detection, resumable uploads, and a refreshable token. Python only; TypeScript parity is a follow-up.

## Why

OneDrive / SharePoint is a common enterprise store. Hand-written on Graph rather than OpenDAL because OpenDAL's onedrive service lacks `read_with_version` and so cannot do the version-pinned reads snapshots need; Graph exposes it natively (`/versions/{id}/content`) — the same reason S3 stays on aioboto3.

## Changes

- **Core / ops / commands**: `core/onedrive` (Graph client with pagination, retry on 429/503/504 honoring `Retry-After`, and token refresh; version-pinned read; write + resumable upload session; monitor-polled copy; versions list/current/restore; find/du/glob), 14 ops, and the full builtin command surface delegating to the shared generic commands.
- **Resource / registry**: `resource/onedrive`, `ResourceName.ONEDRIVE`, registry entry, `SUPPORTS_SNAPSHOT=True`; files served as raw bytes (no Office-doc conversion).
- **Auth**: `OneDriveConfig.access_token` accepts a static bearer or a refreshable callable provider, plus optional `drive_id` / `site_id` / `key_prefix`.
- **Docs**: setup + resource pages under a new Microsoft nav group.

## Verification

- Scoped unit tests (aioresponses) green: core ops, snapshot capture / replay / drift, retry, token refresh, resumable upload, and multi-file commands; registry and multi-file guard pass; pre-commit clean.
- **Not yet run against a live drive** (needs a Files-scoped token): the `write → read → snapshot-pin → restoreVersion` round-trip and the Graph-shape-dependent paths (copy monitor, `$expand=versions`, downloadUrl) are mock-only.

Trade-offs: `mv` / cross-dir rename is copy+delete (not atomic).

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -132,6 +132,13 @@
                 ]
               },
               {
+                "group": "Microsoft",
+                "icon": "microsoft",
+                "pages": [
+                  "home/setup/onedrive"
+                ]
+              },
+              {
                 "group": "Cloud Files",
                 "icon": "folder-open",
                 "pages": [
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -285,6 +285,13 @@
                 ]
               },
               {
+                "group": "Microsoft",
+                "icon": "microsoft",
+                "pages": [
+                  "python/resource/onedrive"
+                ]
+              },
+              {
                 "group": "Cloud Files",
                 "icon": "folder-open",
                 "pages": [
--- a/docs/home/setup/onedrive.mdx
+++ b/docs/home/setup/onedrive.mdx
@@ -1,0 +1,103 @@
+---
+title: OneDrive
+icon: microsoft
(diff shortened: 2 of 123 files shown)


sonhmai:
@zechengz can you see if we can merge this as first version? Thanks

zechengz:
> @zechengz can you see if we can merge this as first version? Thanks

Sure let me merge

Issue #141, opened 2026-05-31
databricks-volume: recursive cp/mv into own subtree bypasses same-path guard


sonhmai:
## Where
- `python/mirage/commands/builtin/databricks_volume/cp.py` / `mv.py` — `same_backend_file(src, target)` guard
- `python/mirage/core/databricks_volume/copy.py` — `_copy_tree_sync`

## Discussion
The same-path data-loss fix (#127) compares **exact** resolved backend keys (`backend_path(src) == backend_path(target)`). That catches `cp/mv a a`, `mv a dir/` (file already in `dir/`), and aliases that collapse to one key.

It does **not** catch a recursive copy/move of a directory **into its own subtree**:

```
cp -r /dbx/dir /dbx/dir     # dst is a dir -> target = /dbx/dir/dir
```

`target`'s backend key (`/dbx/dir/dir`) differs from the source (`/dbx/dir`), so the exact-match guard is bypassed and `_copy_tree_sync` descends into the freshly created copy. coreutils rejects this with `cp: cannot copy a directory, '/dbx/dir', into itself, '/dbx/dir/dir'`.

This is the same self-referential class #127 targets but was deliberately scoped out of that fix.

## Suggested fix
For recursive directory operations, reject when `backend_path(dst)` is **nested under** `backend_path(src)` (i.e. `dst_key == src_key` or `dst_key.startswith(src_key + "/")`), raising the coreutils-style "cannot copy/move a directory into itself" error before any upload.

## Verification
- `cp -r /dbx/dir /dbx/dir` and `mv /dbx/dir /dbx/dir` error with non-zero exit; no partial tree written.
- Normal recursive copy to a sibling/outside path still works.

Found during multi-agent review (codex + gemini + claude-code) of the #127 fix.

zechengz:
Folding in extra context from #152 (closing that as a duplicate of this one):

**Provenance:** #142 (the PR that added the exact-key same-path guard) explicitly flagged this recursive `dir -> dir/dir` case in its description but deliberately did not fix it, and no test covers it. So this is a known, deferred gap, not a fresh discovery.

**Additional sub-task to fold into the same fix (command-layer ordering nit):**
In `cp.py` / `mv.py` the same-path guard currently runs *before* the source-existence check, so:

```
cp /dbx/missing /dbx/missing
```

[...]

PR #142, opened 2026-05-31
fix(databricks-volume): prevent mv/cp onto same path from deleting the file (#127)

### LINKED ISSUE #127: databricks-volume: `mv`/`cp` onto the same path deletes the file (data loss)
## Where
- `python/mirage/core/databricks_volume/rename.py` — `rename()` is `copy(src, dst)` then `unlink(src)` / `rm_recursive(src)`
- `python/mirage/core/databricks_volume/copy.py` — file copy is `download → upload(overwrite=True)`
- `python/mirage/commands/builtin/databricks_volume/mv.py` and `cp.py` — no `src == target` guard before delegating

Introduced in #116.

## Discussion
The Databricks Files API has no native rename, so `mv` is implemented as copy-then-delete. Because the file `copy` is `download → upload(overwrite=True)`, copying a path onto itself **succeeds**, and then `mv` deletes the source — destroying the file.

```
mv /v/a.txt /v/a.txt      # download a.txt -> upload a.txt (overwrite) -> delete a.txt  => GONE
```

This is reachable whenever src and dst resolve to the same key (`mv a a`, `mv ./a a`, a glob that expands to the same path, or `mv a dir/` where the file already lives at `dir/a`).

Note: the same missing-guard pattern exists in other resources' `mv` (disk, s3), but it is **not** triggerable there — `disk` uses `os.rename` (no-op), and `s3`'s `copy_object` rejects a self-copy with no metadata change, so the delete never runs. databricks is uniquely exposed because its copy is a plain download/upload.

POSIX `mv`/`cp` error with *"'a' and 'a' are the same file"*.

## Suggested fix
Guard against `src.strip_prefix == target.strip_prefix` in databricks `cp.py`/`mv.py` (or in core `rename`/`copy`) and skip/raise before any upload+delete.

sonhmai:
## Summary
Fixes #127.

Bug: Moving a path onto itself uploaded the file onto itself and then deleted it. 

New issue detected in #141: recursive `cp -r/mv dir dir` into its own subtree (`target = dir/dir`) bypasses the exact-key guard.

## Tests
- Command: same-path `mv`/`cp` error + preserve file; `mv a dir/` variant.
- Core/ops: `rename`/`copy` same-path no-op; same-path **missing** source raises `FileNotFoundError`; non-recursive same-dir raises `IsADirectoryError`.
- Full `tests/{commands/builtin,core}/databricks_volume` suite (124 tests) passes; isort/flake8/ruff/yapf clean.

### DIFF
--- a/python/mirage/commands/builtin/databricks_volume/_helpers.py
+++ b/python/mirage/commands/builtin/databricks_volume/_helpers.py
@@ -19,6 +19,7 @@ from mirage.accessor.databricks_volume import DatabricksVolumeAccessor
 from mirage.cache.index import IndexCacheStore
 from mirage.commands.builtin.utils.wrap import (call_read_bytes,
                                                 call_read_stream)
+from mirage.core.databricks_volume.path import backend_path
 from mirage.core.databricks_volume.read import read_bytes as _read_bytes
 from mirage.core.databricks_volume.stat import stat as _stat
 from mirage.core.databricks_volume.stream import read_stream as _read_stream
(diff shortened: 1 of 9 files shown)


zechengz (approved):
Thanks! :)

PR #147, opened 2026-06-01
fix(commands): terminate generated record output with newlines

### LINKED ISSUE #137: Missing trailing newline in many command outputs (separator-not-terminator)
## Summary

Many builtin commands build stdout with `"\n".join(records).encode()`, which uses `\n` as a **separator**, not a **terminator** — so the last line has no trailing newline. POSIX coreutils terminate every output line (including the last) with `\n`; without it, downstream tools (`while read`, `tail`, `awk`) can drop the final record.

This is the same bug fixed for `wc` in #126 (file-args + stdin), but it is **systemic** across commands and backends, not wc-specific.

## Affected commands (separator-not-terminator pattern)

`find`, `grep`, `rg`, `du`, `stat`, `ls`, `tree`, `md5`, `cmp`, and `wc` (the 16 backends still pending in #131) — appearing in essentially every backend (generic, disk, s3, ssh, github, gdrive, gmail, slack, telegram, notion, mongodb, postgres, databricks_volume, …).

Some commands already terminate correctly (`shuf`, `fold`, `bc`, `date`, `tar`, `unzip`, `realpath`, `rev`), which shows the inconsistency comes from each command formatting output by hand.

Not affected (faithful pass-through): `sed`, `diff`, `cat` preserve input newlines; `tee` empty case; `stderr` warning strings.

## Proposed fix

Per-call-site patching doesn't scale. Add a single shared output helper (e.g. `to_stdout(records: list[str]) -> bytes` that always ends in `\n` and returns `b""` for empty) and apply it uniformly across commands/backends.

Related: #114, #126, #131.

sonhmai:
Closes #137 

Detected an out-scope issue and created #149 for it.

## Problem

  Many builtin commands built stdout with `"\n".join(records).encode()` — `\n` as a
  **separator**, not a **terminator** — so the last line had no trailing newline (unlike
  POSIX coreutils, where every line including the last ends in `\n`):

  - `find /s3 -type f | wc -l` undercounted by one (last path had no newline to count).
  - `while read` / `tail` / `awk` could drop the final record.

  Systemic across `find/grep/rg/du/stat/ls/tree/md5/cmp/file/wc` in essentially every backend.

## Fix

  Add shared record-output helpers in `commands/builtin/utils/output.py` and apply it uniformly across backends.
  
  Byte-preserving commands (`cat`, `sed`, `diff`, `head`, `tail`, `awk`) untouched.
  This keeps byte-preserving commands untouched while making generated command records end with POSIX-style trailing newlines. 

  ## Tests

  - Updated generic, RAM/Redis, and disk `rm` tests for the new contract.
  - Updated per-backend exact-output assertions (dify, discord, github, notion, s3, workspace) which delegates to generic layer.
  - `integ/truth.txt` `find_pipe_wc` `18 → 19`: the old missing terminator made `wc -l` undercount by one; the new value matches real coreutils.
  - Affected suites: 93 passed, 0 failed. pre-commit (yapf/isort/flake8) clean.

### DIFF
--- a/integ/truth.txt
+++ b/integ/truth.txt
@@ -488,7 +488,7 @@ world
       2 b
       1 c
 === find_pipe_wc ===
-18
+19
 === jq_pipe_wc ===
 3
 === cat_tr_sort ===
--- a/python/mirage/commands/builtin/databricks_volume/find.py
+++ b/python/mirage/commands/builtin/databricks_volume/find.py
@@ -16,6 +16,7 @@ import fnmatch
 
 from mirage.accessor.databricks_volume import DatabricksVolumeAccessor
 from mirage.cache.index import IndexCacheStore
+from mirage.commands.builtin.utils.output import format_records
 from mirage.commands.registry import command
 from mirage.commands.spec import SPECS
 from mirage.core.databricks_volume.glob import resolve_glob
@@ -101,4 +102,4 @@ async def find(
     depth = int(maxdepth) if maxdepth is not None else None
     results = await _find_recurse(accessor, path, name, type_filter, depth, 0,
                                   index)
-    return "\n".join(results).encode(), IOResult()
+    return format_records(results), IOResult()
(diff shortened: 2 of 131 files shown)


integ/truth.txt
```diff
@@ -488,7 +488,7 @@ world
       2 b
       1 c
 === find_pipe_wc ===
-18
+19
```
sonhmai, line 497: find now correctly terminates its final generated record with \n so `find ... | wc -l` counts one additional line

sonhmai:
@zechengz this is now ready

zechengz (approved):
Resolved the conflicts when merged with main and some other minor changes.

PR #155, opened 2026-06-01
fix: add provenance to dify search results

### LINKED ISSUE #143: py/dify: follow-up improvements for the Dify Knowledge resource
Follow-up improvements for the read-only Dify Knowledge resource added in #136. Listed roughly in priority order.

1. **`search` results have no provenance.** `records_to_bytes` in `python/mirage/core/dify/search.py` emits only segment `content`. It drops the source document path and the relevance score Dify returns, so results are not attributable and an agent cannot follow up (e.g. `cat` the source). Prefix each result with its path (and optionally score), e.g. `policies/refunds:0.82\n<content>`. (The #136 description claimed this `path:score\ncontent` format, but it was never implemented.)

2. **`ls` makes one API call per file.** `python/mirage/commands/builtin/dify/ls.py` wires the heavy `stat` (which calls `get_document_detail`), while `tree`/`find`/`rg` use the index-served `stat_light`. Plain `ls` of a folder fires one detail call per file for data already in the index. Switch to `stat_light`. The `integ/dify.py` call accounting already measures this (the `detail=` lines should drop to 0).

3. **A fresh `httpx.AsyncClient` per request.** `python/mirage/core/dify/_client.py` opens and closes a client inside every `dify_get`/`dify_post`, so each call is a new TCP + TLS handshake. A single `ls` + `grep -r` does dozens of calls. Reuse one client (on the accessor) for keep-alive and pooling.

4. **Sequential fan-out.** `grep -r` reads each document's segments one at a time, search's `target_entries` resolves children in an await loop, and `iter_segment_pages` is serial. Parallelize the independent calls with `asyncio.gather` for large datasets.

5. **One bad document breaks the whole mount.** `build_dir_entries` in `python/mirage/core/dify/tree.py` raises on a duplicate slug or a path collision. Since the tree builds lazily on first access, every `ls`/`cat`/`search` then fails. Consider log-and-skip the offending document instead of a hard failure, so one misconfigured doc does not take down the knowledge base.

6. **Commands advertise flags they ignore.** The Dify commands register the shared `SPECS` (full Unix flag sets) but implement a subset and swallow the rest via `**_extra`, so an agent is told about flags that silently do nothing. Either implement the gaps or trim the specs to what is actually supported.

7. **Remove the stray `dify-knowledge-review.md`** at the repo root (an accidental review artifact with absolute local paths).

8. **Error-path tests.** Use the `integ/dify.py` mock server to return 429 / 500 and assert the retry and backoff in `_client.py`, plus tests for `FileNotFound` and the duplicate-slug `ValueError`. These HTTP-level paths are currently untested.

9. **Minor:** `fingerprint` always returns `None` and `load_state` is a no-op. Fine for read-only today, but documents carry `updated_at`, which could feed a fingerprint for cache invalidation or snapshot support later.

ki3nd:
# fix(py/dify): add provenance to search results

Partly fixed #143

## Summary

Update Dify `search` result formatting so each retrieval hit includes source provenance instead of emitting only raw segment content.

This change makes `search` return one block per Dify retrieval record in the format:

```text
/knowledge/path/to/doc:0.82
<chunk content>
```

The path is derived from the configured slug metadata field and falls back to the Dify document name when slug metadata is missing. Multiple hits from the same document remain separate records.

## Changes

- format Dify search results as `absolute_workspace_path[:score]\ncontent`
- propagate mount prefix through the Dify search command/op path so root searches still emit absolute workspace paths
- keep one output block per retrieval record instead of collapsing hits from the same document
- update tests to cover slug-based paths, document-name fallback, optional score formatting, and repeated hits for the same document
- update Dify prompt/docs to match the implemented output format

## Why

The previous implementation dropped document provenance and relevance score, which made search results unattributable and harder for an agent to follow up with commands like `cat`, `head`, or `grep`.

## Testing

```bash
cd python && uv run pytest tests/core/dify/test_search.py tests/commands/builtin/dify/test_search.py tests/ops/dify/test_search.py -v
```

### DIFF
--- a/docs/python/resource/dify.mdx
+++ b/docs/python/resource/dify.mdx
@@ -133,6 +133,17 @@ Supported methods:
 
 `top-k` is capped at `100`. `threshold` must be between `0` and `1`.
 
+Results are emitted one retrieval hit per block:
+
+```text
+/knowledge/policies/refunds:0.82
+Refunds are allowed within 30 days.
+```
+
+Mirage uses the configured slug metadata field to derive the path, and falls
+back to the Dify document name when that metadata is missing. Multiple hits
+from the same document remain separate records.
+
 Scoped search uses Dify metadata filtering. Documents with the configured slug
 metadata field are filtered by that field; name-based documents are filtered by
 `document_name`, which requires Dify Built-in Fields to be enabled in dataset
--- a/integ/dify.py
+++ b/integ/dify.py
@@ -96,24 +96,64 @@ def _document_detail(doc: tuple) -> dict:
 
 
 def _retrieve_records(query: str) -> list[dict]:
+    def build_record(doc_id: str, content: str, score: float) -> dict:
+        doc = DOC_BY_ID[doc_id]
+        return {
+            "segment": {
+                "id": f"{doc_id}:{score:.2f}",
+                "document_id": doc_id,
+                "content": content,
+                "document": {
+                    "id": doc[0],
+                    "data_source_type": "upload_file",
+                    "name": doc[1],
+                    "doc_type": None,
+                    "doc_metadata": _document_summary(doc)["doc_metadata"],
+                },
+            },
+            "child_chunks": [],
+            "score": score,
+            "tsne_position": None,
+            "files": [],
+            "summary": None,
+        }
+
     lowered = query.lower()
     if "throttl" in lowered or "rate" in lowered or "429" in lowered:
-        contents = [
-            "Requests are rate limited to 100 calls per minute per token.",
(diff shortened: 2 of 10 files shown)


zechengz:
I pushed some changes / minor fixes.

I haven't tested them maybe you can help to test if you are available :)

Do you plan to fix all issues listed in the https://github.com/strukto-ai/mirage/issues/143 in this PR? If not we can change Fixed to partly fixed. Thanks.

ki3nd:
Thanks for the clarification. I’ve changed it to “Partly fixed” in the merge request.

zechengz (approved):

PR #157, opened 2026-06-02
fix: correct rg count fallbacks

### LINKED ISSUE #149: grep -c / rg -c report wrong counts (always 1) and don't skip zero-match files
## Summary

`grep_lines(..., count_only=True)` returns a **single-element list** `[str(count)]` (`grep_helper.py:29-30`):

```python
if count_only:
    return [str(count)]
```

Callers that then compute `len(matched)` get `1` as the list length, not the count. And callers that guard with `if not matched:` to skip zero-match inputs never skip, because `["0"]` is truthy. So `-c` is wrong in two independent ways.

### Issue 1: count is always `1`

```
printf 'foo foo\nfoo bar\nbaz\n' | rg -c foo   →  1   (should be 2)
```

Per-file server-side counts collapse the same way: every matched file shows `path:1`.

### Issue 2: zero-match inputs not skipped

`grep_lines(count_only=True)` returns `["0"]` for no matches → truthy → the `if not matched:` short-circuit never fires. Stdin `rg -c` on a non-matching input returns `1` with exit 0 (should be exit 1 / no output); server-side emits the file and sets `any_match=True`.

Affected locations:
1. **stdin `-c` path** (`return str(len(matched)).encode()` → should use the count value):
`email/rg.py:123`, `gmail/rg.py:169`, `langfuse/rg.py:183`, `mongodb/rg.py:195`, `postgres/rg.py:191`, `slack/rg.py:193`, `discord/rg.py:206`, `github/rg.py:194`
2. **server-side `-c` path** (`all_results.append(f"{path}:{len(matched)}")`):
`discord/rg.py:180`, `email/rg.py:99`, `github/rg.py:167`, `gmail/rg.py:145`, `langfuse/rg.py:159`, `mongodb/rg.py:171`, `postgres/rg.py:167`, `slack/rg.py:169`, `email/grep.py:259`

(`general/grep.py` is **not** affected — its `-c` count is correct; only its trailing newline was fixed, in #147.)

### Semantics to decide

`rg -c` omits zero-count files; `grep -rc` lists `file:0`. Mirage commands track coreutils/rg per CLAUDE.md — pick one and apply consistently to the server-side path.

### Root cause: the fallback is duplicated, not shared

The 8 affected backends are push-down search backends (slack/discord/mongodb/postgres/github/gmail/email/langfuse). Each `rg`/`grep` has three sections: native push-down (backend-specific), a byte-scan fallback, and a stdin fallback. Only the push-down is unique — the byte-scan and stdin sections are copy-pasted from the generic logic into every backend, so this one bug spans all 17 sites above. `generic/rg.py:150` already routes stdin through `grep_stream`, which yields the real count, newline-terminated — that's why `generic`/`dify` `rg -c` returns `2` while these return `1`. The push-down backends just don't call it.

### Proposed fix

1. Delete the duplicated byte-scan/stdin sections and delegate the non-push-down fallback to `generic.rg` / `generic.grep` (same consolidation as #99, #102, #119, #145). The count then lives in one place and can't drift again — a per-site `matched[0]` patch only re-syncs the clones until the next drift.
2. Apply the chosen zero-count semantics once, in `generic`.
3. Add coverage: the stdin and server-side `-c` paths are untested, and `email`/`langfuse`/`postgres` have **no** rg/grep tests at all — which is why this went unnoticed.

## Out of scope

#147 only added trailing-newline termination on these lines (POSIX #137). The count-value and zero-skip correctness is tracked here.

sonhmai:
## Summary

Fixes #149.

- Fix fallback `rg -c` semantics so zero-count files are omitted, and stdin no-match exits 1 with no output.
- Add shared count helpers so generic and backend fallback paths parse count output consistently.
- Route backend fallback/stdin `rg` behavior through generic `rg` instead of duplicated implementations.
- Fix email server-side count mode to report actual match counts.

## Testing

- `pytest tests/commands/builtin/generic/test_rg.py tests/commands/builtin/helpers/test_rg_helper.py tests/commands/builtin/github/test_commands.py tests/commands/builtin/email/test_grep.py`

### DIFF
--- a/python/mirage/commands/builtin/discord/rg.py
+++ b/python/mirage/commands/builtin/discord/rg.py
@@ -17,9 +17,8 @@ from collections.abc import AsyncIterator
 
 from mirage.accessor.discord import DiscordAccessor
 from mirage.cache.index import IndexCacheStore
-from mirage.commands.builtin.grep_helper import compile_pattern, grep_lines
+from mirage.commands.builtin.generic.rg import rg as generic_rg
 from mirage.commands.builtin.utils.output import format_records
-from mirage.commands.builtin.utils.stream import _read_stdin_async
 from mirage.commands.registry import command
 from mirage.commands.spec import SPECS
 from mirage.core.discord.channels import list_channels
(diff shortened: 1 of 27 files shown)


zechengz (approved):
LGTM! Thanks! :)

Also fixed something related to github slack etc

We have some scope definition for example if we grep xxx /slack or /github mounts directly (which has a very large scope) we to use the slack / github search
if we just grep an attachment file in slack or we have some smaller scope search in github, then we can reuse the generic grep and fetch all files needed

PR #160, opened 2026-06-04
Fix Dify Knowledge Follow-ups

ki3nd:
# Fix Dify Knowledge Follow-ups

## Summary

Addresses Dify Knowledge follow-ups from #143, covering items 2-9.

- Use light stat metadata for Dify `ls` so listing does not call document detail APIs.
- Add bounded parallelism for Dify grep reads.
- Skip bad Dify documents with warnings instead of failing the whole mount.
- Trim Dify command specs to avoid advertising unsupported flags.
- Add retry/error-path tests and a lightweight Dify fingerprint based on document metadata.

## Validation

- `uv run --project python pytest python/tests/core/dify python/tests/ops/dify python/tests/commands/builtin/dify python/tests/resource/dify python/tests/workspace/test_dify_resource.py -v`
- `./python/.venv/bin/python integ/dify.py`
- `SKIP=ts-prettier,ts-eslint ./python/.venv/bin/pre-commit run --all-files`
- `git diff --check`


ki3nd:
@zechengz please review this pr.

zechengz:
> @zechengz please review this pr.

Can you help to merge with latest main at first?

I need to spend some time think about the semaphore design. May take some time.

Thanks.

zechengz (approved):
merge latest main and refactored
let's move those semaphore logic to the ConcurrencyLimiter as we can use remote one in the future instead of just use local ones
for the retrying we can use the tenacity
Thanks for the contribution.

PR #202, opened 2026-06-06
feat(py/chroma): add read-only ChromaDB resource

ki3nd:
# feat(py/chroma): add read-only ChromaDB resource

## Inspiration

Inspired by Mintlify's [*How we built a virtual filesystem for our Assistant*](https://www.mintlify.com/blog/how-we-built-a-virtual-filesystem-for-our-assistant), which describes **ChromaFs** — a virtual filesystem over their existing Chroma database that eliminated the need for sandboxed containers entirely.

Their key insight: an agent doesn't need a real filesystem, it just needs the illusion of one. The entire directory tree is stored as a single gzipped `__path_tree__` document in the Chroma collection; `ls`, `cd`, and `find` resolve in local memory with zero network calls after the first fetch. Pages are stored as chunks sorted by `chunk_index` and joined on `cat`. `grep -r` is accelerated by pushing the pattern into a Chroma `$contains` / `$regex` filter, pre-fetching only matching chunks, and handing the result back to the shell for in-memory fine filtering.

This PR ports the same architecture into Mirage as a first-class Python resource.

## Summary

Add read-only Python support for `ChromaDB` — mounting a ChromaDB collection as a virtual filesystem via `AsyncHttpClient`.

This branch adds the resource/config/registry wiring, async ChromaDB client, path-tree-based directory caching, slug-based path resolution, semantic vector search, read-only file ops, glob support, expanded tests, integration harness, and docs/examples.

### Supporting

**new resource**
- add `ChromaResource` and `ChromaConfig` (`host`, `port`, `ssl`, `collection_name`, `slug_field`, `chunk_index_field`)
- register `chroma` in the Python resource registry
- `ChromaAccessor` uses `chromadb.AsyncHttpClient` exclusively — no sync `PersistentClient`

**async client (`accessor/chroma.py`)**
- `get_client` / `get_collection` are async methods; client is lazily initialized per accessor instance
- all collection calls (`get`, `query`) are properly awaited

**index / tree (`core/chroma/tree.py`, `path.py`)**
- `ensure_tree` builds a RAM index from the `__path_tree__` document on first access (lazy, per-prefix)
- path tree is stored as a gzip+base64 JSON blob in the collection; `parse_path_tree` handles both plain JSON and encoded formats
- `resolve_path` resolves a `PathSpec` to a document entry or virtual directory
- `build_dir_entries` validates for duplicate and collision paths, raises `ValueError` on conflict

**chunk pagination (`core/chroma/_client.py`)**
- `page_chunks` fetches chunks in batches of 100 (`PAGE_CHUNK_BATCH_SIZE`) with `offset`/`limit` to avoid full collection scans
- `query_contains` does a filtered `$contains`/`$regex` search scoped to candidate slugs
- `fetch_path_tree` fetches the `__path_tree__` document by ID

**search (`core/chroma/search.py`)**
- `search_segments` calls ChromaDB vector search (`collection.query`), returns `path:score\ncontent` formatted output matching Dify convention
- scopes results to a sub-folder by building a `$in` slug filter when paths are provided
- `score_from_distance` converts L2 distance to a `[0, 1]` similarity score

**shell commands**
- `ls`, `find`, `tree`
- `cat`, `head`, `tail`
- `grep`
- `search` (ChromaDB-native vector retrieval)

**docs and examples**
- add Python setup and resource docs (`docs/python/setup/chroma.mdx`, `docs/python/resource/chroma.mdx`)
- add home-level setup doc (`docs/home/setup/chroma.mdx`)
- move ChromaDB from Knowledge to Database in docs navigation (`docs/docs.json`) and resource matrix
- add workspace example (`examples/python/chroma/chroma.py`, `examples/python/chroma/chroma_vfs.py`)

### Path handling

- documents are mapped to virtual paths via `slug_field` metadata (default: `page_slug`)
- the full path tree is stored as a single `__path_tree__` document in the collection, mapping slugs to metadata
- paths are resolved relative to the collection mount prefix (e.g. `/knowledge/`)
- directories are inferred from common slug path prefixes; there are no real folder objects in ChromaDB
- chunk ordering is determined by `chunk_index_field` metadata (default: `chunk_index`)

## Testing

**Added tests**
- unit tests for `_client`, `path`, `tree`, `read`, `glob`, `grep`, `search`
- command tests for `ls`, `find`, `cat`, `head`, `tail`, `grep`, `search`
- async fixture `FakeCollection` with `get` / `query` mocked as async methods
- `conftest.py` uses `functools.partial` over a top-level `_get_collection` helper (no nested functions per `AGENTS.md`)

**Integration**
- `integ/chroma.py` — standalone script that wires a `FakeCollection` mock and runs all commands against a `Workspace`, printing `=== name ===` / output blocks
- `integ/truth_chroma.txt` — expected output for all cases; verified with `check_lines.sh`

**Ran**
```
cd python && uv run pytest tests/commands/builtin/chroma/ tests/core/chroma/ tests/resource/chroma/ → all passed
./python/.venv/bin/python integ/chroma.py 2>&1 | bash integ/check_lines.sh integ/truth_chroma.txt → OK: 65 lines matched
./python/.venv/bin/pre-commit run --all-files
```


zechengz:
Pushed two maintainer commits:

**fc38e445** resolves the merge conflict with main (docs.json: kept both chroma and lancedb under Database; uv.lock regenerated from the merged pyproject).

**dc63f16d** review follow-ups:

- **Thin client**: `chromadb` -> `chromadb-client`. The accessor only uses `AsyncHttpClient`, and the full package was pulling onnxruntime/kubernetes/grpcio/tokenizers/posthog into every `--all-extras` install and CI run. Verified `AsyncHttpClient` works on `chromadb-client` 1.5.9.
- **8 more commands** via generic delegation, mirroring the dify set: `awk`, `cut`, `rg`, `sed`, `sort`, `stat`, `uniq`, `wc` (16 total now).
- **Missing unit tests** for `_client`, `path`, `glob`, `find`, `walk`, `stat`, `readdir` (+36 tests, reusing the existing FakeCollection fixtures).
- **Real ChromaDB OSS integ**: new `integ/chroma_live.py` seeds a real server (service container) with the same dataset, explicit embeddings, and the gzip+base64 path-tree variant, then runs the shared filesystem cases. Output is byte-identical to the FakeCollection harness, which also live-verifies `$regex` / `$contains` pushdown on a real server. Search cases stay on the mock harness (the thin client deliberately ships no embedding function for `query_texts`).
- **CI**: `test_integ_mongodb.yml` is now `test_integ_databases.yml` with mongodb + chroma jobs; the chroma job runs both the mock harness and the OSS harness, so the previously-unwired `integ/chroma.py` now gates CI.
- Mock fidelity fixes: FakeCollection now does real regex for `$regex` (was substring), and `page_chunks` uses `zip(strict=True)`.

All verified locally: 386 tests pass, both integ harnesses diff clean against truth (live one against a real `chromadb/chroma:latest` container, deterministic across runs), full pre-commit passes.

zechengz:
Third maintainer commit (19ddd0a2), follow-ups from review discussion:

- **Real-server-only integ.** The FakeCollection integ harness is gone; `integ/chroma.py` now seeds a real ChromaDB OSS server (unique collection, explicit embeddings, gzip+base64 path tree) and runs 26 cases against it. One truth file (`truth_chroma.txt`). FakeCollection survives only in unit-test fixtures. Vector search is not integ-exercised (the thin client ships no embedding function for `query_texts`); it stays covered by unit tests.
- **`chroma-query` command.** The vector retrieval command is registered as `chroma-query` instead of `search` (docs and example updated). The internal VFS op layer keeps the standard `search` op name shared by all backends.
- **Shared score helper.** `score_from_distance` / `format_score` moved to `mirage/utils/score.py`; chroma and dify both import from there now.
- **TypeScript mirror.** Full `chroma` backend in `packages/core` (tree/path/read/grep pushdown/glob/find/stat/walk, 16 commands delegating to TS generics, `chroma-query`, ops, resource, accessor) plus `integ/chroma.ts`. The TS harness runs the same cases against a real Chroma server and its output is **byte-identical** to the shared `truth_chroma.txt` the Python harness produces. `chromadb` (npm) is an **optional peer dependency** (same pattern as pyodide/redis), loaded via dynamic import only when a chroma mount is used.
- **CI**: `test_integ_databases.yml` now has three jobs: mongodb, chroma (python vs OSS server), chroma-ts (TS vs OSS server, same truth file).

Verified locally: python and TS integ harnesses both diff clean against the one shared truth (real `chromadb/chroma:latest` container), 58 chroma/score python tests + 19 TS vitest tests pass, core typecheck 0 errors, eslint clean, full pre-commit passes.

PR #209, opened 2026-06-07
feat(generic/uniq): parse -f/-s/-w inside generic, distinguish 0 from unset

### LINKED ISSUE #112: feat(generic/uniq): parse -f/-s/-w inside generic, distinguish 0 from unset
## Problem

`generic.uniq` types `skip_fields`, `skip_chars`, `check_chars` as `int`, forcing wrappers into:

```python
skip_fields=int(f) if f else 0
```

Falsy `"0"` is indistinguishable from unset (benign today because 0 is the default, but bug-prone). It also pushes string→int parsing duplication into every backend wrapper.

## Proposed fix

- Move `-f/-s/-w` parsing into generic; accept the raw string from the wrapper.
- Distinguish unset (`None`) from `0` explicitly.
- Wrappers forward strings without conversion.

## Affected wrappers

- `commands/builtin/s3/uniq.py`
- `commands/builtin/databricks_volume/uniq.py`

## Context

Surfaced in review of #109. Same pattern as the s3 wrapper.

Hexecu:
## Summary

Moves `-f` / `-s` / `-w` parsing into `generic.uniq` so the raw string from the
wrapper is parsed in one place, and unset is represented as `None` rather than
collapsing into `0`.

- `generic.uniq` now takes `skip_fields` / `skip_chars` / `check_chars` as
  `str | None` and parses them via a small `_parse_count` helper (`None -> 0`,
  otherwise `int(value)`). `_comparison_key` / `_uniq_stream` keep working on
  `int` internally — only the conversion point moved.
- Wrappers forward the raw `f` / `s` / `w` strings instead of repeating
  `int(x) if x else 0`, so a literal `"0"` is no longer indistinguishable from
  unset.

## Why all the generic-routed wrappers, not just s3 / databricks

The issue lists `s3` and `databricks_volume` as the wrappers that surfaced this,
but changing the `generic.uniq` signature touches every wrapper that delegates to
it — leaving the others on the old `int(x) if x else 0` call would pass strings
into an `int` parameter. So the same one-line forwarding change is applied to all
generic-routed wrappers: `s3`, `databricks_volume`, `dify`, `disk`, `gdrive`,
`hf_buckets`, `nextcloud`, `ram`, `redis`, `ssh`.

`github/uniq.py` is intentionally left untouched: it has its own self-contained
implementation and does not delegate to `generic.uniq`, so it is unaffected by
the signature change.

## Tests

Adds `tests/commands/builtin/generic/test_uniq.py` covering `_parse_count`, the
unset-vs-`"0"` distinction, field/char skipping, `check_chars` limiting (including
`"0"` comparing the full line), `-c` counting, and `-i` case folding.

Closes #112


zechengz:
Pushed two follow-up commits onto this branch (maintainer edit):

- **d51da266**: implement the unset-vs-0 distinction behaviorally: GNU `uniq -w 0` compares zero characters (all lines collapse to the first), while unset compares the full line. Same fix mirrored in `generic/uniq.ts` since TS integ shares `truth.txt`. Negative counts now raise instead of mis-slicing. Added `-f/-s/-w` integ cases for ram/disk/redis (py+ts), opfs, s3/gcs/minio, dify, and regenerated truth files (all verified against GNU coreutils).
- **68c5e072**: delegate `github/uniq.py` to generic via a new `_open_read_stream` normalizer (accepts async-generator or awaitable readers), so the last bespoke Python uniq inherits the same parsing/semantics. Adds wrapper tests.

Follow-ups filed: #223 (TS bespoke uniq wrappers), #224 (unawaited read_stream in other generic commands), #225 (nextcloud integ flag coverage).

zechengz (approved):
Thanks for the contribution :)

Issue #187, opened 2026-06-06
[F] Fsafe: Same-path / same-backend guard (#150, #154)


pzhang15:
**Lane F** · part of #186 · roadmap #165

### Context
S3 rename is `copy_object` + **unconditional** `delete_object` (#150); the cross-mount mover does an **unconditional** unlink (#154). Both delete the only copy when src and dst alias the same backend path.

### Scope
Lift the databricks guard template into a shared `same_backend_file(...)` helper in `core/_shared/path_guard.py`; apply it to S3 rename and the wired cross-mount mover. Guard runs after `stat` (missing source still raises). Remove broad `except Exception` masking in the touched paths.

- **Files:** `core/_shared/path_guard.py`, `core/s3/rename.py`, `workspace/executor/cross_mount.py`
- **Depends on:** — (only Phase 0)
- **Consumes (frozen interface):** —

### ✅ Definition of Done (verifiable)
- [ ] S3 `mv`/`cp` onto an aliased same path no longer deletes the file (#150) — regression test
- [ ] cross-mount `mv` where src/dst alias the same backend path no longer unlinks the only copy (#154) — regression test
- [ ] guard runs **after** stat (a missing source still raises) — test
- [ ] no broad `except Exception` masking remains in the touched paths
- [ ] `pre-commit run --all-files` clean and the relevant tests green (`cd python && uv run pytest` / `pnpm test` for TS)

zechengz:
Closing as superseded by #721. Lane F (backend safety) is Phase 1 there, unchanged in substance. Nothing from this plan was built on main (no workspace/fork.py, no commit_batch, no StagedOp), so nothing is lost by re-homing it.

Issue #186, opened 2026-06-06
Lane F · Backend safety


pzhang15:
**Part of the roadmap: #165** · module(s): `core/_shared/path_guard.py`, `core/s3/rename.py`, `workspace/executor/cross_mount.py`

The same-path / same-backend guard. Fixes the #150 and #154 data-loss bugs. **Zero dependencies — ships immediately**, and is the safety floor commit replay relies on.

- **Consumes (frozen interfaces):** Phase 0 (optional): the `same_backend_file` signature (F4)
- **Produces:** the guard helper Lane G + Lane J reuse
- **Internally divisible** into the sub-issues below — independent files, so multiple engineers can share this lane without contention.

### 🎯 Outcome — done when (verifiable)
- [ ] the **#150** (S3 same-path mv/cp) and **#154** (cross-mount unconditional unlink) data-loss bugs are fixed and covered by regression tests
- [ ] the guard runs after `stat` (a missing source still raises) and no broad `except Exception` masking remains in the touched paths
- [ ] `same_backend_file(...)` is a shared helper Lanes G + J reuse

### Sub-issues
- [ ] #187 — Same-path / same-backend guard (#150, #154)

zechengz:
Closing as superseded by #721. Lane F (backend safety) is Phase 1 there, unchanged in substance. Nothing from this plan was built on main (no workspace/fork.py, no commit_batch, no StagedOp), so nothing is lost by re-homing it.

Issue #154, opened 2026-06-01
cross-mount mv: read->write->unconditional unlink has no same-backend guard (data loss when mounts alias the same storage)


zechengz:
## Summary
The live cross-mount mover `_cross_mv` in `python/mirage/workspace/executor/cross_mount.py` does:

```python
async def _cross_mv(scopes, dispatch, cmd_str):
    src, dst = scopes[0], scopes[1]
    data, _ = await dispatch('read', src)
    await dispatch('write', dst, data=data)
    await dispatch('unlink', src)      # unconditional
```

read -> write -> **unconditional unlink**, with no same-path / same-backend check. This is the same class of bug as #127 / #150, generalized across mounts.

## Why it usually does not bite
Cross-mount only triggers when the two paths are on **different mounts** (`is_cross_mount` returns true only when `len(mounts) > 1`). Two different mounts normally mean two different backing stores, so read->write->unlink is a genuine move.

## When it does bite (data loss)
The routing compares mount **prefix**, never the resolved backend identity. If two different mounts **alias the same underlying storage**, e.g.
- two `disk` mounts rooted at the same directory,
- two `ram`/`redis` mounts sharing one store,
- the same bucket/volume mounted twice,

then `mv /m1/x /m2/x` has two distinct prefixes (routes to cross-mount) but resolves to the same physical object:

```
read  m1:/x        # = the same file as m2:/x
write m2:/x        # overwrites the same object with its own bytes
unlink m1:/x       # deletes it  => GONE
```

## Where
- `python/mirage/workspace/executor/cross_mount.py` (`_cross_mv`, and `is_cross_mount` routing)

## Suggested fix
Before the unlink, resolve each scope to its backend identity (backend type + resolved key/path) and short-circuit when src and dst are the same underlying object: emit coreutils `'a' and 'a' are the same file` (exit 1) and do not unlink. `_cross_cp` should get the same guard (no data loss there, but it should error to match `cp`).

## Related
- #127 / #142 (databricks, fixed), #150 (s3-compatible). A shared same-path/same-backend guard across writable backends + cross-mount would be cleaner than the current backend-by-backend approach.

## Note
There is also a dead, unused `python/mirage/commands/builtin/cross_mount.py` (hardcoded `RESOURCES = ['ram','s3','disk']`, `_mv_cross`) with the same unguarded pattern; not imported anywhere. Worth deleting separately to avoid confusion.

Issue #150, opened 2026-06-01
S3 backend: mv/cp onto same path can delete the file on S3-compatible stores


zechengz:
## Summary

The S3 backend has the same latent same-path data-loss bug that #142 fixes for Databricks Volumes, but it is currently masked only by AWS's own behavior, not by our code.

`core.rename` for S3 is `copy_object` + **unconditional** `delete_object` (`mirage/core/s3/rename.py`), and `core.copy` issues a plain `copy_object` with **no `MetadataDirective`** (`mirage/core/s3/copy.py`).

On **real AWS S3**, copying an object onto its own key without changing metadata is rejected (`This copy request is illegal because it is trying to copy an object to itself...`), so the `delete_object` is never reached and the file survives, by accident.

But the same S3 backend is used for **R2, GCS (S3 API), Supabase, OCI, MinIO**, etc. Not all S3-compatible stores enforce that self-copy rejection. On a store that **allows** self-copy:

```
copy_object(src_key -> src_key)   # succeeds (no rejection)
delete_object(src_key)            # deletes the only copy
```

= **data loss**, the exact bug #142 fixed for Databricks.

## Reachable forms (same as #127)
- `mv a a`
- `mv ./a a`
- a glob that collapses to a single key
- `mv a dir/` when the file already lives in `dir/`

## Why it needs a fix in our code
Today S3's safety depends on the remote refusing the self-copy rather than on a guard in mirage. That is fragile and silently wrong on lenient/compatible stores.

## Proposed fix
Mirror #142: add a same-backend-path guard.
- **Core** (`mirage/core/s3/rename.py`, `copy.py`): if the resolved keys of `src` and `dst` are equal, no-op (run the guard after the existence/stat check so a missing source still raises `FileNotFoundError`).
- **Command** (`mirage/commands/builtin/s3/mv.py`, `cp.py`): emit coreutils-style `'a' and 'a' are the same file` to stderr with exit code 1.

## Notes / related
- Same class of issue also affects the cross-mount mover (`mirage/workspace/executor/cross_mount.py` `_cross_mv`): read -> write -> unconditional unlink with no same-backend guard; it is gated only by the "different mount prefix" routing condition, which breaks down when two mounts alias the same underlying storage.
- The other writable backends (disk, ssh, nextcloud, ram, redis) are structurally safe (atomic rename, or bytes captured before delete), so they do not have the data-loss exposure, though none emit the coreutils "same file" message either.
[...]

PR #213, opened 2026-06-07
fix(backend-safety): shared same-backend guard for S3 + cross-mount (#150, #154)

pzhang15:
## Overview

**Lane F — Backend safety** (#187, part of #186). Fixes two real data-loss bugs with a shared same-path / same-backend guard, mirroring the already-merged Databricks fix (#142). Zero dependencies on other in-flight work.

The lane's product is a shared helper in `core/_shared/path_guard.py` that Lanes G and J reuse:

- `same_backend_file(accessor, a, b)` — same-accessor convenience (the form the S3 commands and the Databricks template use).
- `backend_identity(accessor, path)` — the reusable primitive: resolves a path to a comparable `(backend type, store, key)` identity. Used to compare paths across **two different** accessors (the cross-mount case). It's a leaf module (no cycles): backends are recognized by accessor shape — S3-family via `config.bucket`, disk via `root`, ram/redis via `store` — and paths are normalized (`.`/`..`/`//` collapsed) before comparison.

### #150 — S3 `mv`/`cp` onto the same path can delete the file
`core/s3/rename.py` was `copy_object` + **unconditional** `delete_object`, and `core/s3/copy.py` issued a self-copy with no `MetadataDirective`. Real AWS rejects a self-copy so the delete was never reached — but lenient S3-compatible stores (R2 / GCS / MinIO / Supabase / OCI) **allow** it, so `copy_object(k -> k)` succeeds and `delete_object(k)` then destroys the only copy.

- `core/s3/rename.py`, `core/s3/copy.py`: `stat` the source, then no-op when src and dst resolve to the same key. The guard runs **after** `stat`, so a missing source still raises `FileNotFoundError`.
- `commands/builtin/s3/mv.py`, `cp.py`: emit `mv:`/`cp:` `... are the same file` to stderr with exit code 1, exactly like the Databricks command.

### #154 — cross-mount `mv` has no same-backend guard
`workspace/executor/cross_mount.py` `_cross_mv` did `read -> write -> unconditional unlink`. Cross-mount only fires when mount **prefixes** differ, but two prefixes can alias the same storage (two disk mounts on one dir, a shared ram/redis store, the same bucket mounted twice) — then `mv /m1/x /m2/x` resolves to one object: write overwrites it with its own bytes, unlink deletes it.

- `_cross_mv` / `_cross_cp` now resolve each scope to its backend identity (via the registry) and, on a same-object match, emit the coreutils error + exit 1 **without** unlinking. `_cross_cp` errors too (no data loss there, but matches `cp`). `handle_cross_mount` now receives the `registry`.

### Cleanup
- Narrowed the broad `except (FileNotFoundError, ValueError, Exception)` masking in the S3 `mv`/`cp` `_exists` helpers to `except FileNotFoundError`.
- Deleted the dead, unused `commands/builtin/cross_mount.py` (hardcoded `RESOURCES`, `_mv_cross`, same unguarded pattern, no production importer) and the single registry test that covered it.

## Tests
- `tests/core/_shared/test_path_guard.py` — normalization + the three #154 alias cases (shared disk root, shared store, same bucket twice) and negative cases (different bucket/endpoint/root/store).
- `tests/core/s3/test_rename.py`, `test_copy.py` — same-path is a no-op and the file survives (using the repo's in-memory S3 mock, whose `copy_object` **allows** self-copy so it reproduces the lenient-store data loss); missing source still raises `FileNotFoundError`; different paths still move/copy.
- `tests/commands/builtin/s3/test_mv.py`, `test_cp.py` — coreutils message + exit 1, guard fires before any mutation, different paths still dispatch.
- `tests/workspace/executor/test_cross_mount_same_backend.py` — aliased mounts: `mv`/`cp` error + the only copy survives; distinct backends still move.

Verified each regression test **fails without** the corresponding guard (data loss reproduced), and passes with it.

`tests/core/s3 tests/core/_shared tests/workspace/executor tests/commands/builtin/s3` green; full suite green except pre-existing environmental failures (`test_tac.py`, the redis-server test). `pre-commit` Python hooks clean (TS hooks fail only on `pnpm: not found`, environmental).

## Definition of Done (#187)
- [x] S3 `mv`/`cp` onto an aliased same path no longer deletes the file (#150) — regression test
- [x] cross-mount `mv` where src/dst alias the same backend path no longer unlinks the only copy (#154) — regression test
- [x] guard runs **after** stat (a missing source still raises) — test
- [x] no broad `except Exception` masking remains in the touched paths
- [x] `same_backend_file(...)` is a shared helper (in `core/_shared/`) usable by Lanes G + J


pzhang15:
Will address this issue systematically in https://github.com/strukto-ai/mirage/issues/221

PR #240, opened 2026-06-10
One declarative command-behavior spec run by both Python and TS (#151)

### LINKED ISSUE #151: [DRAFT] Proposal: one declarative spec for command behavior, run by both Python and TS
STILL IN DRAFT.

I'm still thinking about this.

## Problem

The command test suite has three issues that I realized during recent PRs:

- **Duplication.** `wc/du/ls/find/tree` are re-tested per backend (`ram/disk/redis`) with ~80% identical bodies and copy-pasted assertion literals e.g. `disk/test_du.py` and `ram/test_du.py` both assert `io.stdout.decode().strip() == "5\t/f.txt"`, verbatim. No two backends assert the *same property*: the `du`/`wc` tests variously `.strip()`, `.rstrip("\n").split("\t")`, `.split("\t")[0]`, or `endswith(b"\n")`.
- **Coverage gaps.** `md5/file/stat` have essentially no exact-output assertions, and `cmp` had none until #147, only substring/presence checks elsewhere. The PR touched ~10 commands across many backends with almost no per-backend regression net.
- **No Python↔TS parity net.** The only cross-backend/cross-language check today is `integ/check_lines.sh`, which is **substring-only** and so cannot see trailing newlines, exit codes, or stderr. The existing `integ/cases.py` harness is Python code, so TS re-implements the cases by hand and can drift.

Before: same behavior, re-asserted per backend
```
Duplication (one value, physically copied)
 disk/test_du.py   :  .strip() == "5\t/f.txt"
 ram/test_du.py    :  .strip() == "5\t/f.txt"    
 
 <- byte-identical, drifts if either edited

Inconsistency (no two backends check the same property)
 disk   wc  :  .rstrip("\n").split("\t")
 redis  wc  :  .split("\t")[0] == "3"
 check_lines.sh :  substring match
				   `- the stripping / substring styles can't see a missing
					  "\n", exit code, or stderr -- exactly the bug #137 fixed
					  
```

After: one spec, run by both languages (python and TS)
```
 commands.spec                  read by BOTH runners:
   case du_single                 python -> ram / disk / redis
   cmd    du /data/f.txt          ts     -> ram
   stdout 5\t/data/f.txt        each asserts stdout + exit + stderr, byte-exact
```

## Goals and Non-goals

Goal: A single spec read by both languages:
1. **Eliminate per-backend duplication** — define each command's expected behavior once, not once per backend.
2. **Guarantee Python↔TS parity** — one source of truth both languages are checked against, with a comparator that actually sees trailing newlines, exit codes, and stderr.

Non-goals
- This spec does not replace backend-specific tests.
- Provider-specific behavior stays in hand-written tests, including API call counts, pushdown/fallback behavior, cache invalidation, error injection, concurrency, and backend-specific semantics.

## Proposed approach

One definition per command, two runners, identical expectations so duplication and Python↔TS drift both disappear.

Components
1. A **data-only spec file** describes command behavior (seed files + cases with exact expected stdout / exit code / stderr). 
2. Two thin runners (one Python, one TS) consume the *same* file and assert byte-exact results against their respective `Workspace`. 

Side benefit, not a separate goal: an under-tested command like `md5`/`file`/`stat` gains exact-output coverage simply by adding a case.

```
                       commands.spec
              (seeds + cases, exact expected output)
                            |
            +---------------+----------------+
            v                                v
     Python runner                       TS runner
     -> ram / disk / redis               -> ram (today)
            |                                |
            +---------------+----------------+
                            v
                  byte-exact comparator
              stdout + exit code + stderr
                            |
                  addresses both goals: kills duplication + guarantees parity

  not expressible in the spec -> stays as hand-written per-backend code tests:
    call counts, error injection, cache invalidation, concurrency
```

The spec format should preserve exact bytes, including trailing newlines, empty output, stderr, and binary-ish output.
Use a small manifest plus fixture files for expected outputs, or explicit encoded fields such as:
- `stdout_text`
- `stdout_base64`
- `stderr_text`
- `stderr_base64`
- `exit`

This avoids ambiguity around delimiter lines, missing final newlines, trailing blank lines, and non-text bytes.

Rough sketch:

```
cases:
  - id: wc_default
    cmd: wc /data/a.txt
    matrix:
      python: [ram, disk, redis]
      typescript: [ram]
    expect:
      exit: 0
      stdout_text: "       2       2      12 /data/a.txt\n"
      stderr_text: ""

  - id: binary_case
    cmd: cat /data/binary.bin
    matrix:
      python: [ram, disk]
    expect:
      exit: 0
      stdout_base64: "AAECaGVsbG//vg=="
      stderr_text: ""
```

**Applicability is explicit in the spec for v1**: each case declares the backend/language matrix it should run against. The runner still validates that the selected backend actually exposes the command, but it does not auto-apply every case to every registered backend.

Initial matrix:
- Python: `ram`, `disk`, `redis` when `REDIS_URL` is set
- TypeScript: `ram`

Later, this can expand with `only`, `except`, and per-backend expected-output overrides for API-backed resources.

sonhmai:
Closes #151.

## Summary

Adds one shared command behavior spec used by Python and TypeScript.

The spec asserts exact:

- stdout bytes
- stderr bytes
- exit code

## Changes

- Add `conformance/seeds.json` for workspace fixtures.
- Add 26 cases for `cat`, `cmp`, `du`, `file`, `find`, `grep`, `head`, `ls`, `md5`, `stat`, `tree`, and `wc`.
- Run Python cases against RAM, disk, and Redis.
- Run TypeScript cases against RAM.
- Validate language and backend matrix targets at load time.
- Seed through workspace APIs instead of shell commands.
- Document how to add another backend using realistic mocked API payloads.
- Use the existing Python and TypeScript CI test jobs; no new workflow is required.

The behavioral spec stays separate from `spec/`:

- `spec/`: command definitions and options
- `conformance/`: command output and exit behavior

## TypeScript parity fix

The shared cases originally exposed missing trailing newlines across several TypeScript commands. Most of those fixes have since landed on `main`.

This PR retains the remaining `cmp` fix: default difference and EOF diagnostics now end with a newline, matching Python and Unix behavior.

## Scope

This does not replace backend-specific tests for API calls, cache invalidation, error injection, or concurrency.

## Verification

- Python conformance tests
- TypeScript conformance tests
- Python and TypeScript command tests
- TypeScript integration truth output
- Pre-commit

### DIFF
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -57,3 +57,34 @@ empty output, and binary output are all part of the contract.
 - `spec/` (command definitions exported from the registries) is a different
   artifact: it describes the command *surface*; this folder describes command
   *behavior*.
+
+## Adding a backend
+
+Add a backend when its command behavior should be compared against the same
+expectations as the existing implementations:
+
+1. Add its name to `SUPPORTED_MATRIX` in the Python and/or TypeScript runner.
+
+2. Add runner setup that constructs a fresh workspace and resets backend state
+   between cases.
+
+3. For API-backed resources, use mocked clients with payloads shaped like real
+   provider responses. Do not call live APIs from this suite.
+
+4. Add the backend to the applicable case matrices:
+
+   ```json
+   "matrix": {
+     "python": ["ram", "disk", "redis", "github"],
+     "typescript": ["ram", "github"]
+   }
+   ```
+
+5. Run the focused Python and TypeScript conformance tests.
+
+Both runners are discovered by the existing Python and TypeScript CI test jobs;
+no workflow change is required.
+
+Only add a backend to a case when both should produce the same exact stdout,
+stderr, and exit code. Keep provider-specific API calls, cache behavior, error
+injection, and concurrency in backend tests.
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -30,7 +30,7 @@ empty output, and binary output are all part of the contract.
       "matrix": { "python": ["ram", "disk", "redis"], "typescript": ["ram"] },
       "expect": {
         "exit": 0,
-        "stdout_text": "5\t5\t24\t/data/a.txt\n",
+        "stdout_text": " 5  5 24 /data/a.txt\n",
         "stderr_text": ""
       }
     }
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -65,13 +65,13 @@ expectations as the existing implementations:
 
 1. Add its name to `SUPPORTED_MATRIX` in the Python and/or TypeScript runner.
 
-2. Add runner setup that constructs a fresh workspace and resets backend state
+1. Add runner setup that constructs a fresh workspace and resets backend state
    between cases.
 
-3. For API-backed resources, use mocked clients with payloads shaped like real
+1. For API-backed resources, use mocked clients with payloads shaped like real
    provider responses. Do not call live APIs from this suite.
 
-4. Add the backend to the applicable case matrices:
+1. Add the backend to the applicable case matrices:
 
    ```json
    "matrix": {
@@ -80,7 +80,7 @@ expectations as the existing implementations:
    }
    ```
 
-5. Run the focused Python and TypeScript conformance tests.
+1. Run the focused Python and TypeScript conformance tests.
 
 Both runners are discovered by the existing Python and TypeScript CI test jobs;
 no workflow change is required.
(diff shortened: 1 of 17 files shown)


sonhmai:
@zechengz This is for testing the similarity tests between python and TS output. 
It is quite useful as it already detected some discrepancies which I also included the fix.
It keeps having conflict though. Hence can we merge this so that CI will detect any discrepancies, command parity as we add backend resources?

zechengz (approved):
SG, let me approve and merge at first. Sorry that I am doing a lot of fixes and refactorings recently.

PR #258, opened 2026-06-11
refactor: split ConsistencyPolicy into per-mount ReadPolicy and WritePolicy

zechengz:
Replaces the single `ConsistencyPolicy` (LAZY/ALWAYS) with two per-mount policies in both Python and TypeScript:

- `ReadPolicy`: `CACHED` (trust the cache) or `FRESH` (fingerprint-check the backend before serving, evict on mismatch or deletion).
- `WritePolicy`: `THROUGH` (writes and deletes reach the backend immediately). A staged (write-back) value will join once the staging overlay is implemented.

No behavior change: today's LAZY is (CACHED, THROUGH) and ALWAYS is (FRESH, THROUGH).

Internal layers (Mount, MountRegistry) now require both policies explicitly; defaults exist only at the public Workspace constructor and YAML config (`read_policy` / `readPolicy`).

### DIFF
--- a/typescript/packages/server/src/config.ts
+++ b/typescript/packages/server/src/config.ts
@@ -139,7 +139,8 @@ export interface RedisIndexBlock {
 export interface WorkspaceConfigRaw {
   mounts: Record<string, MountBlock>
   mode?: string
-  consistency?: string
+  readPolicy?: string
+  writePolicy?: string
   defaultSessionId?: string
   defaultAgentId?: string
   history?: number | null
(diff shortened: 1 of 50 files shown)


typescript/packages/server/src/config.ts
sonhmai, line 248: Could we pass readPolicy and writePolicy through here? Are they dropped, the workspace always uses the defaults?

zechengz:
bad design
close

Issue #264, opened 2026-06-11
[Feature]: Command parity for Databricks Volume backend (39 shell commands missing vs s3)


sonhmai:
### Area

Python package

### Summary

The `databricks_volume` backend currently wraps **24** shell commands, while the reference object/file backends (`s3`, `gcs`, `oci`, `nextcloud`) wrap **65**. This tracks closing that command-parity gap.

Follow-up to #61 (core resource + read/write ops are complete and shipped). This issue is scoped to the missing **command wrappers** only — the underlying read/write/stat primitives already exist.

### Currently supported (24)

`awk, cat, cp, cut, diff, find, grep, head, jq, ls, mkdir, mv, nl, rg, rm, sed, sort, stat, tail, touch, tr, tree, uniq, wc`

### Missing vs s3 (39)

**Read / inspect (10)**
`du`, `file`, `strings`, `xxd`, `md5`, `sha256sum`, `basename`, `dirname`, `realpath`, `readlink`

**Text processing (16)**
`join`, `paste`, `column`, `fold`, `expand`, `unexpand`, `fmt`, `rev`, `tac`, `look`, `shuf`, `tsort`, `comm`, `cmp`, `patch`, `iconv`

**File operations (5)**
`ln`, `tee`, `mktemp`, `split`, `csplit`

**Compression (7)**
`gzip`, `gunzip`, `zip`, `unzip`, `tar`, `zcat`, `zgrep`

**Encoding (1)**
`base64`

### Notes

- **Quick win:** `du` already has a core helper (`mirage/core/databricks_volume/du.py`) but no command wrapper — likely the smallest first PR.
- Most of these operate on file content and should compose over the existing read/write primitives the same way they do on `s3`/`nextcloud`. The `mirage-add-backend-command` skill is built for exactly this (wrap the generic helper per verb).
- A few may be **N/A** for Unity Catalog volumes and can be skipped with a note rather than forced — e.g. `ln` (no symlink support on the Files API). The implementer should confirm per-command rather than assume all 39 are viable.
- Suggested approach: land in small batches by category (read/inspect, then compression, then text processing), each with tests mirroring the existing `tests/commands/builtin/databricks_volume/` layout.

sonhmai:
Done. the feature is ready. Improvements and bug fixes will be in other issues

PR #265, opened 2026-06-11
docs(databricks): bring volume resource docs to s3 parity

sonhmai:
## What

Brings the Databricks Volume docs up to the same structure/depth as the reference object/file-storage backends (`s3`, `gcs`, `oci`, `nextcloud`). Follow-up to #61.

## Changes

- **Nav fix (`docs/docs.json`)**: add `python/resource/databricks_volume` to the "Cloud Files" group. It was the **only** one of 44 `python/resource/*` pages missing from the sidebar — previously reachable only via the resource-matrix table link.
- **Resource page (`python/resource/databricks_volume.mdx`)**: replace the one-paragraph "Supported operations" blurb with proper `## Shell Commands` tables (Read / Text Processing / File Operations / Path Utilities), scoped to the **24** commands the backend actually supports, and add a `## Use Cases` section — matching the s3/gcs/oci/nextcloud page shape. Kept the read/write primitives note and the non-atomic `mv`/`cp` caveat.
- **Stale-bug fix (`python/setup/databricks.mdx`)**: "The resource is read-only in this branch" → "Supports both read and write mount modes". Write support landed in #93.

## Deliberately not changed

- No `## Cache` section — peers have one, but the Databricks resource doesn't cache (verified in the impl), so adding it would be inaccurate.
- The separate `python/setup/databricks.mdx` page stays inline-linked (not in sidebar) — that matches 15 of 16 backend setup pages; it's the convention, not an orphan.

## Related

- Closes the docs portion of #61.
- Command-parity gap (39 commands still missing vs s3) tracked separately in #264.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -279,7 +279,8 @@
                 "group": "Cloud Files",
                 "icon": "folder-open",
                 "pages": [
-                  "python/resource/nextcloud"
+                  "python/resource/nextcloud",
+                  "python/resource/databricks_volume"
                 ]
               },
               {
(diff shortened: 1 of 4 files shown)


zechengz (approved):
Thanks :)

Issue #130, opened 2026-05-29
cross-mount cp/mv with databricks-volume: no recursion, parents not created, no move-into-dir


zechengz:
## Scope
Cross-mount = `cp`/`mv` where source and destination are on different mounts (e.g. `disk` <-> `databricks_volume`). These are intercepted at `workspace/executor/command.py:614` and handled by the generic `workspace/executor/cross_mount.py` (`_cross_cp` / `_cross_mv`), which dispatch `read` / `write` / `unlink`.

## What works
- **Single-file** `cp`/`mv` between databricks and ram/disk/s3, both directions (databricks has `read`/`write`/`unlink` ops). This is the only case covered by tests (`test_cross_mount_matrix.py` only does single-file).

## Gaps
1. **Recursive / directory cross-mount is unsupported (global).** `_cross_cp` ignores `-r` and `_cross_mv` is single-file only; both read one file -> write one file. `cp -r /disk/dir /db/` fails (read on a directory). This affects **all** pairs, not just databricks — the recursive `_cp_cross` in `commands/builtin/cross_mount.py` (`RESOURCES = ["ram","s3","disk"]`) is dead code, shadowed by the executor. It is newly relevant because #116 added within-mount recursive `cp`/`mv`/`rm` for databricks, so users will reasonably expect disk<->databricks `cp -r` to work too.
2. **databricks-specific: parent dirs are not auto-created.** `core/databricks_volume/write.py` (`write_bytes` -> `_ensure_parent_directory_sync`) only *verifies* the parent exists and raises `FileNotFoundError` otherwise; it does not `mkdir -p`. So even a single-file `mv /disk/a.txt /db/new/sub/a.txt` fails when `/db/new/sub` does not exist, whereas disk auto-creates parents and s3 has no real directories. The cross-mount handler does not create parents either.
3. **No move-into-directory semantics.** `_cross_cp`/`_cross_mv` write to the destination path verbatim, so `mv /disk/a.txt /db/dir/` (where `dir` exists) clobbers the directory as a file instead of writing `/db/dir/a.txt`. (Within-mount cp/mv handle this via `child_path`; the cross path does not.)
4. **databricks absent from the `RESOURCES` list** in `commands/builtin/cross_mount.py` (cosmetic — that path is currently unused/shadowed).

## Suggested direction
- Extend the cross-mount handler to support directories (walk the source tree, create dirs on the destination mount, copy per file) — benefits every resource, not only databricks.
[...]

PR #116, opened 2026-05-28
feat(databricks-volume): add directory/structural write ops (mkdir, rmdir, rm -r, cp, mv)

sonhmai:
## Summary

Follow-up of #109. Part of #61

Adds: `mkdir`, `mkdir -p`, `rm -r`, `rm -d`, `cp`, `cp -r`, `mv`, plus the `mkdir`/`rmdir`/`rename` ops.

## Some correctness notes

- **`create_directory()` is idempotent + mkdir-p**, so `mkdir(parents=False)` pre-checks target-absent (`FileExistsError`) **and** parent-exists-as-dir (`FileNotFoundError`/`NotADirectoryError`) before calling the SDK.
- **`delete_directory()` is empty-only**; `rm_recursive` is a Mirage tree walk (files first, dirs deepest-first, root last). The fake `FakeFiles` test harness models this.
- **Rename is non-atomic** (no native Files API rename): `stat(src) → copy(src,dst) → unlink/rm_recursive(src)`. Documented in the core module and prompt.

## Test plan

- [x] `cd python && uv run pytest tests/core/databricks_volume tests/resource/databricks_volume tests/commands/builtin/databricks_volume` 
- [x] `tests/workspace/test_dispatch_write.py` + `tests/workspace/test_index_integration.py` 
- [x] databricks volume integration tests (locally)

PR #268, opened 2026-06-11
feat(cross-mount): support recursive cp/mv across mounts (#130)

sonhmai:
Closes the re-scoped parts of #130 (claim 3 was already fixed by #201).

## Problem

Cross-mount `cp`/`mv` (operands spanning two mounts) handled **only single files**: `_cross_cp`/`_cross_mv` honored only `-n`, dropped `-r`, and read each source as one file. So `cp -r /disk/dir /dbx/` failed (read on a directory) and directory `mv` was unsupported. This affected every mount pair, and became more visible once #116 added within-mount recursive cp/mv/rm for databricks.

## Changes

- **Recursive walk (executor)** — `cp -r` and directory `mv` now walk the source tree, create destination directories top-down, and copy per file. Directory `mv` then removes the source tree via a universal `readdir`+`unlink`+`rmdir` walk (`rm_recursive` is only registered on some backends).
- **POSIX semantics (issue option B)** —
  - `cp` without `-r` omits directories with the coreutils message (`cp: -r not specified; omitting directory '…'`) and still copies file args.
  - A single-file write into a **missing parent** stays strict (`FileNotFoundError`), matching coreutils — **not** `mkdir -p`. Recursive copy creates the destination subtree it mirrors, which is what coreutils `cp -r` does.
  - (Deliberately did not change `disk`/`databricks` `write` semantics. `disk` auto-creates parents on write, which is non-POSIX; reconciling that is noted below as a follow-up, not done here.)
- **Claim 4 (cosmetic)** — add `databricks_volume` to the `RESOURCES` list in `commands/builtin/cross_mount.py`. Note: that module's `CROSS_COMMANDS` is never registered — it's shadowed by the executor's `handle_cross_mount` — so this is purely for list consistency; full removal of the dead module is a separate cleanup.

## Tests

- `tests/workspace/test_cross_mount_recursive.py` (RAM↔RAM): `cp -r`, copy-into-existing-dir, `cp` without `-r` omits + writes nothing, directory `mv` removes source, `cp -rn` no-clobber.
- `tests/commands/builtin/databricks_volume/test_cross_mount.py` (databricks↔RAM): `cp -r` both directions (ram→databricks exercises databricks `mkdir` during the walk), directory `mv` removes the databricks source, and the POSIX single-file missing-parent error.
- `tests/integration/test_cross_mount_matrix.py`: recursive `cp`/`mv` added across the ram/disk/s3/gdrive matrix.

## Known follow-ups (out of scope)

- **gdrive recursive-source** listing returns nested children only after an `ls`-primed index; the recursive matrix tests skip gdrive-as-source. Tracked for a separate fix.
- **disk non-POSIX parent auto-create** (`core/disk/write.py`) diverges from coreutils; making it strict is a separate change.
- Cache invalidation for `mkdir` on indexed remote dests during the walk relies on per-file write invalidation; empty-dir-only cases may want explicit invalidation.

### DIFF
--- a/python/mirage/commands/builtin/cross_mount.py
+++ b/python/mirage/commands/builtin/cross_mount.py
@@ -210,7 +210,7 @@ async def _paste_cross(
     return output.encode(), IOResult()
 
 
-RESOURCES = ["ram", "s3", "disk"]
+RESOURCES = ["ram", "s3", "disk", "databricks_volume"]
 _RESOURCE_PAIRS = list(product(RESOURCES, repeat=2))
 
 _CROSS_FNS = [
--- a/python/mirage/workspace/executor/cross_mount.py
+++ b/python/mirage/workspace/executor/cross_mount.py
@@ -49,8 +49,10 @@ async def handle_cross_mount(
     Copy and move follow POSIX operand semantics: every path except the final
     path is a source, and the final path is the destination. An existing
     destination directory maps each source to ``destination/basename``;
-    multiple sources require that directory form. Source data is read before
-    mutation so validation or read failures do not partially modify targets.
+    multiple sources require that directory form. Flat-file sources are read
+    before mutation so a read failure does not partially modify targets;
+    directory sources (``cp -r``, directory ``mv``) walk the tree, create
+    destination directories, and copy per file.
 
     Examples:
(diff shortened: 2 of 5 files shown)


zechengz (approved):
Thanks! :)

Issue #152, opened 2026-06-01
databricks-volume: recursive cp -r/mv dir into its own subtree bypasses the same-path guard


zechengz:
## Summary
Follow-up to #127 / #142. The same-path guard added in #142 compares **exact backend keys** (`backend_path(src) == backend_path(dst)`). That catches a file copied onto itself, but **not** a directory copied into its own subtree, because the destination key differs by the appended basename.

## Reproduce
```
cp -r /dbx/dir /dbx/dir
```
- `src    = /dbx/dir`
- `dst    = /dbx/dir` (exists, is a directory)
- `target = /dbx/dir/dir`  (basename of src appended into the dir destination)

Guard check: `backend_path('/dbx/dir') == backend_path('/dbx/dir/dir')` -> **false**, so the guard does not fire. `copy_core(/dbx/dir, /dbx/dir/dir, recursive=True)` then runs `_copy_tree_sync`, which creates `/dbx/dir/dir` and copies `dir`'s entries into it, i.e. copies the directory into its own subtree (redundant/nested copies, potential runaway recursion depending on listing enumeration).

`mv -r` / `mv dir dir` reaches the same path.

## Expected (coreutils)
```
cp: cannot copy a directory, '/dbx/dir', into itself, '/dbx/dir/dir'
```
exit 1, no writes.

## Where
- `python/mirage/core/databricks_volume/copy.py` (`copy` / `_copy_tree_sync`)
- `python/mirage/commands/builtin/databricks_volume/{cp,mv}.py` (`same_backend_file` only checks exact equality)

## Suggested fix
Replace the exact-key equality with a subtree-aware check: target is inside (or equal to) the source tree, i.e.
```
target_key == src_key or target_key.startswith(src_key.rstrip('/') + '/')
```
and emit the coreutils "cannot copy a directory into itself" diagnostic (exit 1) before any write.

## Also fold in (minor, same area)
Command-layer ordering nit: the same-path guard in `cp.py`/`mv.py` runs **before** the source-existence check, so `cp /dbx/missing /dbx/missing` prints `'missing' and 'missing' are the same file` instead of the GNU/POSIX-correct `cannot stat 'missing': No such file or directory`. Exit code is already correct (1); only the message diverges. The core layer is already correct (raises `FileNotFoundError`). Fix = run the same-path guard after the existence check (or let core's error propagate).

## Context
[...]

PR #273, opened 2026-06-11
fix(databricks_volume): reject recursive cp/mv into own subtree

### LINKED ISSUE #141: databricks-volume: recursive cp/mv into own subtree bypasses same-path guard

sonhmai:
Closes #141.

## Problem

A recursive `cp`/`mv` of a directory into its own subtree bypassed the exact-key same-path guard. `cp -r /dbx/d /dbx/d` resolves the directory target to the child `/dbx/d/d`, so `same_backend_file('/dbx/d', '/dbx/d/d')` is false and nothing fired. `_copy_tree_sync` creates `remote_dst` **before** listing `remote_src`, so the freshly-created child appears in the source listing and the walk descends into it forever — `RecursionError` against the mocked Files API, unbounded `create_directory` + uploads against real Databricks.

`mv` was worse: `rename` runs `copy(recursive=True)` then `rm_recursive(src)`, so after the runaway copy it recursively **deletes the source** — data loss.

## Fix

**Core (authoritative — covers `cp`, `mv`, and `ops.rename`, which all reach core directly):**
- `core/databricks_volume/copy.py` and `rename.py` refuse the recursive/dir branch **before any write** when the destination key is nested under the source key (`dst == src` or `dst.startswith(src + "/")`). `rename` checks before the copy+delete, so the source is untouched.

**Commands:**
- `cp.py` / `mv.py` render the coreutils messages:
  - `cp: cannot copy a directory, '<src>', into itself, '<dst>'`
  - `mv: cannot move '<src>' to a subdirectory of itself, '<dst>'`
- The same-path guard now runs **after** the source-existence check, so `cp missing missing` reports `cannot stat '<src>': No such file or directory` instead of `are the same file` (folds in the #152 ordering nit).

Sibling/outside recursive copies still work; prefix-siblings (`d` → `dd`) are not falsely flagged (the trailing `/` in the guard prevents that).

**TypeScript parity:** mirrored the core guards in `typescript/.../databricks_volume/copy.ts` and `rename.ts` (same data-loss path: `rename` → recursive `copy` → `rmRecursive`).

## Tests

- Core (py): `cp -r d d/…` and `mv d d/…` raise with zero `create_directory`/`upload`/`delete`; source intact for `mv`.
- Commands (py): into-itself messages for `cp -r`/`mv`, plus `cp/mv missing missing` → `cannot stat`.
- TS: equivalent core tests asserting no `PUT`/`DELETE` before the guard.

Python: 135 databricks_volume tests pass. TS: 12 `copy_rename` tests pass.

## Out of scope (tracked for #187)

The shared **generic** `cp` handler (`generic/cp.{py,ts}`, used by disk/s3/nextcloud) snapshots `find()` into a list before copying, so `cp -r d d` there yields a *bounded* spurious nested copy rather than a runaway — a milder coreutils divergence. The TS databricks `cp` command also routes through this generic handler (unlike Python's bespoke `cp.py`). A shared `target_within_source` guard in the generic handler would close these in both languages; that belongs to the cross-backend Fsafe umbrella (#187), per this issue's open question.

### DIFF
--- a/python/mirage/commands/builtin/databricks_volume/_helpers.py
+++ b/python/mirage/commands/builtin/databricks_volume/_helpers.py
@@ -39,6 +39,13 @@ def same_backend_file(accessor: DatabricksVolumeAccessor, a: PathSpec,
     return backend_path(accessor.config, a) == backend_path(accessor.config, b)
 
 
+def target_within_source(accessor: DatabricksVolumeAccessor, src: PathSpec,
+                         dst: PathSpec) -> bool:
+    src_key = backend_path(accessor.config, src)
+    dst_key = backend_path(accessor.config, dst)
+    return dst_key.startswith(src_key + "/")
+
+
 def read_bytes_with_index(index: IndexCacheStore | None,
                           prefix: str = "") -> Callable:
     return partial(call_read_bytes, _read_bytes, index=index, prefix=prefix)
--- a/python/mirage/commands/builtin/databricks_volume/cp.py
+++ b/python/mirage/commands/builtin/databricks_volume/cp.py
@@ -16,8 +16,8 @@ from functools import partial
 
 from mirage.accessor.databricks_volume import DatabricksVolumeAccessor
 from mirage.cache.index import IndexCacheStore
-from mirage.commands.builtin.databricks_volume._helpers import \
-    same_backend_file
+from mirage.commands.builtin.databricks_volume._helpers import (
+    same_backend_file, target_within_source)
 from mirage.commands.builtin.utils.copy import (copy_targets, is_directory,
                                                 path_exists)
 from mirage.commands.registry import command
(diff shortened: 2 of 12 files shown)


zechengz (approved):
LGTM

Issue #283, opened 2026-06-12
Cache per-spec parser lookup sets (parse_command builds them on every call)


zechengz:
## Observation (from PR #261 review)
`parse_command` builds 5 sets/dicts from `spec.options` (bool_flags, value_flags, long variants, repeat_flags, value_flag_kinds) on every invocation, and the executor builds `repeat_path_keys` per dispatch. Every shell command parses at least twice (word classification via `classify_argv_by_spec` + executor `_parse_flags`), so this is ~2× O(options) set construction per command on the hot path.

## Proposal
Specs are immutable; precompute these lookups once per `CommandSpec` (cached property or built at construction) in both languages. Same for the path normpath dedup between the classification parse and the executor parse if it ever shows up in profiles.

Low priority — option counts are ~20 and this has never shown in a profile; do it when touching the parser next, not as its own PR.

zechengz:
The parser half of this landed with #676. `compile_spec` is `@lru_cache(maxsize=512)` and `parse_command` calls it, so the five lookup tables are built once per `CommandSpec` rather than per invocation:

`python/mirage/commands/spec/compile.py:144`

TypeScript has the same through `compileSpec`, pinned by an identity test (`expect(compileSpec(spec)).toBe(compileSpec(spec))`).

Still open is the second half this issue named, the executor's per-dispatch sets. `flags.py` rebuilds two of them from `spec.options` on every dispatch:

- `python/mirage/workspace/executor/command/flags.py:99` (`repeat_path_keys`)
- `python/mirage/workspace/executor/command/flags.py:105` (`single_path_keys`)

Both derive only from immutable spec data, so they belong on `CompiledSpec` beside the rest.

Keeping this open scoped to that, and keeping the original guidance: worth doing when the parser is next touched, not as its own PR.

zechengz:
Done. `commands/spec/compile.py` lowers a `CommandSpec` into a frozen `CompiledSpec` holding every lookup table the parser walks (bool/value/attach spellings, `kind_of`, `dest`, `multiple_dests`, ...), built behind `@lru_cache(maxsize=512)` — so the tables are built once per spec instead of on every `parse_command` call. Exactly the proposal here.

PR #284, opened 2026-06-12
refactor: grep-family review fixes (shared resolve_pattern, TS -f dedup, general grep conversion)

zechengz:
Follow-up to #261 from its post-merge code review:
- `resolve_pattern` moves from `generic/grep.py` to `grep_helper.py`, removing the rg/zgrep → grep cross-generic import.
- TS: the triple-duplicated `-f` pattern-file read block in grepGeneric/rgGeneric/zgrepGeneric extracted to a shared `resolvePatternFromFlags` in `grep_helper.ts`.
- `general/grep.py` (unregistered fallback, dead code) converted to the flags-dict contract so it is no longer a latent crash if registered (`-e` now arrives as a list).
- Filed #283 for the per-parse spec lookup-set caching observed during review.

Behavior unchanged: full suites green both languages; shared integ byte-identical (py ram + TS ram verified).

### DIFF
--- a/python/mirage/commands/builtin/general/grep.py
+++ b/python/mirage/commands/builtin/general/grep.py
@@ -17,9 +17,11 @@ from collections.abc import AsyncIterator
 
 from mirage.commands.builtin.grep_context import grep_context_lines
 from mirage.commands.builtin.grep_helper import (NEVER_MATCH, compile_pattern,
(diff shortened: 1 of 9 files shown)


zechengz:
Recreating from a clean branch off main (#261 was squash-merged, so this branch's diff shows stale history).

PR #285, opened 2026-06-12
refactor: delegate py databricks_volume to generics, add coreutils guards to generic cp/mv (py+ts)

### LINKED ISSUE #274: Generic cp/mv: refuse copying/moving a directory into its own subtree (all backends)
Follow-up to #141 / PR #273, which fixed the runaway recursion and mv data loss in databricks_volume's bespoke tree copy. The shared generic handlers still diverge from coreutils:

- `generic/cp.py` and `generic/cp.ts`: `cp -r d d` produces a bounded spurious nested copy (`d/d/...`) instead of refusing with `cp: cannot copy a directory, 'd', into itself, 'd/d'`. Affects every backend wired to the generic cp (disk, s3, nextcloud, ssh, gdrive, ...) and the TS databricks_volume cp command, which routes through the generic handler (only direct `ops.copy` is guarded by #273).
- ram and redis core `rename`: `mv d d/sub` silently relocates the tree under the nested path instead of refusing with `mv: cannot move 'd' to a subdirectory of itself, 'd/sub'`. No data loss (snapshot iteration keeps it bounded), but wrong semantics.
- s3 `mv d d/sub` errors, but with a backend error rather than the coreutils message. disk/ssh/nextcloud are rejected by the underlying OS/SFTP/WebDAV, also without the coreutils message.

## Fix

Add a shared `target_within_source` guard (dst key == src key or starts with `src + "/"`) to the generic cp/mv handlers in both languages, before any write. Reuse/lift the helper introduced in #273 and consolidate the now-duplicated `startswith(src + "/")` checks (py command helper, py core copy/rename, ts core copy/rename).

Distinct from #187, whose scope is exact same-path aliasing (same_backend_file), not the nested-subtree case.

zechengz:
Closes #274.

## What

**py databricks_volume: 9 commands delegated to generics** (cat, stat, ls, tree, find, grep, rg, cp, mv), 12/24 -> 21/24, matching TS. Only mkdir/rm/touch stay bespoke (backend-specific directory semantics, same as s3 and TS). ~900 lines of duplicated engine code removed. Side fixes for free: `cat -n` uses GNU width-6 numbering, `find` gains -size/-mtime/-iname/-path/-mindepth.

**generic cp/mv guards, both languages.** Three per-source guards in coreutils order, before any write: missing source (`cannot stat`, continue with remaining sources), same file, directory into its own subtree. Errors accumulate on stderr with exit 1; `cp -v` output is now GNU-quoted. Guards compare backend storage keys via an injectable `backend_key` (default: mount-relative path; databricks passes `backend_path`). All 13 backend cp/mv wirings inherit the fix with no per-backend code: py ram/disk/redis/s3/ssh/nextcloud/databricks_volume, ts ram/s3(+aliases)/databricks_volume/disk/redis/ssh/opfs.

This is the command-layer half of #150 (`mv a a` no longer reaches the unguarded copy+delete rename). The core-layer s3/cross-mount guard remains #187. Recursive `cp -r` on databricks now fans out via find like every generic backend (empty subdirectories are not copied, the pre-existing generic-wide tradeoff).

## Tests

- generic guard unit tests (py + ts), ram E2E tests, 7 new databricks command test files (previously zero coverage for those commands)
- integ: 8 guard cases appended to the shared suite with dedicated /data/guard seeds; truth.txt regenerated and verified byte-identical on 9 runners locally (py ram/disk/ssh-mock/s3-moto/redis, ts ram/disk/opfs/redis)
- full suites green: py 7346, ts core 3040, ts node 1474; pre-commit clean

### DIFF
--- a/integ/cases.py
+++ b/integ/cases.py
@@ -72,6 +72,10 @@ SEED_FILES = {
     "world\nbar\n",
     "/data/patterns2.txt":
     "baz\n",
+    "/data/guard/g.txt":
+    "guard\n",
+    "/data/guard/sub/s.txt":
+    "inner\n",
 }
 
 CASES: list[tuple[str, str]] = [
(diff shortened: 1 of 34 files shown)


zechengz:
@sonhmai made your fixes to generic to other backends such as ram, disk, s3 etc
when you have time feel free to test it again for databricks volume and let me know if there are some errors :)

PR #293, opened 2026-06-12
feat(databricks-volume): add Python token provider authentication

sonhmai:
## What changed

- Separate Databricks volume location config from runtime authentication.
- Add public `TokenProvider`, `StaticTokenProvider`, and `DatabricksProfileTokenProvider` APIs.
- Route core volume operations through a Mirage-owned async Files client instead of exposing the Databricks SDK client.
- Resolve tokens per Files API operation, including deferred stream reads.
- Require a complete resource override when loading Databricks volume snapshots because token providers are not serialized.
- Update Python tests, examples, and docs for the new API.

## Why

Passing a fixed token into SDK config can conflict with OAuth credentials injected by Databricks Apps. It also prevents cached resources from using rotating or request-scoped OBO tokens.

This keeps authentication under application control and gives Mirage an internal client boundary that can move from the Databricks SDK to HTTP later without changing the public resource API.

## Scope

Python implementation only. TypeScript parity remains tracked in #289.

## Verification

- `PYTHONPATH=. /Users/son/code/mirage/python/.venv/bin/python -m pytest tests/core/databricks_volume tests/resource/databricks_volume tests/resource/test_state_round_trip.py -q --no-cov`
- Python pre-commit hooks
- Codex and Claude autoreview; accepted findings for SDK multipart uploads, deferred stream authentication, and profile config reuse were fixed

Issue #289, opened 2026-06-12
databricks_volume: redesign auth and client boundary for rotating tokens (py+ts)


sonhmai:
## Summary

Redesign the Databricks Volume resource public API so authentication and Files API transport are Mirage-owned abstractions rather than Databricks SDK types.

The immediate bug is that an explicit token conflicts with service-principal OAuth variables injected by Databricks Apps:

```text
ValueError: validate: more than one authorization method configured: oauth and pat
```

The larger problem is architectural: the accessor currently constructs and exposes a Databricks SDK `WorkspaceClient`, stores a token value at construction time, and makes the public resource API depend on SDK authentication behavior.

## Design goals

1. Replacing the Databricks SDK with `httpx`/`fetch` later should not require application changes.
2. Python and TypeScript should expose the same public authentication and client model.
3. Callers should control how tokens are obtained, cached, rotated, and refreshed.
4. A cached resource or `Workspace` must not retain an expired token snapshot.
5. Mirage should own Files API operations, applying credentials to requests, state behavior, and test doubles.
6. Databricks SDK authentication types must not appear in the public Mirage API.

## Current behavior

Python:

- `DatabricksVolumeAccessor.client` lazily constructs an SDK `WorkspaceClient`.
- `DatabricksVolumeConfig.token` is copied into SDK config.
- SDK environment loading can combine that token with injected OAuth credentials and reject the configuration.
- The public `client=` escape hatch accepts an SDK client.

TypeScript:

- The accessor stores a static token at construction.
- `dbxFetch` uses the same token for every request.
- A cached resource eventually fails with `401` after that token expires.

The two implementations therefore differ internally and neither provides a transport-independent rotating credential boundary.

### Python

```python
from collections.abc import Awaitable
from typing import Protocol


class TokenProvider(Protocol):
    def get_token(self) -> str | Awaitable[str]:
        ...


resource = DatabricksVolumeResource(
    DatabricksVolumeConfig(
        host="https://example.cloud.databricks.com",
        catalog="main",
        schema="default",
        volume="documents",
        root_path="/",
    ),
    token_provider=provider,
)
```

### TypeScript

```typescript
export interface TokenProvider {
  getToken(): string | Promise<string>;
}

const resource = await DatabricksVolumeResource.create(config, {
  tokenProvider: provider,
});
```

`TokenProvider` is a Mirage interface. It does not expose SDK configuration, OAuth grants, HTTP headers, `httpx`, or `fetch`.

Provider contract:

[...]

PR #301, opened 2026-06-13
fix(databricks-volume): prefer explicit token authentication

sonhmai:
## What changed

- Tell the Databricks SDK to use `pat` authentication when `DatabricksVolumeConfig.token` is explicitly set.
- Leave profile and environment authentication selection unchanged when no token is provided.
- Add focused coverage for both paths.

## Why

Databricks Apps can inject service-principal OAuth environment variables. When Mirage also passes an explicit token, the SDK detects both `oauth` and `pat` and rejects the client configuration:

```text
ValueError: validate: more than one authorization method configured: oauth and pat
```

Selecting `pat` for the explicit-token path makes the caller-provided token authoritative without changing Mirage's public API.

## Scope

This is the minimal compatibility fix. The broader token-provider redesign remains in #293 and is tracked by #289 which might be in future release.

## Verification

- `PYTHONPATH=. /Users/son/code/mirage/python/.venv/bin/python -m pytest tests/core/databricks_volume tests/resource/databricks_volume -q --no-cov`
- Python pre-commit hooks
- Direct SDK validation reproduction for explicit token plus injected OAuth credentials
- Structured autoreview: clean

### DIFF
--- a/python/mirage/accessor/databricks_volume.py
+++ b/python/mirage/accessor/databricks_volume.py
@@ -46,6 +46,7 @@ class DatabricksVolumeAccessor(Accessor):
                 "host": self.config.host,
                 "token": self.config.token,
                 "profile": self.config.profile,
+                "auth_type": "pat" if self.config.token is not None else None,
                 "http_timeout_seconds": self.config.timeout,
             }
             sdk_config = WorkspaceConfig(**{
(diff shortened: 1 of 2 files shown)


sonhmai:
A small fix for the current public api of databricks volume resource so that it's at least ready for this release.
Intent: bigger refactor can wait to next release.
cc @zechengz

PR #124, opened 2026-05-28
feat(gdrive): surface Shared Drives as top-level directories

sevdeutsch:
Closes #121.

## Problem
The `gdrive` resource lists **My Drive** only — `list_files` queries `'<folder_id>' in parents` with no `corpora`/`driveId`, so files in Shared Drives are invisible.

## Change
- Add `list_shared_drives()` and thread an optional `drive_id` through `list_files` (sets `corpora=drive` / `driveId` / `includeItemsFromAllDrives` / `supportsAllDrives` when present).
- Include `driveId` in the file fields and carry it on the cached `IndexEntry.extra` so nested listings inside a Shared Drive query the right corpus.
- Surface visible Shared Drives as top-level directories alongside My Drive.
- **Best-effort:** if Shared Drives can't be enumerated (missing scope / API error), root listing still returns My Drive. (Flagged for your review — happy to make it strict instead.)

Default behavior (My Drive only) is unchanged when no `drive_id` is given / no Shared Drives exist.

## Tests
`tests/core/gdrive` + `tests/core/google`: **58 passed**. Added:
- shared drives appear at root and the drive id is carried for nested listings
- best-effort fallback when enumeration raises
- updated existing mocks for the new `drive_id` kwarg

## Context
Second of three generic improvements we carry downstream on a Context Engine built on Mirage. Companion: #123 (for #122); synthetic mount-parents (#120) to follow.

PR #307, opened 2026-06-13
fix(gdrive): complete Shared Drive support (Python + TypeScript)

sonhmai:
Completes the Shared Drive work started in #124, which surfaced Shared Drives in listings but left reads, stats, and streamed cat broken against them. Picks up where that PR stalled and brings the TypeScript core to parity.

## What this fixes
- **Shared Drive read/stat/cat**: files inside a Shared Drive now resolve and download. Listing inside a Shared Drive threads its `driveId` through the index so the Drive API query targets the right corpus.
- **`supportsAllDrives=true`** added to `getFileMetadata`, `downloadFile`, `downloadFileStream`, and `deleteFile`, per Google's Drive API guidance for cross-drive operations.
- **Duplicate drive names / cache collisions**: duplicate Shared Drive names get a `[Shared Drive]` (then numeric) suffix so vfs paths stay unique.
- **Best-effort enumeration**: if Shared Drives can't be listed (missing scope, API error), My Drive still lists; the failure is logged, not swallowed silently.
- **Directory classification centralized** in `DIRECTORY_RESOURCE_TYPES` so `read`/`stat`/`stream` treat a Shared Drive root as a directory.

## Python / TypeScript parity
The original change was Python-only. This PR mirrors the full behavior into `typescript/packages/core` (`google/drive.ts`, `core/gdrive/{readdir,read,stat}.ts`, prompt), and adds the previously-absent `drive.test.ts` and `stat.test.ts` alongside readdir/read coverage matching the Python tests.

## Verification
- Python: `gdrive` + `google/drive` suites pass.
- TypeScript: 43 gdrive/drive tests pass; eslint + prettier clean.

Closes #124

### DIFF
--- a/docs/python/resource/gdrive.mdx
+++ b/docs/python/resource/gdrive.mdx
@@ -36,6 +36,9 @@ ws = Workspace({"/gdrive": resource}, mode=MountMode.READ)
     <name>.gdoc.json
     <name>.gsheet.json
     <name>.gslide.json
+  <shared-drive>/
+    <file>
+    <folder>/
 ```
 
 Example:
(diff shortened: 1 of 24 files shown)


zechengz (approved):

sonhmai:
Two design observations are deliberately deferred out of this PR's scope and tracked separately:
- #308 — should `gdocs`/`gsheets`/`gslides` mounts also surface Shared Drive content? (`list_all_files` lacks the all-drives flags; open product question)
- #309 — Shared Drive enumeration cache-poisoning on transient failure, and making the read/delete boundary explicit.

Issue #335, opened 2026-06-16
TS glob resolution returns absolute matched paths after repeated globs (index-cache state)


zechengz:
## Summary

In the TypeScript implementation, resolving a glob produces matched `PathSpec`s whose `.original` form is **inconsistent depending on the index-cache state**. After repeated glob/readdir operations on the same directory, a later glob returns matches in a form that no longer round-trips through `handleCommand`'s argv→PathSpec recovery (`scopeMap.get(value)` in `typescript/packages/core/src/workspace/executor/command.ts`), so the recovered spec is rebuilt fresh and loses metadata.

This is a latent bug independent of any display work — it was surfaced while attempting GNU-aligned relative-glob display (see linked enhancement).

## Reproduction (minimal)

```ts
const ws = new Workspace({ '/data': new RAMResource() }, { mode: MountMode.WRITE })
await ws.execute('mkdir -p /data/disptree')
await ws.execute('tee /data/disptree/x.txt >/dev/null', { stdin: enc('nested\n') })
await ws.execute('tee /data/disptree/z.txt >/dev/null', { stdin: enc('zzz\n') })

await ws.execute('(cd /data/disptree && wc -l *.txt)')   // 1st glob — ok
await ws.execute('(cd /data && wc -l disptree/*.txt)')   // 2nd — ok
await ws.execute('(cd /data && wc -l disptree/*.txt)')   // 3rd — matched paths now come back absolute
```

The first one or two globs behave one way; a subsequent glob on the same directory returns matches in a different `.original` form. Bisecting the integ suite pinned the trigger to a prior glob case on the same subtree (glob-after-glob), not warm cache from `find`/`ls` alone.

## Expected

A glob match's `.original` should be a stable, canonical absolute path regardless of cache warmth, so downstream argv recovery is deterministic.

## Notes / pointers

- Python is **not** affected (consistent cold + warm). Python had a related but separate issue: disk's glob joined dir+entry with a stray double slash (`/data/disptree//x.txt`); that was fixed by normalizing the matched original in `mirage/workspace/node/resolve_globs.py`. The TS form-variance is a different cause (cache-state-dependent), still open.
- Likely lives in the per-resource `glob()` / readdir-cache path and/or `resolveGlobs` in `typescript/packages/core/src/workspace/node/resolve_globs.ts`.
[...]

PR #344, opened 2026-06-17
fix: normalize awaitable read_stream in generic commands (gdrive cold crash)

zechengz:
## What

On a **cold cache**, every gdrive generic-routed streaming command crashed:

```
TypeError: 'async for' requires an object with __aiter__ method, got coroutine
```

gdrive's `core/gdrive/stream.py:read_stream` is a plain `async def` resolving to `bytes` (for gdoc/gsheet/gslide) or an `AsyncIterator`, unlike every other backend whose `read_stream` is an async generator. The generic commands called `read_stream(accessor, path)` **without awaiting** and handed the coroutine to `async for`. A warm cache routes through the generic RAM path and masks it, so it only reproduces cold.

## Fix

Route `read_stream` through the existing `_open_read_stream` helper (awaits awaitables, wraps bytes) — the same normalization `uniq` already uses. Applied to: cut, grep, rg, awk, tr, split, jq, base64, tee, sha256sum, nl, xxd, tac.

## Verification

- Regression test feeds a gdrive-style (bytes-returning) `read_stream` to `cut` (assigned form) and `sha256sum` (inline `async for` form); confirmed it **fails without** the fix (reproduces the exact crash) and passes with it.
- 486 generic + streaming tests green; pre-commit clean.
- TypeScript is unaffected (its gdrive stream is already an `AsyncIterable`); this is the py-side of the cold-vs-warm dispatch split (TS-only sibling is #335).

### DIFF
--- a/python/mirage/commands/builtin/generic/awk.py
+++ b/python/mirage/commands/builtin/generic/awk.py
@@ -7,7 +7,8 @@ from mirage.commands.builtin.generic.awk_types import (CMP_OP_PATTERN,
                                                        PRINT_STMT, AwkBlock,
                                                        AwkBoolOp, AwkBuiltin,
                                                        AwkCmpOp)
-from mirage.commands.builtin.utils.stream import _resolve_source
+from mirage.commands.builtin.utils.stream import (_open_read_stream,
+                                                  _resolve_source)
 from mirage.io.async_line_iterator import AsyncLineIterator
 from mirage.io.types import ByteSource, IOResult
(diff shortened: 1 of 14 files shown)


zechengz:
Closing in favor of #348. Audit showed the 13-command `_open_read_stream` spread here guards a reader shape no backend actually produces (every backend wraps its reader in an async generator before passing it to a generic). The one real inconsistency, github/uniq passing a raw reader, is fixed properly in #348, which also deletes the now-unused helper.

PR #345, opened 2026-06-17
fix(fuse): keep caller-owned mountpoints

sonhmai:
## Summary

Fixing bug: 
- Mirage treats the FUSE mountpoint directory as if Mirage always owns it. Cleanup could delete caller-owned FUSE mountpoint directories
- In TypeScript it could recursively delete through a still-mounted FUSE filesystem if unmount failed.

Removal is only true for auto temp mounts, it created that dir, so unmount may remove the empty dir
```
fuse=True -> /tmp/mirage-abc123
```
Wrong for fixed/ deployment mounts `mountpoint="/mirage/mnt"` (caller chooses the mount path instead of Mirage making a random dir.
```
Workspace(resources, fuse=FuseConfig(mountpoint="/mirage/mnt"))
```

Current bad clean up behavior
```
close()
  -> unmount
  -> delete /mirage/mnt
```
Should be
```
close()
  -> unmount
  -> leave /mirage/mnt alone
```

How?
- Track whether a FUSE mountpoint was generated by Mirage or supplied by the caller.
- Keep caller-owned mountpoint directories after unmount/close.
- Remove generated mountpoint directories only with empty-directory removal, never recursive deletion.
- Add Python and TypeScript regression coverage for caller-owned and generated mountpoint cleanup.

## Tests

- `cd python && uv run pytest tests/workspace/test_fuse.py`
- TS test
- precommit

### DIFF
--- a/integ/cli_fuse.sh
+++ b/integ/cli_fuse.sh
@@ -1,9 +1,12 @@
 #!/usr/bin/env bash
 # CLI end-to-end FUSE parity. Drives the daemon-backed CLI to create a workspace
-# with TWO per-mount FUSE subtrees pinned to known OS paths, writes content
-# through the VFS, then reads it back THROUGH the kernel mountpoints. The daemon
-# hosts the mounts in its own process, so the reading shell never deadlocks.
-# Proves the CLI really mounts each subtree at its own path on both languages.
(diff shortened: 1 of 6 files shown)


zechengz (approved):
Thanks! :)

PR #346, opened 2026-06-17
ci: add informational uv audit job to Python workflow

zechengz:
Adds a non-blocking `uv audit` job to the Python CI workflow that scans locked dependencies for known vulnerabilities (OSV) on PRs touching Python deps.

## Why
`uv audit` (Astral's new dependency vulnerability scanner) flagged a real path-traversal advisory (GHSA-gr75-jv2w-4656) in transitive `langchain` / `langchain-anthropic`. Those were already fixed on main by #342; this job adds standing coverage so the next such regression is surfaced in CI.

## Behavior
- **Informational only:** `continue-on-error: true` and the job is deliberately not part of `test-python-gate`, so it never blocks merges.
- Runs on the same `changes` paths-filter as the rest of the Python suite.
- `--all-extras --no-extra camel` matches the test job's sync scope (camel conflicts with the openai stack).
- `--preview-features audit` silences the experimental-preview warning.

## Notes
- The tool is in preview and currently exits 0 even on findings, so this stays informational for now. Promoting it to a hard gate later would need output parsing plus keeping the dep tree clean.

### DIFF
--- a/.github/workflows/test_python.yml
+++ b/.github/workflows/test_python.yml
@@ -181,7 +181,7 @@ jobs:
 
       - name: Audit dependencies for known vulnerabilities (informational)
         working-directory: python
-        run: uv audit --preview-features audit --all-extras --no-extra camel
+        run: uv audit --preview-features audit --no-extra camel
 
   gate:
     name: test-python-gate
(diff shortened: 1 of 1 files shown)


sonhmai:
@zechengz  NICE. THANKS!

PR #351, opened 2026-06-19
fix(fuse): surface startup worker failures

sonhmai:
## Summary

- propagate immediate Python FUSE worker startup failures back to the caller
- clean Mirage-owned generated mountpoints when startup fails
- add regression coverage for worker-thread startup errors and manager cleanup

## Tests

- `uv run pytest tests/fuse/test_mount.py tests/workspace/test_fuse.py`
- `./python/.venv/bin/pre-commit run --files python/mirage/fuse/mount.py python/mirage/workspace/fuse.py python/tests/fuse/test_mount.py python/tests/workspace/test_fuse.py`

### DIFF
--- a/python/mirage/fuse/mount.py
+++ b/python/mirage/fuse/mount.py
@@ -12,16 +12,20 @@
 # limitations under the License.
 # ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
 
+import logging
 import os
 import subprocess
 import sys
 import threading
-import time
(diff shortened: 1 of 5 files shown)


sonhmai:
Superseded by baf2b67e (per-mount Mount spec + FuseManager ownership). Startup failures now surface via _await_ready (ismount polling, worker-death detection), and add_fuse_mount drops failed manager state.

PR #352, opened 2026-06-19
build(deps): bump pypdf 6.13.2 -> 6.13.3 (fix uv audit)

sonhmai:
## What

Bump `pypdf` 6.13.2 → 6.13.3 in `python/uv.lock`.

## Why

The `uv audit` CI step (`test_python.yml`) fails on:

```
pypdf 6.13.2 has 1 known vulnerability:
- GHSA-jm82-fx9c-mx94: Missing stream length values ignore defined limits
  Fixed in: 6.13.3
```

`pypdf` is a transitive dependency of the optional `openhands` extra:

```
mirage-ai[openhands] -> openhands-tools -> browser-use -> pypdf
```

`browser-use` declares `pypdf` with no version specifier, so this is a lockfile-only bump — no `pyproject.toml` change. Core Mirage and clients not installing `[openhands]` are unaffected.

### DIFF
--- a/python/uv.lock
+++ b/python/uv.lock
@@ -8999,11 +8999,11 @@ wheels = [
 
 [[package]]
 name = "pypdf"
-version = "6.13.2"
+version = "6.13.3"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


zechengz (approved):
Thanks :)

PR #353, opened 2026-06-19
fix: remove unsupported OneDrive ls argument

binggao1230:
Summary:
- Remove the unsupported `trailing_newline` keyword from the OneDrive `ls` wrapper when delegating to the generic `ls` implementation.
- Add a cached OneDrive listing regression test that exercises the wrapper without requiring live OneDrive credentials.

Tests:
- RED before fix: `uv run pytest tests/commands/builtin/onedrive/test_ls.py -q` failed with `TypeError: ls() got an unexpected keyword argument 'trailing_newline'`.\n- `uv run pytest tests/commands/builtin/onedrive/test_ls.py tests/commands/builtin/generic/test_ls.py tests/commands/builtin/gdrive/test_provision.py --no-cov -q`\n- `uv run python -c "import mirage.commands.builtin.onedrive.ls"`\n- `./python/.venv/bin/pre-commit run --files python/mirage/commands/builtin/onedrive/ls.py python/tests/commands/builtin/onedrive/test_ls.py`\n- `uv run pytest --no-cov --ignore tests/fuse --ignore tests/workspace/test_fuse.py` -> 7709 passed, 316 skipped, 2 warnings.\n- Full `uv run pytest --no-cov` is blocked locally by missing `libfuse` while collecting FUSE tests.\n\nFixes #350\n\nAI assistance was used under my direction.

PR #354, opened 2026-06-19
fix(onedrive): repair ls, uniq, and recorded write-key bugs; add command test suite

### LINKED ISSUE #350: OneDrive: ls command fails with unexpected keywork argument trailing_newline

Bug

### Area

Python package

### Summary

The OneDrive `ls` command passes `trailing_newline=True` to the generic `ls()` function, which does not accept that parameter, causing a `TypeError` at runtime.

### Steps to reproduce

1. Mount a OneDrive resource in a workspace.
2. Run `ls` against any OneDrive path:
   ```python
   await ws.execute("ls /onedrive/")
   ```

### Expected behavior

The `ls` command lists files and directories from the OneDrive resource without errors.

### Actual behavior

```shell
The command raises:

TypeError: ls() got an unexpected keyword argument 'trailing_newline'


The OneDrive `ls` wrapper (`mirage/commands/builtin/onedrive/ls.py`, line 79) passes `trailing_newline=True` to `generic_ls`, but the generic `ls` signature (`mirage/commands/builtin/generic/ls.py`, line 189) does not define that parameter.
```

### Mirage version or commit

15ce838d

### Python version

_No response_

### Node.js version

_No response_

### Operating system

macOS

### Browser

_No response_

### Additional context

No other backend `ls` wrapper passes `trailing_newline`. The fix is either to remove `trailing_newline=True` from the OneDrive wrapper call, or to add a `trailing_newline` parameter to the generic `ls` function if trailing-newline behavior is desired.

sonhmai:
Fixes #350. Supersedes #353.

While fixing the `ls` crash from #350, an audit of the OneDrive command wrappers turned up three more correctness/parity bugs in the same area. This PR fixes all of them and adds the missing OneDrive command test suite that would have caught them.

#353 is a narrower fix for the same crash (removes the `trailing_newline` arg + one `ls` test); this PR includes that same fix plus the additional bugs below and a full command suite, so #353 can be closed in favor of this one.

### Fixes

- **`ls` crashes with `TypeError` (#350).** The wrapper passed `trailing_newline=True`, which `generic_ls` does not accept. Removed; no other backend passes it.
- **`uniq -w` default collapsed all output.** The wrapper passed `check_chars=int(w) if w else 0`. With `-w` absent it sent `0`, and the generic truncates each comparison key to `[:0]`, so every line compared equal and `uniq` emitted a single line. Now forwards raw `w`, so the generic's `_parse_count` yields `None` (no truncation).
- **Recorded write keys were double-prefixed.** `cp`/`mkdir`/`mv`/`touch` recorded `IOResult(writes=...)` under `path.original` (already mount-prefixed), producing doubled `/onedrive/onedrive/...` keys after the workspace re-prefix layer. Switched to mount-relative `path.strip_prefix`, matching `rm` and the s3/ram backends.

### Parity cleanup

- **`grep`/`rg`/`uniq` reduced to flag pass-through.** The wrappers pre-parsed and forwarded ~15 explicit kwargs; they now forward raw `texts` + the opaque `flags` bag and let the generics own parsing, per the "wrappers are wiring, generics own flags" convention and the disk/s3 wrappers. This also restores spec validation (a mistyped flag now raises instead of silently reading as `False`/`None`).

### Tests

New fake-store-backed OneDrive command suite (`conftest.py` plus read/write/text/metadata/ls/registration tests). Write-command tests assert mount-relative `io.writes` keys and the absence of the doubled prefix; size assertions are anchored to the target row/token. The conftest fails loudly if a command module's I/O fakes are not installed, guarding against a silent fall-through to the real (network) accessor on a future helper rename.

### Verification

All 32 OneDrive command tests pass.

### DIFF
--- a/python/mirage/commands/builtin/onedrive/cp.py
+++ b/python/mirage/commands/builtin/onedrive/cp.py
@@ -69,4 +69,4 @@ async def cp(
     output = None
     if v:
         output = f"{paths[0].original} -> {paths[1].original}\n".encode()
-    return output, IOResult(writes={paths[1].original: b""})
+    return output, IOResult(writes={paths[1].strip_prefix: b""})
--- a/python/mirage/commands/builtin/onedrive/grep.py
+++ b/python/mirage/commands/builtin/onedrive/grep.py
@@ -49,60 +49,20 @@ async def grep(
     paths: list[PathSpec],
     *texts: str,
     stdin: AsyncIterator[bytes] | bytes | None = None,
-    r: bool = False,
-    R: bool = False,
-    i: bool = False,
-    v: bool = False,
-    n: bool = False,
-    c: bool = False,
-    args_l: bool = False,
-    w: bool = False,
-    F: bool = False,
-    E: bool = False,
-    o: bool = False,
-    m: str | None = None,
-    q: bool = False,
-    H: bool = False,
-    args_h: bool = False,
-    A: str | None = None,
-    B: str | None = None,
-    C: str | None = None,
-    e: str | None = None,
     prefix: str = "",
     index: IndexCacheStore = None,
-    **_extra: object,
+    **flags: object,
 ) -> tuple[ByteSource | None, IOResult]:
-    if e is not None:
-        pattern = e
-    elif texts:
-        pattern = texts[0]
-    else:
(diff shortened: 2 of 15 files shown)


zechengz (approved):
Thanks!

PR #355, opened 2026-06-19
fix: databricks_volume stat reads content_length/last_modified

sonhmai:
## Problem

`databricks_volume` `stat` read `file_size` and `modification_time` off the Databricks SDK's `GetMetadataResponse`. That type has neither field — it exposes `content_length` (int) and `last_modified` (an RFC 7231 HTTP-date string). Both reads returned `None`, so `ls -l` (which calls `stat` per entry) rendered size `0` and an epoch date for every file.

Those names belong to two unrelated SDK types: DBFS `FileInfo` (`file_size`/`modification_time`) and the listing `DirectoryEntry` (`file_size`/`last_modified` as epoch-ms). The Files API `get_metadata` returns `GetMetadataResponse`.

## Fix

- `core/databricks_volume/stat.py`: read `content_length` and `last_modified`.
- `_modified` parses the RFC 7231 HTTP-date into ISO-8601 UTC (the `ls` long-format renderer needs `fromisoformat`, so an unparsed HTTP-date would still have shown the epoch). Naive datetimes and epoch-ms ints normalize to UTC too; unparseable values pass through verbatim.

This brings Python to parity with the TypeScript core, which already reads the `content-length`/`last-modified` headers and converts the date.

`readdir`'s use of `file_size`/`last_modified` is left unchanged — correct for the `DirectoryEntry` listing API.

## Tests

- Realigned both `databricks_volume` test fakes and seed helpers to the real `GetMetadataResponse` shape (`content_length`/`last_modified`). They had mirrored the buggy attribute names, which is why the bug went uncaught. `DirectoryEntry` listing fakes keep `file_size`.
- Added direct `_modified` branch coverage (HTTP-date, unparseable string, empty/None, naive/aware datetime, epoch-ms int) and a resource-level mtime assertion.

## Verification

- `databricks_volume` suites (core + commands + resource): pass.

### DIFF
--- a/python/mirage/core/databricks_volume/stat.py
+++ b/python/mirage/core/databricks_volume/stat.py
@@ -14,6 +14,7 @@
 
 import asyncio
 from datetime import datetime, timezone
+from email.utils import parsedate_to_datetime
 
 from mirage.accessor.databricks_volume import DatabricksVolumeAccessor
 from mirage.cache.index import IndexCacheStore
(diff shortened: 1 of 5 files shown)


zechengz (approved):
Thanks! :)

PR #357, opened 2026-06-19
fix(onedrive): pass shared command suite (s3 parity) + integ harness

zechengz:
OneDrive now passes the full shared command suite (`integ/truth.txt`, 501/501) — the same coverage every other backend has. The recent OneDrive PRs fixed command wrappers but the deeper core bugs survived because there was no integ test exercising the end-to-end read/write/index paths.

## Core fixes
- `core/onedrive/find.py`: don't apply `-size` to directories; set `is_empty` on entries (fixes `find -size`/`-empty`)
- `core/onedrive/stream.py`: wrap `capture_metadata` so a missing file raises enoent instead of a raw `GraphError` (fixes `head`/`tail` on missing files)
- `core/onedrive/mkdir.py`: support `parents=` (fixes `mktemp`, `tar --strip-components`)
- `commands/builtin/onedrive/tee.py`: write-cache key uses `strip_prefix`, not the mount-prefixed path (fixes the double-prefix that left the dir index stale, so a freshly written file read back as not-found)

## Wrapper migrations
- `cp`/`mv`/`tar`/`csplit`/`mktemp` now delegate to the generic commands (multi-source, coreutils guards). `cp` keeps OneDrive's server-side folder copy via a new optional `dir_copy` hook on the generic `cp` (additive; other backends unchanged).

## Tests
- `integ/onedrive_cases.py` runs the shared suite against `truth.txt` (mirrors `s3_cases.py`)
- `integ/onedrive.py` curated checks against `truth_onedrive.txt` (mirrors `s3.py`)
- both run against an in-process fake Microsoft Graph server (`integ/onedrive_server.py`); both wired into CI


zechengz:
merge at first as align onedrive with s3 (generic)

PR #367, opened 2026-06-20
feat: Add a Qdrant resource (py + ts)

anush008:
## Description

This PR adds a Qdrant resource.

[Qdrant](https://qdrant.tech/) is an open-source vector search engine for high-performance and massive scale.

With this resource, we can mount a Qdrant collection as a filesystem and use standard shell commands to browse it. The `search` command runs vector search and returns results as file paths, so you can pipe them straight into `cat` or `grep` without any extra glue.

For creating vector embeddings, Python supports both [Qdrant Cloud Inference](https://qdrant.tech/documentation/inference/) (`cloud_inference=True`) and local [FastEmbed](https://github.com/qdrant/fastembed) (`cloud_inference=False`, default). TypeScript has no local embedding option. it always uses Qdrant Cloud Inference.

You can get a free forever Qdrant instance with inference at https://cloud.qdrant.io/.

I've tested the implementation end-to-end with a Qdrant cloud instance of mine. Unit tests have been added to the CI.

### DIFF
--- a/.github/workflows/test_install.yml
+++ b/.github/workflows/test_install.yml
@@ -124,6 +124,7 @@ jobs:
           - { extra: langfuse, module: mirage.resource.langfuse }
           - { extra: chroma, module: mirage.resource.chroma }
           - { extra: lancedb, module: mirage.resource.lancedb }
+          - { extra: qdrant, module: mirage.resource.qdrant }
           - { extra: databricks, module: mirage.resource.databricks_volume }
           - { extra: pydantic-ai, module: mirage.agents.pydantic_ai }
           - { extra: openai, module: mirage.agents.openai_agents }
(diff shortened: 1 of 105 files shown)


zechengz:
Current VFS mapping for this resource:

- **Folders** = facets: each `group_by` payload field is a directory level; descending adds a `WHERE field=value` filter.
- **Files** = one point: `<id>.json` (full payload), `<id>.txt` (embedded source text, when `text_field` set), `<id>.<ext>` (raw blob, when `blob_field` set). The vector is hidden.
- **`search`** = vector kNN, returned as the content path + score.

How to map the underlying **source files / chunk lineage** (e.g. `document_id`/`source`/`page` → VFS paths, extracting the source filename) is tracked separately in #376.

zechengz (approved):
@Anush008 Thanks for the contribution! As described created an issue for how to represent qdrant in vfs.

Anush008:
Thanks 🙇

zechengz:
@Anush008 Actually let me know whether we can discuss more on that.

Anush008:
Yeah. I'm looking at it.

PR #368, opened 2026-06-21
perf(databricks_volume): serve ls/tree/find entry stats from the index

sonhmai:
## Problem

Listing a Databricks volume directory cost **1 list call + one `get_metadata` per entry** (plus a second `get_directory_metadata` per subdirectory), even though `list_directory_contents` already returns `size`, `last_modified`, and `is_directory` for every entry. `ls`, `ls -l`, `tree`, `find`, and glob all paid this N+1. A wide directory of 500 files issued ~501 Files API round-trips for one `ls -l`.

## Root cause

`readdir` already populated the per-resource index cache with each entry's `size` and type, but **dropped the modified time** and there was **no reader** — `stat` never consulted the index, always hitting the SDK. So the shared generic `ls`/`tree`/`find` walkers re-fetched metadata for every entry they had just listed.

## Fix

Adopt the existing s3 / nextcloud / onedrive idiom (index-backed `stat`), so the generic walkers stay backend-agnostic and untouched:

- **`readdir`** now also stores each entry's modified time (`remote_time`), converting the listing's epoch-ms `last_modified` through the same converter `stat` uses for the SDK's RFC-7231 string — so the ISO value is byte-identical on either path.
- **`stat`** consults `index.get()` first: a cached entry returns size/mtime/type with **no API call**; a path whose parent was listed but is absent resolves to `ENOENT` without a probe; SDK fallback on miss. `exists`/`cp`/`rm`/`rmdir` (which all `stat` first) benefit too.
- **TS mirrored.** The TS `find` wrapper additionally now threads `opts.index` into the walker (Python already did), so the cache is actually used there.

## Before → after (directory of N files, D subdirs)

| command | before | after |
|---|---|---|
| `ls`, `ls -l`, `ls -t`, `ls -S` | `1 list + N get_metadata` | `1 list + 0 get_metadata` |
| `tree` | `(D+1) list + (N+D) metadata` | `(D+1) list + 0 metadata` |
| `find -type/-size/-mtime` | per-entry classify + size/mtime stats | served from the index (only the search root is stat'd) |

Output of all these commands is unchanged.

## Tests

- **Python** — index-population assertions in `readdir`; `stat` fast-path / negative-cache / SDK-equivalence tests; command-level API-call-count tests (`ls`/`ls -l`/`-t`/`-S`, `tree`, `find -type`/`-size`/`-mtime`) asserting exactly 1 `list_directory_contents` and 0 per-entry `get_metadata`, via the `FakeFiles` recorders.
- **TypeScript** — mirrored vitest coverage including fetch-count assertions over the routed-fetch harness.


zechengz (approved):
Thanks :)

PR #391, opened 2026-06-25
Fix compact Python JSON rendering

C1-BA-B1-F3:
## Summary
- add compact JSON separators to Python single-line JSON renderers
- leave indented JSON output unchanged

## Test
- python3 -m compileall -q python/mirage/core python/mirage/commands/builtin
- git diff --check


zechengz (approved):
Thanks for the contribution :)

PR #392, opened 2026-06-25
Reject unsupported path-bound dispatch

### LINKED ISSUE #383: Dispatcher: path-bound command on a backend that lacks it falls back to /_default/ and returns a misleading "No such file"
## Problem

When a command has a path operand that resolves to a concrete mount X, but X does not register that command, `resolve_mount` does **not** reject it. It falls back to `mount_for_command(cmd_name)` → the hidden `/_default/` mount, which runs the command against its **own** (empty) namespace.

Result: the command appears to "run" but operates on the wrong backend, returning a misleading error even when the file exists on the real backend.

### Repro

A backend that doesn't register `sed` (e.g. gmail) with a path operand:

```
$ sed 's/x/y/' /gmail/<existing email>
sed: /gmail/<existing email>: No such file or directory   (exit 1)
```

The email exists, but `sed` resolved to `/_default/` (which can't see gmail's data), so it reports "No such file" instead of transforming. Same for any write command the backend lacks (`cp`, `tar`, `mkdir`, ... on a read-only backend).

### Root cause

`mirage/workspace/mount/registry.py` `resolve_mount()`:

```python
if mount is None or mount.resolve_command(cmd_name) is None:
    mount = self.mount_for_command(cmd_name)   # -> /_default/
```

The `/_default/` fallback is intended for **path-less** general commands (`echo`, `seq`, `bc`, `date`) that have no specific backend. Applying it to a command whose path clearly belongs to mount X is the bug.

## Proposed fix

Gate the fallback on whether the command is bound to a concrete mount by its path operands:

- If the command's path resolves to a real mount X and X doesn't register the command → **return a clear error** (e.g. `<cmd>: not supported on the <backend> backend`, distinct exit code) instead of routing to `/_default/`.
- Only fall back to `/_default/` when there is **no** path operand binding the command to a specific backend.

This turns today's misleading `No such file or directory` (for a file that exists) into an honest "this backend can't do that."

## Notes / sequencing

- Pre-existing dispatcher behavior, independent of the command-factory refactor (#379); the refactor reduces how often this triggers (more native commands per backend) but does not cause it.
- Most valuable **after** #382 (generic `sed`), so the only remaining fall-throughs are genuinely-unsupported write commands on read-only backends — exactly the cases that should error clearly.

## Acceptance

- A path-bound command unsupported on its target backend errors clearly (and does not read `/_default/`'s namespace).
- Path-less general commands (`echo`/`seq`/`bc`/...) still work via the `/_default/` fallback.
- Regression test covering both paths.

C1-BA-B1-F3:
## Summary

Closes #383. Scope: the mount-registry dispatch fix only. (The find changes that were previously bundled here have moved to #393.)

When a command's path operand resolves to a concrete mount X, but X does not register that command, `resolve_mount` no longer falls back to the hidden `/_default/` mount (which ran against its own empty namespace and reported a misleading `No such file` for a file that exists). Instead it raises `MountCommandUnsupported`, surfaced as exit 1 with `"<cmd>: not supported on the <backend> backend"`.

Path-less general commands (`echo`, `seq`, `bc`, ...) still fall back to `/_default/`.

## Changes
- `mount/registry.py`: add `MountCommandUnsupported`; gate the `/_default/` fallback on whether the command is path-bound.
- `executor/command.py`: map the exception to exit 1.
- `mount/__init__.py`: export the exception.
- `tests/workspace/mount/test_registry.py`: regression coverage for path-bound (rejected) vs path-less (fallback).

### DIFF
--- a/python/mirage/workspace/executor/command.py
+++ b/python/mirage/workspace/executor/command.py
@@ -37,7 +37,7 @@ from mirage.workspace.executor.find_action_dispatch import _apply_find_actions
 from mirage.workspace.executor.fs_error import format_fs_error
 from mirage.workspace.executor.jobs import (handle_jobs, handle_kill,
                                             handle_ps, handle_wait)
-from mirage.workspace.mount import MountRegistry
+from mirage.workspace.mount import MountCommandUnsupported, MountRegistry
 from mirage.workspace.session import Session, assert_mount_allowed
 from mirage.workspace.types import ExecutionNode
 
@@ -355,7 +355,15 @@ async def handle_command(
                 stderr=err.encode(),
             ), ExecutionNode(command=cmd_str, exit_code=1)
 
-    mount = await registry.resolve_mount(cmd_name, path_scopes, session.cwd)
+    try:
+        mount = await registry.resolve_mount(cmd_name, path_scopes,
+                                             session.cwd)
+    except MountCommandUnsupported as exc:
+        err = f"{exc}\n".encode()
+        return None, IOResult(exit_code=1,
+                              stderr=err), ExecutionNode(command=cmd_str,
+                                                         exit_code=1,
+                                                         stderr=err)
     if mount is None:
         return None, IOResult(
             exit_code=127,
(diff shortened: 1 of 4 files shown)


zechengz (approved):
Made some modifications. Thanks for the contribution :)

PR #393, opened 2026-06-25
fix: align find with GNU coreutils across all backends (#312)

### LINKED ISSUE #312: find: GNU coreutils divergences — start path omitted, -maxdepth 0 lists nothing, -empty/-not broken, unknown predicates silently match-all
## Summary

`find` diverges from GNU/POSIX coreutils in several ways. Some are cosmetic (start path), but two are dangerous: unsupported predicates (and typos) are **silently ignored and cause `find` to return *every* entry with exit 0**, and `-empty` / `-not` produce wrong result sets. All issues are **cross-backend** (ram/redis/disk/ssh/s3/api share the generic depth model) and **cross-language** (py + ts share `integ/truth.txt`).

## Reproduction

Tree: `data/{a.txt, sub/nested.txt, sub/deep/deeper.txt, emptydir/}` (a.txt non-empty, emptydir empty).

| command | GNU | mirage |
|---|---|---|
| `find data` | includes `data` | **omits the start dir** |
| `find data -type d` | includes `data` | **omits `data`** |
| `find data -maxdepth 0` | `data` | **(nothing)** |
| `find data -mindepth 0 -type d` | includes `data` | **omits `data`** |
| `find data/sub` | includes `data/sub` | **omits `data/sub`** |
| `find data/emptydir` | `data/emptydir` | **(nothing)** |
| `find data/a.txt` | `data/a.txt` | `data/a.txt` ✅ |
| `find data -name data` | `data` | **(nothing)** |
| `find data -empty` | only `data/emptydir` | **ALL entries** |
| `find data -not -name '*.txt'` | non-txt entries | **the `.txt` files (inverted)** |
| `find data -regex '.*deep.*'` | matches | **warns "unknown option ignored", returns ALL, exit 0** |
| `find data -newer data/a.txt` | newer entries | **warns, returns ALL, exit 0** |
| `find data -boguspredicate` | error, exit 1 | **warns, returns ALL, exit 0** |

## Bugs

**B1 — Start path is never emitted (depth-1 walk model).** The walk deliberately starts at depth 1, so the search root itself is never a result. This breaks bare `find <dir>`, `find <dir> -type d`, `-maxdepth 0` (GNU lists just the start), `-mindepth 0` (start excluded), matching the start by `-name`/`-path`, and `find <emptydir>` (returns nothing). File start paths work.
- Root: `python/mirage/commands/builtin/generic/find.py` `_walk_collect` starts at depth 1 (explicit comment: *"the walk starts at 1 and -maxdepth 0 lists nothing"*); `python/mirage/core/ram/find.py` & `core/redis/find.py` skip `if key == p and kind == "d": continue`; the per-backend walks (disk/ssh/s3/github/notion/onedrive/nextcloud/hf/chroma/dify) all omit the root.

**B2 — `-empty` is unimplemented and silently returns everything.** Not a field in `FindArgs`/`parse_find_args`, so it's dropped and no filtering happens. `integ` `find_empty` truth currently encodes the broken all-output.

**B3 — `-not` negation is broken.** `find -not -name '*.txt'` returns the `.txt` files (the un-negated set). `integ` `find_not_name` truth encodes the inverted result.

**B4 — Unsupported/unknown predicates silently match-all (exit 0).** `-regex`, `-newer`, `-perm`, `-prune`, `-delete`, `-exec`, `-depth`, and typos emit `find: warning: unknown option '...' ignored` (from `commands/spec/parser.py`) and then return **all** entries with exit 0. GNU errors `unknown primary or operator` and exits 1. This is the most dangerous case: a typo or unsupported flag turns `find` into "match everything" instead of failing.

**B5 — Cross-cutting.** Same behavior in every backend and both languages; `integ/truth.txt` (shared by py+ts ram/disk/redis/ssh/s3 + opfs) locks in B1/B2/B3, so fixing requires regenerating truth + adding cases (the deferred `root_create` block in `integ/cases.{py,ts}` is the placeholder).

## Supported predicates today
`-name -iname -path -type -size -mtime -maxdepth -mindepth -o`, and `-not -name` (via `_extract_not_name`, currently buggy). Everything else is dropped.

## Proposed fix
1. Emit the start path at depth 0 subject to predicates; fix `-maxdepth 0` (just the start) and `-mindepth 0` (include start).
2. Implement `-empty` and fix `-not`/`!` negation.
3. Make unknown/unsupported predicates a hard error (exit 1) instead of silently matching all — at minimum never return all on an unrecognized predicate.
4. Apply across all backends (generic walk + per-backend cores) in py + ts; regenerate `integ/truth.txt`/`truth_lancedb.txt` and re-enable the `root_create` integ block.

## Context
Discovered while writing root-create integ coverage during the path-primitive consolidation (#311), which deferred this. Not bundled there because it is a systemic, cross-backend behavior change.

C1-BA-B1-F3:
## Summary

Closes #312. Implements the GNU/POSIX `find` alignment across every backend and both languages (previously this PR only carried tests; the implementation now lives here).

### Bugs fixed
- **B1 — start path:** emit the search root at depth 0 (subject to predicates) in the generic walker and each per-backend find: ram/disk/redis/s3/onedrive/ssh/nextcloud (py) and ram/disk/opfs/redis/s3/ssh (ts). Fixes bare `find <dir>`, `-type d` on the root, `-maxdepth 0`, `-mindepth 0`, and matching the start by `-name`. Also fixes root-key rebasing in the generic find command.
- **B2 — `-empty`** returns only empty files/dirs.
- **B3 — `-not`** negation returns the complement set.
- **B4 — unknown/unsupported predicates** (`-regex`, `-boguspredicate`, typos) are a hard error (exit 1, `find: unknown predicate '...'`) instead of silently matching all entries.
- **Fan-out:** a descendant mount's own root line is dropped; the mount point is synthesized centrally with its display name and the full predicate tree, so `find /data -not -name ram` correctly excludes the `ram` child mount.

### Tests
- Regenerated `integ/truth.txt` (start path now appears in the 5 cases where the root matches the predicate + two unknown-predicate error cases).
- `integ/cases.{py,ts}`: unknown-predicate error cases.
- Unit coverage for `-empty`, `-not`, unknown predicates, start-path emission, and per-backend root inclusion.

### Verification
- `integ` RAM/disk/redis and TS RAM produce byte-identical output matching the regenerated truth.
- Full PY find/fanout unit tests and TS core/node/browser suites pass.

### DIFF
--- a/integ/cases.py
+++ b/integ/cases.py
@@ -465,6 +465,9 @@ CASES: list[tuple[str, str]] = [
     # ----- find more -----
     ("find_empty", "find /data -empty"),
     ("find_not_name", 'find /data -not -name "*.txt"'),
+    ("find_name_start", "find /data -name data"),
+    ("find_maxdepth_zero", "find /data -maxdepth 0"),
+    ("find_mindepth_zero", "find /data -mindepth 0 -type d"),
     ("find_size_lt", "find /data -size -5c"),
     ("find_depth", "find /data -depth -type f"),
     ("find_mtime", "find /data -mtime +0 -o -mtime -1"),
--- a/integ/cases.py
+++ b/integ/cases.py
@@ -825,6 +825,8 @@ FIND_ARG_ERROR_CASES: list[tuple[str, str]] = [
     ("find_bad_size", "find /data -size abc"),
     ("find_empty_size", "find /data -size ''"),
     ("find_bad_mtime", "find /data -mtime abc"),
+    ("find_unknown_predicate", "find /data -regex '.*deep.*'"),
+    ("find_bogus_predicate", "find /data -boguspredicate"),
 ]
 
 SLEEP_CASES: list[tuple[str, str, float]] = [
--- a/integ/cases.ts
+++ b/integ/cases.ts
@@ -703,6 +703,8 @@ export const FIND_ARG_ERROR_CASES: ReadonlyArray<readonly [string, string]> = [
   ['find_bad_size', 'find /data -size abc'],
   ['find_empty_size', "find /data -size ''"],
   ['find_bad_mtime', 'find /data -mtime abc'],
+  ['find_unknown_predicate', "find /data -regex '.*deep.*'"],
+  ['find_bogus_predicate', 'find /data -boguspredicate'],
 ];
 
 export const SLEEP_CASES: ReadonlyArray<readonly [string, string, number]> = [
--- a/integ/cases.ts
+++ b/integ/cases.ts
@@ -387,6 +387,9 @@ export const CASES: ReadonlyArray<readonly [string, string]> = [
 
   ["find_empty", "find /data -empty"],
   ["find_not_name", 'find /data -not -name "*.txt"'],
+  ["find_name_start", "find /data -name data"],
+  ["find_maxdepth_zero", "find /data -maxdepth 0"],
+  ["find_mindepth_zero", "find /data -mindepth 0 -type d"],
   ["find_size_lt", "find /data -size -5c"],
   ["find_depth", "find /data -depth -type f"],
   ["find_mtime", "find /data -mtime +0 -o -mtime -1"],
--- a/integ/truth.txt
+++ b/integ/truth.txt
@@ -1273,6 +1273,18 @@ baz,
 /data/sub/deep
 /data/user.json
 /data/users.json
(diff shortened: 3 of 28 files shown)


zechengz (approved):
Please let me know if you want to contribute more (you can join the https://discord.com/invite/u8BPQ65KsS)

Issue #415, opened 2026-06-29
Symlinks: explore registering ln/readlink as general commands with scoped symlink-table access


zechengz:
Currently `ln -s` and plain `readlink` are router intercepts (`_dispatch_command_body`), because the symlink table lives on `MountRegistry` and the command-calling convention never passes the registry to a command (a command gets a backend accessor + resolved paths and returns output only).

This issue tracks the alternative: register `ln`/`readlink` as general commands (like `date`/`bc`) that are granted **scoped** access to the symlink table, instead of intercepting in the router.

What it would take:
- A way for a command to read and mutate the symlink table. Either inject a narrow `symlinks` handle into the general-command call convention, or add a side-effect return channel so a command can request a table mutation without holding the registry.
- Keep it scoped: do NOT hand commands the full `MountRegistry` / topology mutation power. Inject only the symlink-table read/write capability.

Tradeoffs to weigh:
- Pro: `ln`/`readlink` become normal commands (spec-based flag parsing, consistent with the rest), router shrinks.
- Con: widens the shared command contract for ~2 of ~60 commands; needs a clean, narrow capability so it does not become a general topology-mutation hole.

Relates-to: #413 (storage shape), #414 (follow-on-read).

zechengz:
Resolved, by a different mechanism than the one proposed here.

`ln` and `readlink` are no longer router intercepts. They are a formal **NAMESPACE** tier in the precedence table (`workspace/names.py`: `NAMESPACE_COMMANDS = frozenset({"ln", "readlink"})`), sitting between shell builtins and functions in `_layers`/`layers`, and they parse through the ordinary spec machinery like every other command (`ln`: `-s -f -n -v`; `readlink`: `-f -e -m -n`). `type` reports them, and `lookup` classifies them as their own consumer.

That gets what the issue wanted — spec-based flag parsing, a smaller router — without widening the shared command contract for 2 of ~60 commands or handing a command the `MountRegistry`. The symlink table stays namespace state that no backend or command can reach directly (see the Symlinks section in CLAUDE.md).

Closing as resolved-differently; reopen if the injected-handle route is still wanted for its own sake.

Issue #414, opened 2026-06-29
Symlinks: follow-on-read so all commands resolve links (VFS-style)


zechengz:
Today only `cd` and `readlink` consult the symlink table. `cat`, `ls`, `grep`, `rm` on a link path hit ENOENT because nothing resolves the link for them.

A real OS resolves symlinks in the kernel VFS on every path lookup, so all programs follow links transparently. The mirage-equivalent is to resolve symlinks during central path classification (the layer that already holds the registry and resolves cross-mount paths), the same way `$VAR` env expansion is a cross-cutting transform applied before dispatch.

The catch (why this is more than env expansion): resolution needs a per-command follow / no-follow policy.
- Follow: `cat`, `grep`, `cd` (default), most readers.
- Do NOT follow (operate on the link itself / lstat-style): `ls -l`, `readlink`, `ln`, `rm` of the link.

Scope:
- Add a follow/no-follow flag to the classification step and resolve via the existing registry table.
- Decide the default and the per-command exemptions (mirror GNU find -P/-L/-H and ls/rm semantics where practical).
- Mirror in py and ts, add integ coverage.

Blocked-by / relates-to: #413 (storage shape).

zechengz:
## Design decision: registry-as-path-authority, not scattered injection

Building follow-on-read by (a) resolving symlink path args in the dispatcher and (b) injecting link entries per-consumer (`ls`, then glob, `find`, `tree`) was tried and rejected as ugly: symlink logic ends up scattered across two unrelated layers, and every new directory-enumerating command needs its own patch.

The smell is that there is no single path authority. The registry already owns path routing (`mount_for`), so both symlink behaviors belong there as one chokepoint each:

1. **Resolution.** Canonicalize a virtual path through links inside the registry's path resolution, so every read/stat/lookup follows links with zero per-command code. No `_follow_symlinks` in the dispatcher, no hardcoded no-follow set sprinkled around: the follow/no-follow policy lives in one resolver.
2. **Enumeration.** One `registry.list_dir(path)` that merges registry symlink entries whose parent is `path`, which `ls` / glob-expansion / `find` / `tree` all call, instead of N separate injections.

Two single chokepoints in the layer that already owns paths. This is consolidation, not a VFS rewrite, and is a focused PR of its own.

### Scope for the follow-up PR
- Resolution chokepoint (covers `cat link`, `head a b`, cross-mount, and globbed symlinks once they enumerate).
[...]

Issue #413, opened 2026-06-29
Symlinks: revisit table storage location and data structure


zechengz:
Symlinks are currently a flat `dict[str, str]` (link abs path -> target) on `MountRegistry`, mirrored in py and ts.

Rationale for the registry as the home: it is the one workspace object that is both shared across subshells and aware of all mounts. A symlink must survive a subshell (`( ln -s a b )` then use `b`) and can point across mounts (`/s3/x` -> `/ram/link`). Session forks per subshell; a single resource only knows itself. The registry is our filesystem layer (analogous to the kernel VFS), so it is the OS-faithful place for symlinks to live.

Open questions to evaluate:
- Should the bare dict become a small `SymlinkTable` type (longest-prefix match, cycle cap) instead of free functions over a dict?
- Should the target be stored verbatim (as typed, possibly relative) like a real symlink, instead of resolved-to-absolute? This affects `readlink` fidelity and relative-link semantics.
- Does anything symlink-related belong on `Session` rather than the registry? (current answer: no, because of subshell persistence.)
- Snapshot shape: today it is a plain str->str map under `StateKey.SYMLINKS`. Confirm this stays the right serialized form if the in-memory type changes.

PR #418, opened 2026-06-29
Add symlink support (ln -s, readlink, cd)

zechengz:
Adds symbolic links as a workspace-level overlay on the mount registry, mirrored in Python and TypeScript.

## What works
- `ln -s target link` creates a link in the registry. The target is stored verbatim (relative or absolute), so `readlink` echoes exactly what was typed, matching GNU. `-f` overwrites, `-v` is verbose, mount roots and existing links are protected.
- `readlink link` returns the verbatim target; a non-link exits 1 with no output. `readlink -f/-e/-m` (canonicalize) still falls through to the existing command.
- `cd` follows links, with `-L` (default) vs `-P` divergence on `..` after a link, and a "Too many levels of symbolic links" (ELOOP) error on a cycle.
- Links are workspace-level: they survive subshells and can point across mounts (`/ram/link` -> `/s3/x`).
- Snapshots capture and restore the link table (`StateKey.SYMLINKS`).

## Design
Links live on `MountRegistry` (the one workspace-shared, all-mounts object), not on a session or a backend. `ln`/`readlink` are router intercepts because they mutate registry state, which mount commands are not handed; `cd` reads the same table. See #413 for the storage rationale and #415 for the general-command alternative.

## Deferred (#414)
Follow-on-read, where `cat`/`ls`/glob/`find`/`tree` follow or list links, is intentionally not in this PR. The clean fix is a single resolution chokepoint plus a `registry.list_dir` enumeration seam (not scattered per-command injection); the design and caching notes are in #414.

## Tests
- Unit: `resolveSymlinks` and the `ln`/`readlink`/`cd`/snapshot behavior in both languages.
- Integ: `ln -s` / `readlink` / `cd`-through / ELOOP cases in the shared `truth.txt`, cross-mount checks in `cross_commands`, and symlink probes in `parity.sh`. The previous `ln`-then-`cat` read-back case became a `readlink` case, since `ln -s` is now a real link rather than a byte copy.


zechengz:
Bad design, close for now 😒

PR #443, opened 2026-07-07
fix(cache): clamp the background drain budget to the cache limit

sonhmai:
## Problem

image

```python
ws = Workspace({"/s3": (S3Resource(cfg), MountMode.READ)})   # default RAM cache, 512MB
await ws.execute("head -c 100 /s3/data/events.parquet")      # events.parquet is 20GB
```

What happens before this PR:

1. User runs `head -c 100` on a 20GB S3 file. Head reads one 8KB chunk,
   prints 100 bytes, done. 8KB was all we need.
2. Mirage sees the read stopped mid-file and starts a background download
   to finish it, hoping to cache the file for next time.
3. That download has no size limit (`max_drain_bytes=None` was the default).
   It quietly pulls the remaining 20GB and keeps all of it in memory. The
   user's command already returned, the process just grows by 20GB in the
   background until the OS kills it.
4. If it survives, it tries to cache the 20GB. The cache holds 512MB: it
   throws out every existing entry to make room, then throws out the 20GB
   entry too. Final score: 20GB downloaded, 20GB of RAM, cache emptied,
   nothing cached.

`max_drain_bytes` defaults to `None`, which the background drain treated as unbounded. A partially consumed streamed read (`head -c 100` of a multi-GB S3 object) spawned `_background_drain`, which kept downloading and buffering the entire remainder in process memory. The oversized `add` then evicted every other entry and finally the new entry itself, so the full download bought nothing and wiped the cache.

Draining more than `cache_limit` bytes can never produce a cache hit: eviction runs right after the add and an entry larger than the limit is always dropped. Unbounded draining is therefore pure waste and there is no reason to keep it as an option.

## Change

- `FileCacheMixin.drain_budget` (TS `drainBudget(cache)`) resolves the effective budget: `min(max_drain_bytes or cache_limit, cache_limit)`.
- `_background_drain` / `backgroundDrain` always use `drain_bounded`; the unbounded `drain()` branch is gone. Exceeding the budget logs and skips the fill, and the next read fetches fresh from the resource.
- Config surface is unchanged. `None` now means "default to the cache limit" instead of "unbounded"; values above the limit are clamped.
- Snapshot state still round-trips the raw configured `max_drain_bytes`, so old snapshots with `None` load with the new bounded semantics.

## Tests

- Python: `drain_budget` default/keep/clamp cases; default budget stops a stream larger than `cache_limit` and still caches smaller ones; clamp applies when `max_drain_bytes` exceeds the limit.
- TypeScript: same three `drainBudget` cases plus the mirrored drain behavior tests.


zechengz:
## Follow-up changes after review

We extended this PR beyond the original drain-budget clamp to make bounded draining safe end to end.

### Concrete flow

Given a 24 KiB streamed object, an 8 KiB backend chunk size, and a 4 KiB `max_drain_bytes`:

```text
cat object | head -c 100
  -> head returns after 100 bytes
  -> the cache drain sees the already-buffered 8 KiB chunk
  -> 8 KiB exceeds the 4 KiB drain budget
  -> no additional backend chunk is read
  -> the partial buffer is discarded
  -> the backend source is closed
  -> no incomplete cache entry is written
```

### What changed

- `None` now derives the drain budget from the cache limit; explicit drain budgets must be non-negative and no larger than the cache limit.
- Over-budget drains return no materialized value, clear their buffered chunks, and close the underlying source. This avoids constructing a second concatenated copy of data that cannot be cached.
- Python mount, revision, and read-through wrappers now propagate early close correctly while preserving the original `CachableAsyncIterator`.
- S3 response bodies are closed after early termination, while lightweight test bodies that only implement `iter_chunks()` remain supported.
- Databricks Volume response readers are cancelled and their locks released when consumers stop early.
- S3 and Databricks Volume bounded-drain integration coverage was added for both Python and TypeScript, and the Databricks checks now run in the integration workflow.

### CI follow-up

The dependency audit was also repaired by updating `httplib2` to 0.32.0 and `setuptools` to 83.0.0.

All PR checks are green, including Python tests/audit, TypeScript tests, S3/Databricks integrations, pre-commit, and the required gates.

PR #447, opened 2026-07-08
fix(crossmount): append GNU strerror suffix on cat's not-found branch (#296)

### LINKED ISSUE #296: Cross-mount cat: missing GNU strerror suffix on the not-found branch
## Problem

When catting files across multiple mounts and one is missing, e.g.:

\`\`\`
cat /disk/a /ram/missing
\`\`\`

the not-found branch emits the virtual path correctly but omits the GNU strerror suffix:

\`\`\`
cat: /ram/missing
\`\`\`

Expected (GNU parity, matching the single-mount case which is correct):

\`\`\`
cat: /ram/missing: No such file or directory
\`\`\`

## Scope

The virtual-path invariant itself holds (the path shown is correct). What is missing is appending the strerror via the shared \`fs_strerror\` / \`gnuStrerror\` formatter on the multi-path fan-out branch, so cross-mount errors match the single-path chokepoint exactly. Fix in both Python and TypeScript; add an integ case with two mounts where one path is missing.

rikaqu0223-arch:
## Problem

Cross-mount commands route filesystem errors through a different chokepoint than single-mount ones, and that chokepoint dropped the GNU strerror suffix. A missing operand printed:

```
cat: /ram/missing
```

instead of the single-mount (and GNU) format:

```
cat: /ram/missing: No such file or directory
```

## Fix

One shared formatter, used by both chokepoints, in both languages:

- `format_fs_error` moves from `workspace/executor/fs_error.py` into the leaf `mirage/utils/errors.py` (the generics layer cannot import workspace, so the formatter lives below both). TypeScript mirrors it with `formatFsError` in `utils/errors.ts`.
- The single-mount executor catch and the cross-mount `handle_cross_mount` / `handleCrossMount` catch both call it, so the two paths can never diverge again.

After the STREAM/FANOUT/RELAY rework on main, STREAM and FANOUT commands (cat/head/tail/grep/...) format errors inside their per-mount native runs; the cross-mount catch is the chokepoint for the RELAY commands (diff/cmp/cp/mv) and a backstop for the rest. The conflict resolution also renames the removed `PathSpec.display` to `raw_path` / `rawPath`; without that, cross-mount `diff`/`cmp` with a missing operand crashed with an AttributeError.

## Tests

- Unit: cross-mount missing-operand for `cat`, `diff`, `cmp` in `tests/workspace/test_cross_mount_errors.py`; `tests/utils/test_errors.py` mirrors the moved formatter; TS `cmp` missing-operand case in `cross_mount.test.ts`.
- Integ: two-mount missing-path cases in `check_read_family` and `check_compare` (Python + TS).

Fixes #296

Contributor GitHub ID: @rikaqu0223-arch


zechengz (approved):
In general LGTM. Made some modifications.

Issue #448, opened 2026-07-08
index cache: readdir keys are not canonical; mount-root cache hits return doubled-slash paths


zechengz:
## Symptom

On a disk or redis mount with a warm index, commands that print full paths from a mount-root listing return doubled slashes:

```
ls /data > /dev/null          # warms the index for the mount root
grep -r foo /data             # -> /data//a.txt:foo     (GNU: /data/a.txt:foo)
wc -l /data/*.txt             # -> 1 /data//a.txt
```

The paths still read fine (normalization strips `//` on access), so no wrong bytes and no stale listings, but agents echo the malformed paths into subsequent commands and records.

## Root cause

Two layers combine:

1. Backend readdir builds the index cache key by concatenation (`prefix + path`). At a mount root the stripped remainder is `/`, so the key comes out as `/data/` while other code paths look up `/data`. One directory, two keys.
2. The index store builds a cached child's full path as `key + "/" + name`. Under the slashed key that yields `/data//a.txt`, returned verbatim on every cache hit.

The dual key also means cache hits/misses depend on which code path asks, and the cache manager already works around it by invalidating both shapes on every write (`invalidate_dir(virtual)` and `invalidate_dir(virtual + "/")`).

Why CI never saw it: the RAM backend has `index_ttl=0` (always cold, always clean) and RAM generates `truth.txt`; the existing recursive-grep case targets a subdirectory, whose key was never malformed.

## Audit (all backends, both languages)

**Confirmed buggy, fixed in the cross-mount glob PR** (canonical key: trailing slash stripped, root excepted):
- Python: `core/disk`, `core/ram`, `core/redis`, `core/ssh` (ssh defensively; doubling not reproduced in the shared runner)
- TypeScript: `core/ram` (core pkg), `core/disk`, `core/redis` (node pkg)

**Suspected, same construction, not yet reproduced** (not shared-truth runners):
- Python `core/postgres`: `virtual_key = (prefix or \"\") + raw` with `raw = \"/\"` at mount root -> slashed key
- TypeScript `core/postgres`: same expression
- Python `core/github`: `list_dir(path)` with the raw path, shape depends on caller

**Canonical by construction (verified):**
[...]

PR #449, opened 2026-07-08
Cross-mount commands expand glob operands; canonical readdir index keys

zechengz:
## What

Cross-mount execution bypasses the resource command handlers, so pattern operands never resolved: `cat /ram/*.txt /disk/*.txt` failed with `cat: /ram/*.txt`. Patterns now expand at the cross-mount entry in `executor/command`, one backend listing per pattern, in operand order, with a zero-match pattern keeping the literal word so the error reads like GNU. All four families (read, cp/mv, diff/cmp, du/file/md5) inherit the expansion.

Cross-mount `cat -n` also numbers continuously across mounts now (Python ignored `-n` on this path; TS already honored it).

## Index key fix

Wiring this up exposed a latent index-cache bug, user-visible on main: with a warm index, `grep -r foo /data` on a disk or redis mount printed `/data//a.txt:foo`. The readdir cache key was built by concatenation, so a mount root indexed under both `/data` and `/data/`, and the store returned children of the slashed key as `/data//a.txt` on every hit. Keys are now canonical (trailing slash stripped, root excepted) for disk/ram/redis in both languages plus Python ssh; the cache manager already invalidates both shapes, so eviction still pairs. CI never saw it because the RAM backend's `index_ttl=0` generates the shared truth. Full backend audit and remaining suspects (postgres, github) tracked in #448.

## Testing

- Unit/e2e both languages: cross-mount glob expansion in operand order, `cat -n` continuity, `wc` totals, `grep -c` labels, `cp` glob sources, zero-match literal; readdir cache-hit cleanliness and key-shape insensitivity
- 5 new shared-truth integ cases (`xm_glob_*` cross-mount globs, `grep_r_warm_root` warm-mount-root sentinel); truth regen is purely additive
- 10 of 13 shared-truth runners verified byte-identical locally (py ram/disk/redis/ssh/onedrive, ts ram/disk/opfs/redis/ssh); the remaining three (py nextcloud/s3, ts s3) use canonical key construction
- Full Python suite and TS core+node suites green; tsc baseline unchanged

Refs #448


zechengz:
Bad approach ;)

Issue #336, opened 2026-06-16
GNU-align relative-glob display: cd /data && wc *.txt should print a.txt, not /data/a.txt


zechengz:
## Summary

A relative glob argument is rendered as its **resolved absolute path** instead of the as-typed form. GNU shells expand a relative glob to relative paths, so the command sees (and prints) `a.txt`, not `/data/a.txt`.

```
cd /data
wc -l *.txt          # mirage: "1 /data/a.txt ..."   GNU: "1 a.txt ..."
wc -l disptree/*.txt # mirage: "1 /data/disptree/x.txt"  GNU: "1 disptree/x.txt"
grep -l p *.txt      # mirage: "/data/a.txt"         GNU: "a.txt"
```

Absolute-glob args (`wc /data/*.txt`) must stay absolute.

## Context

This is the one remaining piece of the relative-path display work (the rest shipped in #331: direct args, errors, head/tail, grep/grep -r/-l, rg, ls -R, find all render the as-typed path via the `PathSpec.as_typed`/`display` field + `rebase_one`/`rebase_display`).

## Why it's deferred

Path globs expand in the per-backend `resolve_glob` (pushdown path), which builds matches via `PathSpec.from_str_path(...)` and drops the as-typed form. A working approach was prototyped:

1. `classify` already stamps `as_typed` on relative glob specs.
2. In the central `resolve_globs` (`workspace/node/resolve_globs.py` / `.ts`), stamp each match's `as_typed = <typed dir-prefix> + basename(match)` (e.g. `*.txt` -> `a.txt`, `sub/*.txt` -> `sub/x.txt`).
3. Hand the glob-resolved specs to the mount command instead of raw `classified` (pushdown already happened in `resolve_globs` via `resource.resolve_glob`).

This worked in **Python across all backends** and in **TS in isolation**, but is **blocked by a TS glob index-cache bug** (linked below) where a later glob's matched `.original` comes back in a form that breaks `handleCommand`'s `scopeMap` recovery, so `as_typed` is lost. Shipping py-only would violate py/ts parity, so the whole feature was reverted from #331.

## Acceptance

- `cd /data && wc -l *.txt` / `disptree/*.txt` / `grep -l p *.txt` render relative, consistent across ram/disk/redis/s3/ssh in **both** languages.
- Absolute globs unchanged.
- Shared integ cases added (`disp_glob_*`) + truth regenerated.

## Blocked by

[...]

PR #456, opened 2026-07-10
fix(shell): relative words follow bash semantics

zechengz:
Fixes four bash-parity bugs around relative words, trailing redirects, and error spelling, and consolidates the word-spelling model, in both langs.

## test resolves relative path operands

`cd /data && test -f plain.txt` returned 1 even when the file exists: `-f` and `-d` scoped string operands without the session cwd. Relative strings now resolve through the shared `resolve_path` before the stat or readdir probe; empty operands stay false without a backend call.

## Words render as typed

The Argv text view, for/select loop variables, and function positional args used `PathSpec.virtual`, so `echo sub/file.txt` printed the resolved absolute path where bash hands programs their words unchanged. All three now render the typed spelling, and `resolve_globs` stamps glob matches with the spelling the typed word implies: `sub/*.txt` expands to `sub/a.txt`, absolute patterns keep absolute matches, and a zero-match relative glob keeps the typed literal (the TS glob clone also dropped `rawPath`, absolutizing kept literals). Relative values re-resolve on use, so `for f in sub/*.txt; do cat $f; done` still reads the right files.

## Trailing redirect binds to the last command of a list

tree-sitter-bash hoists a trailing redirect over the whole `&&`/`||` list, so `cd /data && echo hi > BARE` expanded the target before cd ran (writing /BARE) and `a && b > f` captured both commands' output. The executor now re-associates at the redirected-statement branch: `redirected(list(L, op, R), r)` runs as `list(L, op, redirected(R, r))`, expanding the target only when R runs. Multi-redirect chains compound correctly, compound and subshell bodies keep bash's whole-body group redirect, and the provision planner mirrors the re-association. Replaces the partial list handling inside handle_redirect, which ran after eager expansion, dropped the left side's stdout, and never ran the right side of a semicolon list.

## raw_path owns word spelling; drain errors respell as typed

Follow-on from reviewing the spelling semantics (commit 5f79bdb3):

- `raw_path`/`rawPath` is now total (defaults to `virtual`) and the `display` property is deleted: one name for the word's spelling. A word's typed form and the cwd it was typed under are two halves of one fact; the new `relative_spec`/`relativeSpec` constructor owns that construction (four hand-rolled copies collapsed), `word_text`/`wordText` owns the str-or-spelling coercion, and `rebase_display` is renamed `rebase_raw` with its dead nullable parameter dropped.
- Fix: lazy-stream drain errors now respell operands as typed. `cd /data && cat sub` said `cat: /data/sub: ...` in Python, and the TS drain printed the bare message with no command prefix at all. `ExecutionNode` carries the leaf command's operands (transient, not serialized) so the drain formats like the eager chokepoint, byte-identical across languages.

Related pre-existing bugs found during review, filed separately: #457 (cat on a directory reports ENOENT on keyed backends), #458 (ls with a glob operand prints bare names). #336 (mount-command glob labels, `wc *.txt`) is not fixed here but the new `_match_raw` stamping is the groundwork its prototype described.

## Tests

- Unit: new test_condition.py, test_redirect.py, test_relative.py, test_program.py and their TS mirrors; word_text/wordText and glob spelling cases; test_parse updates.
- Integ: 22 new shared-truth cases (relative test/echo/for/function words, redirect after cd, last-command capture, multi-redirect chain, group redirect, walker label rebasing, error spelling, ./ glob, du trailing slash); truth diff is purely additive, no existing output line changed.

### DIFF
--- a/integ/cases.py
+++ b/integ/cases.py
@@ -635,6 +635,14 @@ CASES: list[tuple[str, str]] = [
     ("glob_relative_for", "for f in sub/*.txt; do echo $f; done"),
     ("glob_relative_func",
      "f() { echo $1 $#; }; f sub/*.txt && cd / && rm -r /data/g12"),
+    ("redirect_after_cd", "mkdir -p /data/g13 && cd /data/g13"
+     " && echo hi > CDOUT && cat /data/g13/CDOUT"),
+    ("redirect_list_last_only",
+     "echo one && echo two > captured && cat captured"),
+    ("redirect_chain",
+     "echo a > chain && echo b >> chain && cat chain && wc -l < chain"),
+    ("redirect_group",
+     "{ echo g1; echo g2; } > gout && cat gout && cd / && rm -r /data/g13"),
 
     # ----- cp / mv multi-source into a directory (last; these mutate) -----
     ("cp_multi_into_dir", "cp /data/a.txt /data/b.txt /data/sub"),
--- a/integ/cases.py
+++ b/integ/cases.py
@@ -624,6 +624,17 @@ CASES: list[tuple[str, str]] = [
     ("redirect_relative_write", "echo one > sub/LOG"),
     ("redirect_relative_append", "echo two >> sub/LOG && cat sub/LOG"),
     ("redirect_stdin_relative", "wc -l < sub/LOG && cd / && rm -r /data/g11"),
+    ("relword_prep", "mkdir -p /data/g12/sub && cd /data/g12"
+     " && printf 'y\n' | tee plain.txt > /dev/null"
+     " && printf 'x\n' | tee sub/a.txt > /dev/null"
+     " && printf 'x\n' | tee sub/b.txt > /dev/null"),
+    ("test_relative_paths",
+     "test -f plain.txt && echo yes-f && test -d sub && echo yes-d"
+     " && test -f missing.txt || echo no-f"),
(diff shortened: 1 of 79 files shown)


zechengz:
Added 5f79bdb3, a follow-on from reviewing the spelling semantics:

- `raw_path`/`rawPath` is now total (defaults to `virtual`) and `PathSpec.display` is deleted: one name for the word's spelling. New `relative_spec`/`relativeSpec` constructor owns typed word + cwd -> PathSpec; `word_text`/`wordText` owns the str-or-spelling coercion; `rebase_display` renamed `rebase_raw`.
- Fix: lazy-stream drain errors now respell operands as typed (`cat sub` after `cd /data` said `cat: /data/sub: ...`, and the TS drain printed the bare message with no command prefix at all). ExecutionNode carries the leaf operands so the drain formats like the eager chokepoint, byte-identical across languages.
- Integ: relspell battery (grep -r / find label rebasing, error spelling, ./ glob display, du trailing slash, wc/head labels).

Related pre-existing bugs found during review, filed separately: #457 (cat on a directory reports ENOENT on keyed backends), #458 (ls with a glob operand prints bare names). Mid-path glob support (`s*/x.txt`) errors out today; the fix for that already exists on the cross-fuse worktree branch and should land as its own PR.

Local verification: full py suite, core 3630 + node 1564 vitest, tsc at the 24-error baseline, pre-commit clean, ram/disk/redis truth runners byte-identical in both languages.

PR #466, opened 2026-07-10
fix(daemon): make pid file path configurable via MIRAGE_HOME / MIRAGE_PID_FILE

### LINKED ISSUE #460: [Issue]: pid file path is hardcoded to `Path.home()` and can fail on read-only / restricted `$HOME`

Bug

### Area

CLI

## Summary

When the Mirage daemon starts, it writes a pid file to a path hardcoded under the
user's home directory. There is no way to configure this location. On any
environment where `Path.home()` is not writable (containers, sandboxed users,
`nobody`, unset/read-only `$HOME`, immutable rootfs), daemon startup fails.

## Where

- `python/mirage/server/app.py:38-45`

  ```python
  def _pid_file_path() -> Path:
      return Path.home() / ".mirage" / "daemon.pid"

  def _write_pid_file() -> None:
      p = _pid_file_path()
      p.parent.mkdir(parents=True, exist_ok=True)
      p.write_text(str(os.getpid()))
  ```

  `_write_pid_file()` is called unconditionally from the ASGI lifespan
  (`app.py:70`), so both the `mkdir()` and the `write_text()` run at startup.

- A **second hardcoded copy** exists in `python/mirage/cli/daemon.py:29`:

  ```python
  _PID_FILE = Path.home() / ".mirage" / "daemon.pid"
  ```

  Any fix must keep these two in sync (ideally by sharing one resolver).

## Why it's a problem

1. **No override.** The path cannot be changed via argument or env var, unlike
   the sibling roots which already are configurable:
   - `build_app(version_root=..., snapshot_root=...)` (`app.py:84-85`, `131-135`)
   - `MIRAGE_VERSION_ROOT`, `MIRAGE_SNAPSHOT_ROOT` (`python/mirage/server/env.py:17-18`)

   So the pid file is inconsistent with the rest of the daemon's data-path
   handling.

2. **Hard failure, not swallowed.** `_write_pid_file()` does not catch
   `PermissionError` / `OSError`. If `$HOME` (or `~/.mirage`) is not writable,
   the `mkdir`/`write_text` raise and the daemon crashes during lifespan startup.
   (This is correct per the repo rule "never silently swallow exceptions" — the
   fix is to make the location configurable, not to wrap it in `try/except`.)

3. **Unwritable / unset `$HOME` is common** in the target deployment shapes for
   Mirage (agent sandboxes, containers), which makes this more than a corner case.

## Reproduction

Run the daemon with a home directory that isn't writable, e.g.:

```bash
HOME=/nonexistent python -m mirage.server ...   # or docker with read-only $HOME
```

Startup raises `PermissionError` / `FileNotFoundError` from `_write_pid_file()`.

## Proposed fix

Make the pid file location configurable, following the existing
`version_root` / `snapshot_root` pattern:

1. Add a resolver that picks the path in priority order:
   - explicit `build_app(pid_file=...)` argument, then
   - env var (e.g. `MIRAGE_PID_FILE` or a broader `MIRAGE_DATA_DIR`/`MIRAGE_HOME`),
     added to `python/mirage/server/env.py`, then
   - the current `Path.home() / ".mirage" / "daemon.pid"` as the default.
2. Have both `app.py` and `cli/daemon.py` read from that single resolver so they
   never diverge.
3. Consider a broader `MIRAGE_HOME` / `MIRAGE_DATA_DIR` that also backs
   `version_root` and `snapshot_root` defaults, so the whole `~/.mirage` tree can
   be relocated in one setting.

## Parity note

Check the TypeScript daemon for the same hardcoded pid/home assumption and mirror
whatever fix lands here (see the Python/TypeScript parity guidance in `CLAUDE.md`).

## Verdict

Confirmed valid. The observation is correct: the pid file path is hardcoded to
`Path.home()`, is not always writable, and should be user-configurable.

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Actual behavior

```shell

```

### Mirage version or commit

_No response_

### Python version

_No response_

### Node.js version

_No response_

### Operating system

None

### Browser

_No response_

### Additional context

_No response_

ki3nd:
Closes strukto-ai/mirage#460

## Problem

The daemon pid file path was hardcoded to `Path.home() / ".mirage" / "daemon.pid"` in **two** places (`python/mirage/server/app.py` and `python/mirage/cli/daemon.py`, plus the TypeScript equivalents) and written **unconditionally** at ASGI lifespan startup. On any environment where `$HOME` is unset or not writable (containers, sandboxed users, `nobody`, immutable rootfs) daemon startup crashed with `PermissionError` / `FileNotFoundError`, and there was no way to override the location — inconsistent with the sibling `version_root` / `snapshot_root` which are already configurable via arg and env var.

## Fix

Make the location configurable, following the existing `version_root` / `snapshot_root` pattern.

- New shared resolver `server/paths.py`: `mirage_home()`, `pid_file_path()`, `default_version_root()`, `default_snapshot_root()`.
- Two new env vars: **`MIRAGE_HOME`** (base dir, defaults to `~/.mirage`) and **`MIRAGE_PID_FILE`** (pid-file-specific override).
- Pid file resolution order: `build_app(pid_file=...)` → `$MIRAGE_PID_FILE` → `$MIRAGE_HOME/daemon.pid` → `~/.mirage/daemon.pid`.
- `MIRAGE_HOME` also backs the `repos/` and `snapshots/` defaults, so the whole `.mirage` tree relocates with one variable. The existing `MIRAGE_VERSION_ROOT` / `MIRAGE_SNAPSHOT_ROOT` still take precedence when set.
- Both `app.py` and `cli/daemon.py` now read the single resolver, so the two copies can never diverge.
- No `try/except` swallowing added — per the repo rule, the fix is configurability, not hiding the error.

Mirrored in TypeScript per the repo parity guidance: `server/src/env.ts`, `paths.ts`, `app.ts`, `index.ts`, `bin/daemon.ts`, and `cli/src/daemon.ts`.

**Scope note:** `auth_token` and `config.toml` still key off `Path.home()`; left as a follow-up (agreed during design).

## Tests

- **Python:** extended `tests/server/test_paths.py` (env precedence, explicit-arg wins, roots follow `MIRAGE_HOME`) + new `tests/server/test_app.py` (`build_app` pid wiring, write/mkdir/remove pid = the reproduction). `tests/server` + `tests/cli`: **228 passed, 1 skipped**.
- **TypeScript:** extended `packages/server/src/paths.test.ts`. server **154 passed**, cli **51 passed** (including e2e that spawns the daemon and reads the pid via the shared resolver). `tsc --noEmit` clean for server and cli.
- pre-commit: all Python hooks pass; prettier + eslint clean on changed TS files.


zechengz:
Pushed 900564dd on top (maintainer edit): completes the MIRAGE_HOME migration. The CLI spawn path still hardcoded ~/.mirage for the daemon log dir and auth token (and config.toml), so with MIRAGE_HOME set and an unwritable $HOME the spawn crashed before the lifespan fix could help; those now route through mirage_home() / defaultTokenFile(). Also: relative MIRAGE_HOME / MIRAGE_PID_FILE values are absolutized, TS buildApp gains a pidFile option resolved once (mirroring py build_app / app.state.pid_file, and write/remove now share one resolved path), and the server barrel exports only the consumed path helpers. Both languages, tests added (py 236, ts server 161 / cli 52, tsc + pre-commit green).

zechengz (approved):
Thanks! :)

PR #467, opened 2026-07-11
fix(databricks): avoid redundant cat stat

sonhmai:
## Summary

- let command backends opt out of `cat`'s metadata preflight and disable it for Databricks Volumes in Python and TypeScript
- make a successful explicit-path `cat` use one file download instead of metadata plus download
- after a failed file download, probe the directory endpoint so directories report `Is a directory` while missing paths remain `No such file or directory`
- add regression coverage for request counts and error semantics

## Behavior

Before, `cat /dbx/file.txt` fetched metadata whose result was discarded, then downloaded the file. Removing that call alone would retain the existing incorrect directory error because the Databricks Files API returns 404 when a directory is downloaded.

After this change, successful file reads take the single-download fast path. Only a failed file download performs the directory probe needed to distinguish a directory from a missing path. Other backends keep their existing `cat` preflight behavior.

## Verification

- Python Databricks `cat`, generic command, and stream tests
- TypeScript generic `cat`, generic-bind, and Databricks stream tests

### DIFF
--- a/python/mirage/commands/builtin/databricks_volume/__init__.py
+++ b/python/mirage/commands/builtin/databricks_volume/__init__.py
@@ -41,6 +41,7 @@ _DATABRICKS_CMD_OPS = CommandIO(
     stat=_stat,
     is_mounted=lambda a: True,
     local=False,
+    cat_preflight_stat=False,
     write=_write,
     exists=_exists,
     mkdir=_mkdir,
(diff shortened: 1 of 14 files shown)


sonhmai:
cc @zechengz 

Closing this one. It has fallen far behind main.

The two things it set out to fix went different ways. 

1. cat on a directory now correctly says "Is a directory" since [#477](https://github.com/strukto-ai/mirage/issues/477). 
2. The extra metadata call before each download is still there, and it is now tracked across all read commands in [#1023](https://github.com/strukto-ai/mirage/issues/1023), with a first fix in [#1036](https://github.com/strukto-ai/mirage/issues/1036).

Getting cat down to a single request will need a different approach than this PR took what is better written as an issue and link.

PR #473, opened 2026-07-11
feat(daemon): config.toml path settings + `mirage config` command

### LINKED ISSUE #469: Support daemon path settings in config.toml and add a `mirage config` command
## Motivation

PR #466 (follow-up to #460) makes the `.mirage` tree relocatable via env vars: `MIRAGE_HOME`, `MIRAGE_PID_FILE`, plus the existing `MIRAGE_VERSION_ROOT` / `MIRAGE_SNAPSHOT_ROOT`. Today these are env-only. The CLI config file (`$MIRAGE_HOME/config.toml`, `[daemon]` table) only holds connection settings (`url`, `socket`, `auth_token`, `idle_grace_seconds`), and there is no CLI command to manage it: users hand-edit the TOML.

Docker supports this via files on both sides: the daemon reads `/etc/docker/daemon.json` (which has `data-root` and `pidfile` keys), and the client reads `~/.docker/config.json`. Git goes further with a `git config set/get/list` command. We can adopt both halves.

## Proposal

**1. New `[daemon]` keys in config.toml** (docker `daemon.json` model):

- `pid_file`
- `version_root`
- `snapshot_root`

Precedence per key, most specific wins: env var > config.toml > `$MIRAGE_HOME/<default>`.

Note the bootstrap exception: `MIRAGE_HOME` itself can never be a config.toml key, because the config file location is derived from it (docker has the same rule, `DOCKER_CONFIG` is env-only).

**2. A `mirage config` subcommand** (git model):

- `mirage config list`
- `mirage config get <key>`
- `mirage config set <key> <value>`
- `mirage config unset <key>`

Thin typer wrapper over `mirage/cli/settings.py`, writing `$MIRAGE_HOME/config.toml`.

## Wiring question

The server process currently never reads config.toml; `load_daemon_settings` is CLI-side and the daemon resolves everything from env at startup. Two options:

- (a) the spawner translates config values into env vars for the child (`_spawn_daemon` already builds the child env), small change, but only covers CLI-spawned daemons;
- (b) the daemon reads config.toml itself, docker style, which also covers systemd/manually launched daemons.

Leaning (b) for the same reason docker chose it, with flags/env still winning over the file.

## Parity

Mirror in TypeScript per repo convention: `packages/cli` settings + a matching `config` command, and the server-side file read if we go with (b).

ki3nd:
Closes strukto-ai/mirage#469 (follow-up to #460 / #466).

## What

Two capabilities, mirrored in Python and TypeScript:

1. **`config.toml` `[daemon]` now carries daemon path settings** — `pid_file`, `version_root`, `snapshot_root` (previously env-only). The daemon reads them itself at startup (Docker `daemon.json` model), so systemd/manually-launched daemons are covered, not just CLI-spawned ones.
2. **`mirage config` command** — `list` / `get` / `set` / `unset` over `$MIRAGE_HOME/config.toml`, so users stop hand-editing TOML (Git `git config` model).

## Precedence (per key, most specific wins)

```
explicit arg > env var > config.toml [daemon] > $MIRAGE_HOME/<default>
```

| key           | env var                | config key     | default                   |
| ------------- | ---------------------- | -------------- | ------------------------- |
| pid_file      | `MIRAGE_PID_FILE`      | `pid_file`     | `$MIRAGE_HOME/daemon.pid` |
| version_root  | `MIRAGE_VERSION_ROOT`  | `version_root` | `$MIRAGE_HOME/repos`      |
| snapshot_root | `MIRAGE_SNAPSHOT_ROOT` | `snapshot_root`| `$MIRAGE_HOME/snapshots`  |

`MIRAGE_HOME` stays env-only (never a config key) — the config file location is derived from it (bootstrap rule, same as Docker's `DOCKER_CONFIG`).

## How

- **Leaf reader** `daemon_config` (`read_daemon_table(home)` / `readDaemonTable(home)` + pure `parseDaemonTable(text)` in TS) parses the `[daemon]` table with no dependency on the path resolvers — avoids a circular import. CLI settings reuse it.
- **Path resolvers** (`pid_file_path` / `version_root_path` / `snapshot_root_path`; TS `pidFilePath` / `versionRootPath` / `snapshotRootPath`) own the full precedence chain. The daemon entrypoint no longer reads root env vars itself.
- **Line-based writer** for `config set` / `unset` — surgical edit that preserves comments and unrelated lines (no new TOML dependency; strings quoted with `\`/`"` escaped, numeric `idle_grace_seconds` bare).
- `mirage config list` reports the **file** contents (not env-resolved runtime values).

Renames (no back-compat needed): `default_version_root`→`version_root_path`, `default_snapshot_root`→`snapshot_root_path` (and TS equivalents).

## Notes

- Path-setting changes take effect on the next daemon start/restart (no hot reload).
- No new runtime dependency; TOML read via stdlib `tomllib` (Py) / the existing flat reader (TS).

## Testing

- Python: unit tests for `daemon_config`, full precedence matrix for all three resolvers, writer round-trips (create/update/preserve-comments/unset/reject-unknown-key/quoting), and the `mirage config` subcommands.
- TypeScript: mirrored vitest suites (`daemon_config.test.ts`, `paths.test.ts`, `settings.test.ts`, `config.test.ts`).
- Full Python suite + TS build/test/typecheck + `pre-commit` all green.


zechengz:
Pushed 16 follow-up commits (`73a538de..a1b4a855`) that build the Docker-model consolidation on top of this PR, mirrored in Python and TypeScript:

**Validation (daemon.json model)**
- Shared key registry, `DaemonConfigError`, and `validate_daemon_table` live in the `daemon_config` leaf; the CLI imports them instead of keeping its own copy.
- The daemon refuses to start on unknown `[daemon]` keys or malformed TOML, naming the offenders. Enforced in `build_app`/`buildApp`, in the CLI spawner before forking, and at both CLI entry points (clean one-line stderr + exit 2; the Python entry point moved from `main:app` to `main:main` for this).
- Repairability: `config unset` accepts unknown keys and `config list` warns about them, so a file the daemon rejects can always be fixed from the CLI.

**Resolved view**
- `mirage config list --resolved` prints each key's effective value and origin (`env MIRAGE_X` / `file` / `default`), applying the same env > file > default precedence. `auth_token` is masked.

**Hardening**
- `config set`/`unset` chmod the file to 0600 (it may hold `auth_token`).
- Restored py/ts parity for `load_daemon_settings` with a missing explicit path, and fixed the quoted `str(KeyError)` message on unknown keys.

**New keys** (same precedence, validated)
- `port` (adds `MIRAGE_DAEMON_PORT` to Python for parity; spawn precedence env > config > url-derived > 8765)
- `allowed_hosts` (comma-separated, threaded into `resolve_allowed_hosts`)
- `auth_mode` + flat `jwt_alg`/`jwt_issuer`/`jwt_audience`/`jwt_pubkey_file`/`jwt_clock_skew`/`jwt_authorized_parties` via a new `table` parameter on `resolve_auth_config`. Raw secrets (`MIRAGE_AUTH_TOKEN`, `jwt_pubkey`) deliberately have no config key. Also fixed the spawner forcing `MIRAGE_AUTH_MODE=local` over a config-set `auth_mode`.

**Tests**
- Unit suites for all of the above in both languages.
- `integ/cli_config.sh` (wired into `test_cli.yml`): drives both CLIs through the full surface and diffs the outputs, so py/ts behavior is byte-identical by construction. Part A relocates `pid_file`/`version_root`/`snapshot_root`/`port` via `mirage config set`, runs a real workload (create, execute, commit, log, snapshot, readback), and asserts artifacts land in the configured dirs with nothing leaked to defaults, settings re-read on restart, and env-beats-config proven by where the repo lands. Part B covers 0600 perms, refuse-to-start plus repair, malformed TOML, resolved origins/masking, allowed_hosts via real Host-header requests, and the jwt key rules.

**Docs**: `docs/home/cli.mdx` (worked `--resolved` example, key table, dockerd comparison), `docs/typescript/server-and-cli.mdx`, CLI README env table.

zechengz (approved):
Thanks :)

PR #475, opened 2026-07-11
feat(nextcloud): server-side find using Files Search API (partial predicates + fallback)

### LINKED ISSUE #472: feat: Use Nextcloud's server-side Files Search API to accelerate find on WebDAV resources
The current Nextcloud implementation of `find` performs a recursive WebDAV scan of the directory tree and evaluates predicates client-side.

For large Nextcloud instances, this can become significantly slower than necessary, especially for common searches such as:

```bash
find "/Documents" -type d -name "Invoices"
find "/Projects" -name "*.pdf"
find "/Accounting" -mtime -30
```

Nextcloud already exposes a server-side Files Search API that performs indexed metadata searches. Mirage could leverage this API whenever the search expression can be represented by the server, falling back to the current implementation only when necessary.

## Motivation

The current implementation roughly follows this pattern:

1. Recursively scan the subtree over WebDAV.
2. Receive every filesystem entry.
3. Apply `-name`, `-type`, `-size`, `-mtime`, etc. in Python.

While this is simple and works correctly, it means even highly selective searches may require traversing a very large directory tree.

For remote storage, network traversal is the dominant cost.

## Proposed approach

When the resource is Nextcloud and the expression consists only of predicates supported by Nextcloud's Files Search API, use server-side search instead of recursive WebDAV traversal.

For example:

```bash
find "/Accounting" -type d -name "Documenti 2024"
```

could become:

- search term: `Documenti 2024`
- path prefix: `/Accounting`
- directory filter
- pagination handled by Mirage

Mirage would then perform any remaining checks locally, such as:

- exact basename matching (if desired)
- `-maxdepth`
- `-mindepth`
- unsupported predicates

If the expression cannot be represented server-side, Mirage should transparently fall back to the existing implementation.

## Benefits

- Dramatically fewer WebDAV requests
- Much lower latency on large Nextcloud instances
- Better scalability for AI agents and automation workflows
- Reuses Nextcloud's indexed metadata instead of recursively traversing remote storage

## Possible implementation

1. Detect whether the parsed predicate tree is representable by the Files Search API.
2. Issue a request to Nextcloud's Files Search endpoint.
3. Paginate through results.
4. Apply any remaining client-side predicates.
5. Fall back to the existing recursive scan when required.

This preserves the current behavior while significantly improving performance for common search operations.

nnayz:
## Summary

Accelerates Nextcloud `find` by using Nextcloud's server-side Files Search API for fully representable predicates in both Python and TypeScript.

Supported searches include `-name` / `-iname`, `-type f|d`, `-size`, `-mtime`, and fully representable boolean combinations. Mirage applies exact predicate, scope, and depth checks to the returned metadata. Expressions containing unsupported predicates transparently use the existing scoped recursive scan.

## Changes

- Adds typed, paginated WebDAV SEARCH clients for Python and TypeScript.
- Mirrors Nextcloud `find` behavior and architecture across both runtimes.
- Keeps unsupported expressions on the correctness-preserving scan fallback.
- Avoids recursive traversal for `-maxdepth 0` by evaluating only the starting path.
- Adds unit coverage for query compilation, pagination, response validation, fallback behavior, exact post-filtering, and scan avoidance.
- Adds declarative Nextcloud coverage to `integ/unix/find.json` and a TypeScript Nextcloud integration adapter.

## Performance

Representative live Nextcloud measurements showed server-search queries around 140 ms versus roughly 900 ms for recursive fallback on the integration fixture. The `-maxdepth 0` path dropped from roughly 830 ms to under 1 ms with a warm metadata cache.

Performance correctness is guarded by request-shape tests: supported searches cannot invoke recursive scans, zero-depth searches cannot list recursively or invoke Files Search, and pagination asserts exact SEARCH request counts.

## Verification

- Full pre-commit suite passes.
- Full TypeScript test suite passes.
- Python Nextcloud declarative integration: 1,218 passed.
- TypeScript Nextcloud declarative integration: 1,218 passed.

Fixes #472


nnayz:
Fixes #472

nnayz:
Updated PR with comment reverts (per review). Pre-commit (Python) and relevant tests pass.

zechengz (approved):
Cleaned and modified. Should be good now. Let me know if there are other issues.

PR #493, opened 2026-07-13
integ(fuse): size-unknown probes + advisory WinFsp job

zechengz:
PR D of the FUSE consolidation arc (#486 #488 #489 #491).

## What

- **Deterministic size-unknown coverage.** The fuse integ (both languages) now mounts a RAM workspace through a stat wrapper that strips sizes (`SizelessOps` proxy in python, a `ws.fs.stat` wrapper in TS), simulating API-backed resources (Linear, Slack, Trello, ...). Shared truth lines assert the /proc-style semantics end to end through a real kernel mount: `stat` 0 before open, full content on read, real size served after open.
- **Advisory Windows job.** `integ-fuse-windows` installs WinFsp via choco and runs the python fuse battery on `windows-latest`. It is `continue-on-error: true` and not in the integ gate: it exists purely to surface what actually breaks on Windows so WinFsp support can be built against real signal instead of guesses.

## Verified

Both probes run live on macOS (kext) with output matching the truth lines exactly:

```
api_stat_preopen=0
api_cat={"messages": 2}
api_size_postread=16
```

The Linux CI fuse job validates the same lines through libfuse; the Windows job's first run is expected to be rough, that is its job.


zechengz:
Update since the PR body: the advisory job did its work faster than expected. Three rounds of signal (mountpoint-in-use, os.getuid crash, CRLF + stat-opens-a-handle) each got a researched fix, and the Windows battery now passes outright: all 12 truth lines matched over a real WinFsp mount, including two simultaneous mounts and the size-unknown probe. WinFsp conventions are documented in CLAUDE.md (nonexistent mountpoint, existence-based readiness, no fusermount, uid=-1/gid=-1 ownership, guarded getuid, and the Windows quirk that stat opens a handle so size-unknown files hydrate on first stat). The job stays advisory for now: one green run is not yet a support statement.

PR #511, opened 2026-07-15
fix: correct path.strip() calls and empty-input edge cases in commands

fix2015:
- Fix path.strip() to path.strip('/') in collision error message (was stripping whitespace instead of slashes)
- Fix sort, rev, fold to return empty bytes for empty input instead of spurious newline
- Fix expand -i to match tabs preceded by spaces in leading whitespace
- Fixes applied to both Python and TypeScript implementations


zechengz:
Pushed a follow-up commit to this branch (thanks for the fixes!). Two things:

**1. CI was red on formatting.** The edits weren't run through the repo's `yapf`/`prettier`, so the `pre-commit` job failed on those hooks. I reformatted `fold.py`, `rev.py`, `fold.ts`, and `sort.ts` to match.

**2. The `fold` empty-input fix had an edge-case regression.** `fold` splits with `splitLinesNoTrailing`, which collapsed **both** `""` and `"\n"` to `[]`. So the new `lines.length === 0` guard emitted empty output for a single blank line, where GNU (and the Python side) emit `\n`:

| Input | GNU / Python `fold` | TS `fold` before this commit |
|---|---|---|
| `""` (empty) | `` | `` ✅ |
| `"\n"` (one blank line) | `\n` | `` ❌ |
| `"a\n"` | `a\n` | `a\n` ✅ |

Root cause: `split_lines("\n")` → `[""]` in Python, but `splitLinesNoTrailing("\n")` → `[]` in TS. Fixed by having it distinguish truly-empty from blank-only, matching the sibling `splitSortLines` (which is why `sort` in this PR was already correct):

```ts
function splitLinesNoTrailing(text: string): string[] {
  if (text === '') return []
  const stripped = text.endsWith('\n') ? text.slice(0, -1) : text
  return stripped.split('\n')
}
```

The other fixes (`expand -i` leading blanks, `sort` blank-line, `rev`/`fold` empty, chroma `strip('/')`) all look correct in both languages.

Follow-up worth adding: a regression test for `printf '\n' | fold` (and `rev`/`sort`) to lock this edge case in — I didn't push one since I couldn't run the TS suite in my worktree.

zechengz:
Added integ coverage for these edge cases (commit `5210fbb8`), in both `integ/cases.py` and `integ/cases.ts`:

```
fold_empty_file   echo foldemptyf=$(fold /data/empty.txt | wc -c)  -> foldemptyf=0
sort_empty_file   echo sortemptyf=$(sort /data/empty.txt | wc -c)  -> sortemptyf=0
rev_empty_file    echo revemptyf=$(rev  /data/empty.txt | wc -c)   -> revemptyf=0
fold_blank_in     echo foldblankin=$(echo | fold | wc -c)          -> foldblankin=1
sort_blank_in     echo sortblankin=$(echo | sort | wc -c)          -> sortblankin=1
rev_blank_in      echo revblankin=$(echo | rev  | wc -c)           -> revblankin=1
expand_i_spacetab echo expandisp=$(printf ' \tX\n' | expand -i | wc -c) -> expandisp=10
```

Byte counts are labelled so `truth.txt` can assert them as fixed substrings (`check_lines.sh` ignores blank lines, so a bare `\n` isn't assertable directly). Each expected value **differs from the pre-fix output** (`foldemptyf=1`, `sortblankin=0`, `expandisp=4`), so the cases actually guard the regressions rather than just documenting current behavior.

**Does this cover all backends?** Yes. `fold`/`rev`/`sort`/`expand` are generic commands: the backend only supplies the bytes via `read_bytes`/stream, and all the empty-vs-blank logic lives in the generic command downstream of that. `integ/cases.py` is the shared case list every backend harness runs, so the empty-*file* cases exercise each backend's read path, while the blank-line (stdin) cases exercise the backend-independent generic logic. Verified green locally on **ram, disk, and redis**; CI runs the same `truth.txt` against s3 and the TS harness (`integ-ts`).

zechengz (approved):
Thanks for the contribution :) You can join our discord for better contribution related communication https://discord.com/invite/u8BPQ65KsS

PR #537, opened 2026-07-18
feat(version): whole-world commits via the .mirage/ control-plane subtree

zechengz:
A version commit now captures the whole workspace world, not just file content. Sessions, namespace nodes, and command history ride inside the tree under a reserved `.mirage/` control-plane subtree; checkout restores all of them alongside the files. Cache stays permanently excluded (derived and rebuildable).

## What changed

- `state_tree` / `stateTree`: the tree gains `.mirage/sessions.json`, `.mirage/namespace.json`, and per-session `.mirage/history/<session>.jsonl` next to `.mirage/meta.json`; `to_state` reads them back into the state dict, and commit meta carries `default_session_id`.
- History is sharded one jsonl per session (mirroring the live ObserverStore layout and the Claude Code convention), merged on read by stable timestamp sort (the `events()` contract). Idle sessions keep identical blobs across commits, so they dedup in the content-addressed store.
- `version_diff` / `status` stay files-only: `.mirage/` entries are filtered out of file diffs (the structured cross-plane state diff is a separate follow-up).
- Bug fix (py): `_restore_sessions` skipped sessions that already existed live, so checkout on a running workspace kept the live env/cwd/mount modes instead of the committed ones. The committed table now wins, matching the `replace_from_snapshot` contract.
- Bug fix (ts): `restoreSessions` called `adoptDefaultSession(undefined)` for states without the pointer; now guarded like the Python side, and `default_session_id` is honestly typed `string | undefined` through `MountArgs`.

## Tests

- Both langs: whole-world round trip (files + sessions + nodes + multi-session history), control-plane files never leak into mount files, and an end-to-end "checkout restores the whole world" test (file content, symlink, session env/mount-mode narrowing, history rewound).
- Branch merged up to latest main. Py version+workspace suites green; TS server (17) + core workspace (1446) tests green; `pnpm -r typecheck` clean; pre-commit all-files clean.


zechengz:
Closing to rework the commit history layout (per-session sharding instead of a single history.jsonl) before landing.

PR #583, opened 2026-07-20
feat(google): per-service API base URL overrides for `GoogleConfig`

khj809:
## Background

`GoogleConfig` recently gained a single `api_base` field so that every Google backend
(drive/docs/sheets/slides/gmail + the OAuth token endpoint) can be pointed at a fake server for
testing. It rewrites the host for **all** services at once, deriving each service's URL by
appending that service's real path suffix to the one base.

## Problem

A single host only works when every service can be told apart by path. Google's real APIs live on
distinct hosts, so collapsing them onto one `api_base` produces paths that are ambiguous — or
outright colliding — at the subpath level:

| service | derived from `api_base` |
|---------|-------------------------|
| drive   | `{api_base}/drive/v3`   |
| docs    | `{api_base}/v1`         |
| slides  | `{api_base}/v1`  ← **identical to docs** |
| sheets  | `{api_base}/v4`         |
| gmail   | `{api_base}/gmail/v1`   |

`docs_base` and `slides_base` both resolve to `{api_base}/v1`, so a fake server hosting both under
one origin has **no way to distinguish a Docs request from a Slides request** — a concrete,
unavoidable collision. More broadly, `api_base` offers no subpath-level control: you can't point an
individual service somewhere else, and you can't line mirage up with a mock whose per-service paths
don't match the derived ones.

## Change

Add optional **per-service base-URL overrides** to `GoogleConfig`. Each takes precedence over
`api_base`, then falls back to the `api_base` derivation, then the real Google host — so nothing
changes unless you set them:

| field                    | overrides                 | fallback when unset                      |
|--------------------------|---------------------------|------------------------------------------|
| `token_url`              | OAuth token endpoint      | `{api_base}/token` → real `TOKEN_URL`    |
| `drive_api_base`         | Drive                     | `{api_base}/drive/v3` → real `DRIVE_API_BASE`        |
| `drive_upload_api_base`  | Drive upload              | `{api_base}/upload/drive/v3` → real `DRIVE_UPLOAD_BASE` |
| `docs_api_base`          | Docs                      | `{api_base}/v1` → real `DOCS_API_BASE`              |
| `sheets_api_base`        | Sheets                    | `{api_base}/v4` → real `SHEETS_API_BASE`              |
| `slides_api_base`        | Slides                    | `{api_base}/v1` → real `SLIDES_API_BASE`              |
| `gmail_api_base`         | Gmail                     | `{api_base}/gmail/v1` → real `GMAIL_API_BASE`        |

Resolution order in every base helper is:
`service-specific override → api_base-derived → real default`.

With these, Docs and Slides can resolve to genuinely different bases (e.g. `…/docs/v1` and
`…/slides/v1`), which removes the collision, and a mock whose paths differ from the derived ones
can be matched exactly.

### Notes

- `token_url` is the **full** token endpoint URL and is used verbatim (it replaces the whole
  `TOKEN_URL`; no `/token` is appended) — the OAuth endpoint is a single URL, not a base. The
  remaining fields are service bases that callers append paths to.
- `api_base` behavior is unchanged, and with nothing set every service still points at its real
  Google host. Fully backward compatible.

## Tests

- Per-service overrides take precedence over `api_base` for every service.
- Fallback chain: explicit override → `api_base` derivation → real default.
- Existing default-host and single-`api_base` tests remain unchanged and green.

### DIFF
--- a/python/mirage/core/google/_client.py
+++ b/python/mirage/core/google/_client.py
@@ -32,36 +32,56 @@ TOKEN_BUFFER_SECONDS = 300
 
 
 def token_url(config: GoogleConfig) -> str:
+    if config.token_url:
+        return config.token_url
     return f"{config.api_base}/token" if config.api_base else TOKEN_URL
 
 
 def drive_base(token_manager: "TokenManager") -> str:
-    base = token_manager.config.api_base
+    config = token_manager.config
+    if config.drive_api_base:
+        return config.drive_api_base
+    base = config.api_base
     return f"{base}/drive/v3" if base else DRIVE_API_BASE
 
 
 def drive_upload_base(token_manager: "TokenManager") -> str:
-    base = token_manager.config.api_base
+    config = token_manager.config
+    if config.drive_upload_api_base:
+        return config.drive_upload_api_base
+    base = config.api_base
     return f"{base}/upload/drive/v3" if base else DRIVE_UPLOAD_BASE
 
 
 def docs_base(token_manager: "TokenManager") -> str:
-    base = token_manager.config.api_base
+    config = token_manager.config
+    if config.docs_api_base:
+        return config.docs_api_base
+    base = config.api_base
     return f"{base}/v1" if base else DOCS_API_BASE
 
 
 def slides_base(token_manager: "TokenManager") -> str:
-    base = token_manager.config.api_base
+    config = token_manager.config
+    if config.slides_api_base:
+        return config.slides_api_base
+    base = config.api_base
     return f"{base}/v1" if base else SLIDES_API_BASE
 
 
 def sheets_base(token_manager: "TokenManager") -> str:
-    base = token_manager.config.api_base
+    config = token_manager.config
+    if config.sheets_api_base:
+        return config.sheets_api_base
+    base = config.api_base
     return f"{base}/v4" if base else SHEETS_API_BASE
 
 
 def gmail_base(token_manager: "TokenManager") -> str:
-    base = token_manager.config.api_base
+    config = token_manager.config
+    if config.gmail_api_base:
+        return config.gmail_api_base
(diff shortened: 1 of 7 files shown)


zechengz:
Thanks for working on this, and sorry for the confusion. The original api_base design was a mistake on my side. IMO Google’s service endpoints should remain fixed and should not be exposed as user configuration.
For integration tests, the better approach is to redirect those requests to our mock entirely within the test setup.
Can you allow the maintainer edits so I can quickly push an edit? Thanks in advance.

zechengz:
close because of https://github.com/strukto-ai/mirage/pull/584 and cannot force override because of the organization level clone?

khj809:
@zechengz 
Sorry, I was late to reply.
You said that the `api_base` is only for internal test purpose, but I'm the user who really want this as a public feature.
I'm working on building a [mock service](https://github.com/brekkylab/enterprise-mock) which provides read-only APIs from several enterprise providers such as Google and Slack, and I want to add a [mirage integration example](https://github.com/brekkylab/enterprise-mock/tree/main/examples/using-mirage) with pointing to my mock server's url. (Currently this example monkeypatches mirage to point to our desired URLs). I'm also waiting for https://github.com/strukto-ai/mirage/pull/578, which also introduces `base_url` for configuring the fake Slack server.
So I'd like to ask if you can take these use cases into consideration.

zechengz:
@khj809 I see, thanks for the information.
If possible maybe you can join our discord channel for faster discussion https://discord.com/invite/u8BPQ65KsS
Yes I actually considered this before as mock server is quite useful for LLM post training etc
Let me think about this tomorrow.
Thanks! :)

zechengz:
@khj809 reverted in latest main
will have a new release tomorrow

PR #605, opened 2026-07-22
feat(agents): canonical ops layer + stream background job output

zechengz:
Two independent changes to the agent-facing job and backend layers.

## 1. Canonical ops layer for agent backends

`mirage/agents/types.py` and `ops.py` hold one set of result types and one set of parsers. Each harness adapter becomes a thin field-mapper over them rather than carrying its own copy of the parsing (the two `_convert.py` files shrink by 70 lines).

Two protocols were only partly implemented and are now complete:

- **deepagents** `SandboxBackendProtocol` was missing `ls_info`, `glob_info` and `grep_raw` plus async variants. They are now projections of the existing Result-returning ops.
- **pydantic-ai** `BackgroundSandboxProtocol` was not implemented at all. `mirage/agents/background.py` wraps the workspace job table and adds the per-reader cursor the protocol requires, so `read_background` returns only what is new since the last call.

## 2. Background job output streams instead of arriving at exit

A background job materialized its whole stdout inside the task and returned it as the task result, which `JobTable._refresh` could only read through `task.result()`. That value does not exist until the task is done, so output was invisible for the entire run. Measured, for a job printing one line per second:

```
0.5s  running=True   new_stdout=''
2.5s  running=True   new_stdout=''
3.0s  running=False  new_stdout='tick-1\ntick-2\ntick-3\n'
```

The runner now drains the stream chunk by chunk into the job, so any command that streams is observable while it runs.

Known limitation, deliberately not addressed here: compound constructs (`for`, `while`, `;` lists) still batch, because they compute their output eagerly inside `execute_node` before yielding any of it. That is an executor-level change.

Three related fixes fall out:

- `JobTable.submit` takes a runner factory and creates the task itself, so the job exists before output can arrive. Creating the task first left a window where output had nowhere to go.
- `kill` appends its notice to stderr instead of overwriting it, which previously discarded everything the job had already printed.
- Detached tasks (no job table) are held in a set, since the event loop keeps only a weak reference and could otherwise collect them.

## Tests

New: `tests/agents/test_ops.py`, `tests/agents/test_background.py`, `tests/workspace/executor/test_jobs.py` (including a gate-based test that output is visible before the stream ends). Updated: the deepagents and pydantic-ai backend tests, and `tests/shell/test_background_jobs.py` for the runner-factory signature.

Python only; no TypeScript behavior changes.

### DIFF
--- a/python/mirage/agents/background.py
+++ b/python/mirage/agents/background.py
@@ -1,0 +1,161 @@
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+
+from dataclasses import dataclass
+
+from mirage.agents.io_text import decode
+from mirage.shell.job_table import Job, JobStatus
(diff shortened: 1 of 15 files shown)


zechengz:
bad design

Issue #610, opened 2026-07-22
Coreutils gaps (easy, priority-tiered): pure commands + formatting/alias flags + --version (part of #597)


zechengz:
Part of #597. The **easy** GNU Coreutils slice: pure, deterministic commands and formatting / long-alias / `--version` flags — no new VFS, metadata, symlink, follow-loop, or job-table dependencies. Output is a pure function of input bytes + flags.

Reorganized by **priority for agent workloads** (how often agents actually reach for the command). Almost everything here is `easy` difficulty; a few items that need a small engine are tagged `med`. Partitioned **by command** (a command's whole flag set lives in one issue) so this and #609 never touch the same command. Check a box only when Python and TS match GNU coreutils 9.x and focused tests cover success, error shape, and exit status. Exhaustive per-flag spellings per command are in #597.

---

## Tier 1 — high priority (used constantly)

- [x] `env` (new) — print environment; `env NAME=val cmd` run form
- [x] `cat` — `-n/-b/-A/-e/-t/-v/-s/-u` + long forms
- [x] `head` — `--bytes/--lines/-q/-v/-z`
- [x] `wc` — `--bytes/--chars/--lines/--words/--max-line-length/--total` aliases
- [x] `cut` — `-b/-n/-s/-w/-F/-O/-z` + long forms
- [x] `sort` — remaining flag aliases `-c/-d/-g/-i/-m/-o/-r/-u/-z` + long forms (KEYDEF `-k` already shipped in #607)
- [x] `tr` — `-C/-t/--complement/--delete/--squeeze-repeats/--truncate-set1`
- [x] `uniq` — `-D/-z/--group/--all-repeated/--check-chars/--skip-chars/--skip-fields`
- [x] `tee` — `-a/-i/-p/--output-error`

## Tier 2 — medium priority

- [x] `truncate` (new) — `-s` grow/shrink a file — done (#619).
- [x] `od` (new, `med`) — octal/hex/char dump (full `-t`/`-A`/`-j`/`-N` format spec) — done (#619; #622 completed it: the spec declared `positional=(PATH,)` i.e. exactly one FILE, now `rest=PATH` so `od FILE...` is declared as well as implemented).
[...]

PR #618, opened 2026-07-23
feat(coreutils): env builtin + tr -C/-t + tee --output-error

zechengz:
Finishes issue #610 Tier 1 (after #613 shipped cat/head/wc/cut/sort/uniq).

## env (new shell builtin)
- Print environment (unsorted, insertion order — GNU-correct), `-i/--ignore-environment`, `-u NAME/--unset`, `-0/--null`.
- Run form `env NAME=val cmd`: runs the command under the modified environment, forwards stdin, restores the session env after.
- Errors exit 125 with GNU's `Try 'env --help'…` hint; unknown `--long` → `unrecognized option`.
- Deliberate divergences: options must precede assignments (no getopt permutation); missing command keeps the shell's own exit 127 (message differs from GNU's `env: 'x': No such file`).

## tr
- Adds `-c/-C/--complement`, `-d/--delete`, `-s/--squeeze-repeats`, `-t/--truncate-set1`.
- `-t` truncates set1 to len(set2) (`abcde`+`xy` → `xycde`); default pads set2. `-C` ≡ `-c` (ASCII).
- Generic migrated to the `FlagView`/`parseFlags` struct pattern; builders are wiring-only.

## tee
- Adds `-a/--append`, `-i/--ignore-interrupts`, `-p`, `--output-error[=MODE]`.
- MODE validated against `{warn, warn-nopipe, exit, exit-nopipe}` (bad → exit 1, GNU message). `-i`/`-p` accepted no-ops.
- Validator shared with the bespoke s3/gridfs tee so backends stay consistent.

## Tests
- Unit: py `test_tr.py`/`test_tee.py`/`test_vars.py`; ts `ram/tr.test.ts`/`generic/tee.test.ts`/`workspace/env_builtin.test.ts`.
- Integ: `integ/unix/tr/t.json`, `integ/unix/tee/output_error.json`, `integ/bash/builtin/env.json` (16 cases). Both languages 1718/1718 on ram.
- All GNU behaviors pinned against `debian:stable-slim`. pre-commit + tsc + mypy clean.


zechengz:
@codex review

zechengz:
Addressed the Codex review in e1934202b:

- **tr/tee direct callers (P2)** — fixed. `test_phase_m.py` and `test_phase_q.py` now pass `flags={"d": True}` / `{"s": True}` / `{"a": True}` instead of the removed keyword params. Full `pytest` is green (the only local failures are the pre-existing FUSE/redis env-dependent tests).
- **env `-0`/`--null` with a command (P2)** — fixed. Now exits 125 with `env: cannot specify --null (-0) with command` before dispatching, in both languages. Integ: `env_null_with_command_rejected`.
- **env lone `-` (P2)** — fixed. A bare `-` implies `-i`, so `env -` prints an empty environment and `env - A=1` prints only `A=1`, in both languages. Integ: `env_lone_dash_implies_ignore`.

**tee `--output-error` mode application (P2)** — the value is validated (bad modes exit 1 with GNU's message) but intentionally not applied per-mode. mirage `tee` writes to a **single** sink plus stdout, so the four modes (`warn`/`warn-nopipe`/`exit`/`exit-nopipe`) are observationally identical here: the mode's only real distinction is whether to *continue to further output files* vs *exit early* on a write error, which is meaningless with one output, and the `nopipe` variants concern `EPIPE`, which the VFS layer doesn't surface. stdout passthrough is preserved on success. Applying per-mode diagnostics would add an error path that can't be exercised across the backend targets. Happy to implement it if you'd prefer the mode stored and a diagnostic emitted on write failure.

All 22-target integ battery and unit tests pass in both Python and TS.

zechengz:
Fixed the tee `--output-error` finding in a361eb358 (shared `write_output` helper, py+ts): on a write failure tee now still copies stdin to stdout, prints a GNU-style `tee: <path>: <error>` diagnostic, and exits non-zero. With a single output sink the four modes (`warn`/`warn-nopipe`/`exit`/`exit-nopipe`) collapse to this behavior. Covered by unit tests in both languages (write success + write-error paths). All 5 review threads resolved; battery 1749/0 both languages.

PR #619, opened 2026-07-23
Add Tier 2 coreutils support

zechengz:
Part of #610.

## Summary

- add Python/TypeScript implementations for Tier 2 coreutils, including `truncate`, `numfmt`, and `od`
- extend existing command specs and generic implementations with the Tier 2 flag set while preserving Python/TypeScript parity
- route attached optional path flags to the owning mount and align checksum, fmt, mktemp, shuf, and parser behavior
- add 47 JSON integration cases covering the new commands and flags on RAM and disk targets

## Testing

- `./python/.venv/bin/pre-commit run --all-files`
- `pnpm build`
- targeted Python command/spec tests: 209 passed
- TypeScript Node suite: 1,655 passed, 97 skipped
- Python RAM/disk integration: 3,516 passed
- TypeScript RAM/disk integration: 3,516 passed
- Python/TypeScript RAM/disk parity: 3,516 compared, 0 mismatches
- full Python suite: 10,012 passed, 355 skipped; unrelated environment failures in CLI daemon startup, Redis (no URL), and macOS FUSE
- full TypeScript core suite: 4,831 passed; unrelated live-network and Pyodide PIL failures

PR #622, opened 2026-07-23
Fully close the od and numfmt review findings from #619

zechengz:
Follow-up to #619, which merged before these two review findings were fully closed.

## od takes FILE..., but the spec said one file

#619 fixed the `od` generic to concatenate every operand, but the spec kept
`positional=(Operand(PATH),)`, meaning "exactly one FILE". It is the only
`FILE...` command declared that way; eleven siblings use `rest=`. The generic
worked only because the parser falls back to the last positional kind for extra
operands, and the generated contract in `spec/` misdescribed the command.

Declares `rest=PATH` in both languages and regenerates the specs.

## numfmt scaled output did not match GNU

The review comment's own example asked for `1.0K foo`; we printed `1K foo`.
GNU prints `1.0k foo`. Implements GNU's actual rule in both languages:

- round away from zero, keeping one decimal only while the scaled value is below 10
- re-check the unit afterwards, since rounding can push a value back over the base
  (`999.4` becomes `1000` becomes `1.0k`)
- render through a half-even printf, which is why an unscaled `2.5` prints `2`
- SI spells kilo lowercase, IEC stays uppercase
- `--from` now accepts lowercase suffixes (`2k`)

Both implementations were diffed against GNU coreutils 9.8 across 37 cases
covering si, iec, iec-i and none, with zero differences.

## Testing

- `pre-commit run --all-files`: 18/18 pass
- spec drift check (`gen_specs.py` + `gen-specs.ts` + `git diff --exit-code spec/`): clean
- TypeScript core 4857 passed, node 1748 passed
- Python command and spec suites pass
- integ ram + disk cross-language parity: 3604 pairs, 0 mismatches

### DIFF
--- a/integ/crossmount/version/basic.json
+++ b/integ/crossmount/version/basic.json
@@ -1,0 +1,66 @@
+{
+  "cases": [
+    {
+      "id": "xm_version_spans_mounts",
+      "seq": 782,
+      "targets": [
+        "ram",
+        "disk",
+        "redis",
+        "opfs",
+        "s3",
+        "databricks",
+        "gridfs",
+        "s3-prefix",
+        "databricks-prefix",
+        "gridfs-prefix",
+        "hf",
+        "hf-prefix",
+        "dropbox",
+        "dropbox-root",
+        "onedrive",
+        "sharepoint",
+        "sharepoint-prefix",
+        "ssh",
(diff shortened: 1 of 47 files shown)


zechengz:
@codex review

PR #668, opened 2026-07-31
fix(jq): preserve output arity instead of guessing it from the program

zechengz:
## The bug

`jq_eval` / `jqEval` collapsed a program's output stream into a single value: one output returned bare, many returned as a *list*, zero as a `JQ_EMPTY` sentinel. A list of outputs is indistinguishable from one output that happens to be an array, so the printer recovered the difference from the program text: `has_top_level_spread` / `hasTopLevelSpread` scanned for a `[]` at bracket depth 0.

The text cannot answer that question. Any multi-output program with no literal `[]` printed a JSON array instead of one line per value:

| program | mirage before | GNU jq |
| --- | --- | --- |
| `jq -r '.a, .b'` | `[1,2]` | `1` / `2` |
| `jq -c 'range(3)'` | `[0,1,2]` | `0` / `1` / `2` |
| `jq -c '..'` | `[{...},"alice",30]` | three lines |

Neither libjq nor jq-wasm was wrong. Both returned the right arity and it was discarded immediately after the call. In TypeScript the correct output was already the string `"1\n2\n"` in `result.stdout`, then split, parsed, collapsed, and re-derived from the program text.

## The fix

`jq_eval` returns every output in order, `format_jq_output` prints one per line, and the heuristic plus both `JQ_EMPTY` sentinels are deleted (an empty list means no output). `eval_jsonl_stream` loops the outputs of each line as well. `[.a, .b]` still emits one value and prints one line, now for the right reason.

## Tests

`integ/unix/jq/comma.json`, 12 cases (seq 700000-700011): comma under `-c` / `-r` / default pretty-printing, comma over arrays vs a collector, per-document and slurped, `range`, `..`, through a pipe, from stdin, and over the jsonl stream path.

Expected output is pinned against GNU jq 1.7 in `debian:stable-slim`, not written from memory.

With arity preserved, `jq_eval(x, ".a")` returns `[1]`, and about 130 existing assertions read `== 1`. Instead of adding `[0]` everywhere, the unit tests go through `eval_one` / `evalOne` helpers that assert `len(outputs) == 1` and return it. That turned the migration into a detector: it flagged 13 tests whose programs were genuinely multi-output, and each moved to the arity-preserving call with its expected list unchanged.

`docs/home/bash.mdx` documented the old rule and is updated.

## Verification

- pre-commit clean and stable across a second run
- full Python suite exit 0, zero FAILED/ERROR
- TS core 5646, node 2000, browser 181, agents 175, server 210, cli 92, all passing; `pnpm -r typecheck` clean
- integ 3979 passed / 0 failed on both the python and typescript hosts for ram+disk, plus 1920 on redis
- no spec drift from either generator

## Known divergence, left alone

The `.jsonl` streaming fast path (taken when the path ends `.jsonl`/`.ndjson` and the program starts `.[]`) treats the file as one array and always prints compact, ignoring the absence of `-c`. GNU jq would read each line as its own document. That array premise is already pinned by `jq_jsonl_chain` (seq 28), so the streaming case here is pinned with `-c` rather than quietly encoding the pretty-print behavior either way.

### DIFF
--- a/docs/home/bash.mdx
+++ b/docs/home/bash.mdx
@@ -134,9 +134,19 @@ Because evaluation is per document, commands that emit newline-delimited
 JSON (such as [paginated `gws` list calls](/python/resource/gdrive#pagination))
 pipe straight into `jq` with no reshaping.
 
-Output arity follows the program, not the input: a program that collects
-into an array (`[.a[] | .t]`) prints one array, and only a top-level
-iterator (`.a[]`) prints one line per element.
+Output arity follows the program, not the input. A jq program emits a
+stream of values and each one prints on its own line, so `.a[]`, `.a, .b`
+and `range(3)` all print several lines, while a program that collects
+into an array (`[.a[] | .t]`) emits one value and prints one line.
+
+```bash
+jq -c '.name, .age' /data/user.json
+"alice"
+30
+
+jq -c '[.name, .age]' /data/user.json
+["alice",30]
+```
 
 ## Supported bash syntax
 
--- a/integ/unix/jq/comma.json
+++ b/integ/unix/jq/comma.json
@@ -1,0 +1,446 @@
+{
+  "cases": [
+    {
+      "id": "jq_comma_two_outputs",
+      "seq": 700000,
+      "targets": [
+        "ram",
+        "disk",
+        "redis",
+        "opfs",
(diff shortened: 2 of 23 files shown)


zechengz:
@codex review

PR #672, opened 2026-08-01
Split the TypeScript builtin command specs by family

zechengz:
## What

`commands/spec/builtins.ts` held all 93 command specs in one 1,189-line table, while Python splits the same specs across nine family modules. TypeScript now mirrors that layout.

| | Before | After |
|---|---|---|
| `builtins.ts` | 1,189 lines, all 93 specs | 29 lines, façade only |
| `builtin_specs/` | — | 9 family modules + aggregator |

Families match Python one for one: `archive`, `fs_mutate`, `hashing`, `listing`, `net`, `runtime`, `search`, `text_proc`, `viewing`. The aggregator raises on a duplicate name the same way Python's `__init__` does, and `builtins.ts` still exports `specOf` and `BUILTIN_SPECS`, so no caller changes.

The command-to-family assignment was taken by importing Python's modules rather than reassigned by hand. The two key sets already agreed exactly at 93 names, with nothing on either side alone.

## This is a pure move, and that is checked rather than asserted

Regenerating the spec dump from the split code produces **byte-identical JSON** for all 93 commands in both the node and browser outputs. Zero drift.

That is the parity gate from #666 doing the job it was built for, and it is not vacuous: deleting a single spec fails generation with `no builtin spec: wget`.

Preserved deliberately: the null-prototype freeze on `BUILTIN_SPECS`, so a script-controlled name like `toString` still misses instead of resolving an `Object.prototype` member.

## Verification

- spec regeneration: byte-identical, zero drift
- spec parity gate: 93 commands match
- `tsc --noEmit`: clean
- core: 541 files / 5,703 tests
- node: 202 files / 1,853 tests
- browser: 44 files / 181 tests
- pre-commit: exit 0


zechengz:
@codex review

PR #673, opened 2026-08-01
test(slack): assert the size invariant over every user and channel day

zechengz:
Closes the last open risk in the fskit size push-down: whether slack's `users/*.json` size can be trusted.

## The assumption

`readdir` sizes a user file from the **`users.list`** member object, but `read` renders the **`users.info`** response. Same renderer, two endpoints. So `stat().size == len(read())` holds only while those payloads encode identically, key order included, since `json.dumps` preserves insertion order.

That was never checked. It is now: **verified live on 2026-08-01**, calling both endpoints for every real user in a live workspace and comparing through mirage's exact encoder.

```
real users in workspace: 3
  U04JMKD1LJK: IDENTICAL (1677 bytes)
  U04K21SEVR9: IDENTICAL (1721 bytes)
  U04KRSCK28Y: IDENTICAL (1874 bytes)

VERDICT: 3/3 identical, 0 mismatched
```

No live bug. Slack ships the same object from both endpoints.

## What this PR adds

Two integ cases pinning the invariant where integ can pin it:

```sh
n=0; for f in /slack/users/*.json; do
  test "$(stat -c %s "$f")" = "$(cat "$f" | wc -c)" && n=$((n+1))
done; echo "$n match"        # -> "11 match"
```

plus the same shape over `/slack/channels/general__C1/*/chat.jsonl` (90 days).

Better than the pins that existed in three ways:

- **Breadth.** The old pins covered one user and one channel-day. These cover all 11 users and all 90 days, including the empty ones.
- **They state the invariant, not a constant.** `stat` compared to `wc -c` directly, so they do not drift when the fixture changes.
- **They cannot pass vacuously.** Counting matches is deliberate: if the glob stops expanding, the count is 0 rather than silence.

Confirmed they bite, by mutating `size=len(user_json_bytes(u)) + 1` in slack's readdir:

```
FAIL sz_slack_user_stat: expected '178', got '179'                              <- old pin, 1 user
FAIL sz_slack_every_user_size_matches_read: expected '11 match', got '0 match'  <- all 11
```

## What integ cannot cover, now written down

The fake server builds `users.list` and `users.info` from one row through one `userJson()` helper, so they agree by construction. If real Slack ever changes one endpoint and not the other, integ stays green and only a live call catches it. Both languages carried a bare "verified live" comment with no date or scope; they now record the scope and this limitation, so nobody mistakes green CI for validation of the Slack side.

## Verification

pre-commit clean, slack battery green on both hosts (90 cases, ~7s).

### DIFF
--- a/integ/resources/sizes/stat_read.json
+++ b/integ/resources/sizes/stat_read.json
@@ -1275,6 +1275,45 @@
         "stdout": "90 match\n",
         "stderr": ""
       }
+    },
+    {
+      "id": "sz_discord_every_member_size_matches_read",
+      "seq": 700102,
+      "targets": [
+        "discord"
+      ],
+      "command": "n=0; for f in /discord/*/members/*.json; do test \"$(stat -c %s \"$f\")\" = \"$(cat \"$f\" | wc -c)\" && n=$((n+1)); done; echo \"$n match\"",
+      "expect": {
+        "exit": 0,
+        "stdout": "3 match\n",
+        "stderr": ""
+      }
+    },
+    {
+      "id": "sz_discord_every_chat_day_size_matches_read",
+      "seq": 700103,
+      "targets": [
+        "discord"
+      ],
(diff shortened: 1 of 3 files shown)


zechengz:
@codex review

PR #676, opened 2026-08-01
Compile command specs once and unify short/long flags onto one dest

zechengz:
## What

- New `spec/compile.py`: `CommandSpec` lowers once (cached) into the lookup tables the parser walks, instead of being rebuilt on every `parse_command` call. Attached-value candidates are ordered longest-first, so `-name` vs `-n` resolution is deterministic instead of set-iteration-order dependent.
- Dest unification: both spellings of an option land on one canonical key (the long name when an option declares both). `tee -a` and `tee --append` now produce the same flag bag.
- `spec_flag_names` returns canonical names only, so a stale short-name read raises through `FlagView` instead of silently reading False. Every site that raised is fixed in this PR.

## Why

- GNU semantics get more correct for free: `cp --update=all -u` is last-wins with no per-spelling mirror hack, and `sort -k1 --key=2 -k3` accumulates `[1, 2, 3]` in true command-line order (the old two-list scheme reassembled it as `[1, 3, 2]`).
- Deletes the both-name or-chains (`fl.as_bool("a") or fl.as_bool("append")`) from 12 generics, and fixes readers that were silently broken for one spelling.
- First step of the parser work in the CLI design doc; `count=`/`multiple=`/`choices=`/`required=`/`default=` build on the canonical dest and follow in the next PR together with the TS mirror.

## Notes

- Python only; the TS parser keeps its current per-spelling behavior until the mirror PR (which builds on #674's FlagView). Command outputs are unchanged, so the cross-language integ suites are unaffected.
- Real reader fixes beyond the or-chains: the wc formatter helper family (`args_l/w/c/m/L` params renamed to canonical), fanout's head/tail header forcing (head pairs `-q/-v` with longs, tail declares them short-only), and the s3 mkdir wrapper, which read flags through signature params and is the one class of site FlagView cannot surface.

## Testing

- `tests/commands/` fully green; full `tests/` green except pre-existing environmental failures (local redis URL, missing e2b extra) that fail identically on main.
- New `tests/commands/spec/test_compile.py`; unification semantics pinned in `test_parser.py` (cross-spelling last-wins, in-order accumulation, canonical-only allowed-set).

### DIFF
--- a/python/mirage/commands/builtin/general/curl.py
+++ b/python/mirage/commands/builtin/general/curl.py
@@ -64,7 +64,7 @@ async def curl(
     F: str | None = None,
     o: str | None = None,
     L: bool = False,
-    f: bool = False,
+    fail: bool = False,
     s: bool = False,
     S: bool = False,
     dispatch: Callable[..., Any] | None = None,
@@ -107,7 +107,7 @@ async def curl(
             f"{exc.port}: Could not connect to server\n").encode()
         return None, IOResult(exit_code=EXIT_CONNECT, stderr=err)
     # Only -f makes an error status an error, and then nothing is written.
-    if f and resp.is_error:
+    if fail and resp.is_error:
         err = b"" if quiet else (
             f"curl: ({EXIT_HTTP_ERROR}) The requested URL returned error: "
             f"{resp.status}\n").encode()
--- a/python/mirage/commands/builtin/generic/cat.py
+++ b/python/mirage/commands/builtin/generic/cat.py
@@ -18,17 +18,15 @@ class CatFlags:
 
 def parse_flags(flags: Mapping[str, object]) -> CatFlags:
     fl = FlagView(flags, spec=SPECS["cat"])
-    show_all = fl.as_bool("A") or fl.as_bool("show_all")
+    show_all = fl.as_bool("show_all")
     return CatFlags(
-        number_lines=fl.as_bool("n") or fl.as_bool("number"),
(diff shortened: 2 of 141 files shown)


zechengz:
@codex review

zechengz:
Both Codex findings addressed in 9970525e3:

**P1 (wrappers consuming only the short keyword):** fixed the named sites and the rest of the class, in both languages. Python: curl -f (the four red integ scenarios), gridfs mkdir -p/-v, mongodb tail -f, and head -c provisioning (the integ-shared-parity mismatch: the provision helper read 'c' while head now delivers the canonical 'bytes', so python degraded to a range estimate while TS stayed exact). TypeScript: s3/gridfs/opfs mkdir, postgres head pushdown, mongodb tail -f, curl -f. Audited the remaining wrappers mechanically: every short-named signature param either has a canonical fallback or its flag is short-only in that command's spec.

**P2 (TypeScript mirror):** done in the same commit, same shape and file layout: compile.ts (WeakMap-cached CompiledSpec), parseCommand rewritten over it, specFlagNames canonical-only, and the reader sweep including sort.ts, whose per-spelling alias helpers concatenated short-list-then-long-list and lost key order exactly as described; both languages now produce [1, 2, 3] for `sort -k1 --key=2 -k3`.

Verified: python tests/commands green; TS core 5718/5719 (one pre-existing local redis hook timeout), node 2013 green after core rebuild, browser green, tsc clean.

Issue #646, opened 2026-07-26
opfs: migrate to makeGenericCommands instead of hand-registering commands


zechengz:
## Problem

opfs is the only backend of 35 that does not use `makeGenericCommands`. It hand-registers roughly 69 commands across 61 files under `typescript/packages/browser/src/commands/builtin/opfs/`.

Every other backend gets `du`, `ls`, `find`, `wc` and the rest automatically from the factory by declaring a `CommandIO`. opfs declares each command individually.

## Why it matters

Each hand-registered command is a place where opfs can drift from the shared generic. `du` is the concrete example: it needed its own `du.ts` purely because opfs is not on the factory, and before #642 that file carried its own copy of the flag handling.

#642 pointed it at `runDu` so it now shares all semantics, but the wrapper still exists and still wires by hand:

```ts
const out = await runDu(paths, opts, (targets) => Promise.resolve(targets), ...)
```

Note the pass-through `resolveGlob` where factory backends pass a readdir-backed resolver. That one is not currently observable (globs are expanded upstream in `workspace/expand/globs.ts`, and opfs was probed against a factory backend on bare and quoted globs), but it illustrates the shape of the divergence risk.

## What is not blocking this

opfs has every primitive `CommandIO` needs. This is legacy wiring, not a structural limitation. Earlier notes claiming opfs was structurally unable to use the factory were wrong.

## Scope

Browser-only, but wide: all 69 registrations plus their tests. Worth its own PR rather than riding along with a command change.

Files: `typescript/packages/browser/src/commands/builtin/opfs/`

Found while reviewing #642.

zechengz:
Done by #678, which moved opfs onto the factory alongside the du op pairing.

`commands/builtin/opfs/index.ts` on main is now the one-line spread:

```ts
export const OPFS_COMMANDS: readonly RegisteredCommand[] = [
  ...makeGenericCommands<OPFSAccessor>(ResourceName.OPFS, OPFS_IO),
]
```

The directory is down to three files (`index.ts`, `io.ts`, `opfs.test.ts`) from 61. `du.ts` is gone along with the pass-through `resolveGlob` this issue flagged, so opfs now gets `du`, `ls`, `find`, `wc` and the rest from the same factory as the other 34 backends.

Closing. It was fixed but never linked to the PR that did it.

Issue #320, opened 2026-06-15
find: SaaS backends silently ignore -size / -mtime predicates


zechengz:
## Summary

The pure-API `find` backends — **slack, discord, trello, linear, langfuse, email, github_ci** — accept `-size` and `-mtime` (they're in `SPECS["find"]`) but their bespoke wrappers never read them. So the predicate is **silently ignored**: the user asks to filter and gets unfiltered results back.

(`onedrive`, `gmail`, and the generic filesystem backends — ram/disk/redis/s3/ssh/nextcloud — do implement `-size`/`-mtime`, so they are unaffected.)

## Current behavior

| Command | Today |
|---|---|
| `find /slack -size +1k` (valid) | silently ignored → returns **unfiltered** results |
| `find /slack -mtime -7` (valid) | silently ignored → returns **unfiltered** results |
| `find /slack -size abc` (invalid) | ignored → proceeds to the network walk → fails with a **network/auth error**, not a clean message |

The invalid-value row is the most confusing: a typo surfaces as an unrelated API error instead of `find: invalid argument 'abc' to '-size'`.

Context: the `-maxdepth`/`-mindepth` numeric validation was just made uniform across all backends in #316. `-size`/`-mtime` were left out because these backends don't implement those predicates at all.

## Options

- **A. Validate syntax everywhere.** Reject malformed `-size`/`-mtime` values in every wrapper (raise `FindParseError`), even if the backend can't filter. Fixes the confusing-error case; valid-but-unsupported still silently ignored. Low risk.
- **B. Reject as unsupported.** If a backend can't honor `-size`/`-mtime`, error (`find: -size not supported on slack`). Honest, but diverges from GNU find (which supports `-size` everywhere) and could break harmless usage.
- **C. Actually implement `-size`/`-mtime`.** Filter via per-entry stat, like the generic `walk_find` already does. Most GNU-correct; adds network cost and needs per-backend semantics for "size"/"mtime" on messages/cards/etc.

## Recommendation

Ship **A** as a small follow-up (uniform invalid-arg rejection, no GNU divergence). Decide **B vs C** deliberately per backend — some (e.g. `-mtime` on message backends) are meaningful enough to warrant C; for others `-size` may never be meaningful and B is fine.

Must keep Python and TypeScript in parity, and extend `integ/find_arg_errors.{py,ts}` to cover `-size`/`-mtime` once decided.

zechengz:
[...]

PR #678, opened 2026-08-02
Pair the du ops, move opfs to the command factory, cover find predicates in integ

### LINKED ISSUE #645: du: du_size without du_entries silently drops directory rows
## Problem

`CommandIO` exposes `du_size` and `du_entries` (`duSize`/`duEntries`) as two independent optional fields. They are not actually independent.

If a backend sets `du_size` but not `du_entries`, the builder leaves `compute_entries` as `None`:

```python
compute_entries = (partial(_op_entries, entries_op, accessor, index)
                   if entries_op is not None else None)
```

and the generic then degrades to printing only the operand line:

```python
if flags.s or compute_entries is None:
    ...  # total only, no directory rows, no -a
```

That is exactly the pre-#642 behavior: no per-directory rows, `-a` inert, `--max-depth` nearly meaningless.

## Why it matters

Nothing enforces the pairing and no test fails. A new backend that implements only the cheap half (a subtree total is usually easier to get than a full leaf listing) quietly gets non-GNU `du` output, and the failure is silent rather than loud.

Today all 11 Python and 13 TypeScript native-du backends set both, so the path is unreachable. It is a trap for the next backend, not a live bug.

## Suggested fix

Make the two a single optional pair rather than two fields, so "native du" is all-or-nothing at the type level and the degraded branch stops being reachable by accident. Mirror in both languages.

Files:
- `python/mirage/commands/builtin/generic_bind/adapter.py`
- `python/mirage/commands/builtin/generic_bind/builders/du.py`
- `python/mirage/commands/builtin/generic/du.py`
- the TypeScript mirrors under `typescript/packages/core/src/commands/builtin/`

Found while reviewing #642.

zechengz:
Closes #645, #646. Adds the integ coverage #320 asked for.

## #645: du_size + du_entries become one pair

`CommandIO` exposed the two halves of a native `du` as independent optional fields. They are not independent: the generic derives its per-directory rows from `entries`, so a backend wiring only the cheaper `size` silently got operand totals with no directory rows and an inert `-a`.

They are now a single value, `DuOps` (python) / `DuOps<A>` (typescript), with both halves required. Half a native du is unconstructable, which lets `compute_entries` go from optional to required through `run_du` / `runDu`, so the degraded branch is deleted rather than documented:

```python
if flags.s or compute_entries is None:   ->   if flags.s:
```

Updated 11 python and 13 typescript backends. The `resource/*` op tables keep their own `du_size` / `du_entries` op names; that is the VFS dispatch layer, not this one.

## #646: opfs joins the factory

opfs was the only backend of 35 not using `makeGenericCommands`, hand-registering 69 command families across 70 files. Several of those wrappers were reimplementations whose own comments said "mirrors the generic builder", which is exactly the drift surface the issue describes.

It now declares an `OPFS_IO: CommandIO<OPFSAccessor>` like every other backend. Net effect on behavior is a strict superset: opfs goes from 69 to 72 command families, gaining `numfmt`, `od` and `truncate`.

## #320: integ proof that find predicates filter

`-size` and `-mtime` are already honored on all seven SaaS backends (python routes through `parse_find_args` + `walk_find`, typescript through `findGeneric`). The malformed-value half was already covered by `argerr/{size,mtime}.json` across 16 targets, but nothing asserted that a valid predicate actually filters.

Nine cases across trello, linear and discord: one precise positive filter each, plus two unsatisfiable filters (`-size +1000000c`, `-mtime +36500`) that must return empty. The unsatisfiable pair is the real regression guard, since a silently ignored predicate returns the whole tree rather than nothing.

## Verification

Integ, every backend with a native du, python host unless noted:

| Target | Cases |
|---|---|
| ram, disk, redis | 5950 (also 5950 on the typescript host) |
| opfs (typescript-browser) | 1939 |
| s3, s3-prefix | 3824 |
| onedrive | 1927 |
| sharepoint | 1921 |
| sharepoint-prefix, dropbox-root, databricks-prefix | 5380 |
| dropbox | 1914 |
| ssh | 1905 |
| gdrive | 1824 |
| gdrive-folder, hf-prefix | 3359 |
| databricks | 1743 |
| box | 1700 |
| hf | 1580 |
| gridfs, gridfs-prefix | 3434 |
| trello, linear, discord | 210 (also 210 on the typescript host, byte-identical) |

All 0 failures. Nextcloud was not run locally and is left to CI.

Unit: full python suite green; typescript core 5710, browser 181, node 1990. `tsc --noEmit` and `pnpm -r build` clean. `pre-commit run --all-files` exits 0 on two consecutive runs.

The #645 invariant itself is type-level, so it is unit-tested rather than integ-tested: python asserts a one-sided `DuOps` raises `TypeError`, and on the typescript side `tsc` rejects the old `duGeneric(..., undefined)` call. Integ is what proves the collapse changed no du output.


zechengz:
@codex review

PR #686, opened 2026-08-02
refactor(executor): split command.py and command.ts into a command package

zechengz:
Splits the two largest executor files (Python `command.py` 1,029 lines, TS `command.ts` 1,101 lines) into per-concern modules, mirrored across languages: `flags`, `guard`, `routing`, `functions`, `run`, `types`, with a same-named facade. Behavior-preserving except for proven defects.

Fixes shipped with proof:
- Deleted TS `tryUnmountIntercept` and its whole `unmount` dependency chain. It was unreachable: the mount-root guard fires first for every rm/rmdir on a mount root (verified by mutation, the suite stays green with the action replaced by a throw, and statically). A new test pins that `rm -rf` on a mount root refuses with EBUSY, never unmounts.
- Aligned parse-time PathSpec synthesis in both languages: `resource_path` stays empty, the mount stamps the backend key at execute time. The parse-time stamp was dead (full suites in both languages pass with the field set to a sentinel). The three previously disagreeing inline constructions are now one `synthesize_path_spec` / `synthesizePathSpec` helper, and `parse_flags` drops its unused `mount_prefix` parameter.
- Typed the executor seams: `DispatchFn` and `ExecuteNodeFn` protocols replace `Callable[..., object]`, `scalar_find_flags` no longer mixes `object`/`Any`.

Cleanups: job builtin dispatch tables in both languages, Python's shell-function block extracted as `run_shell_function` (matching TS), shared `-p/--parents` detection helpers, dead `_FIND_ACTION_FLAGS` removed, TS `ReturnSignal` moved to `control.ts` matching Python's `control.py`.

Tests:
- New unit tests for guard, flags, and routing in both languages.
- New `integ/unix/guard/root.json` pins mount-root refusals end to end (rmdir/mv EBUSY, mkdir EEXIST vs `mkdir -p` no-op, touch EISDIR, `ln -s` EEXIST). Both harnesses pass 4,077/4,077 on ram and disk with byte-identical expectations.


zechengz:
@codex review

PR #704, opened 2026-08-04
feat(policy): pre_execute hook absorbs the runtime routing policy

zechengz:
The routing script (`policy=`) was still a separate system in `mirage/runtime/policy/` with its own `PolicyError`. This PR makes it the fifth hook on the policy chain, so one `Policies` chain now owns every decision point: `pre_command -> pre_execute -> pre_ops -> post_ops -> post_execute`.

- `pre_execute` hook: fires parse-before-dispatch once per typed line. `Deny` refuses it (exit 126, existing `policy denied` wording); `Route` places it on a named runtime like the `runtime=` argument. First Route wins; a later Deny still beats it; `PolicyError` propagates instead of failing closed.
- `Route` joins the closed Action vocabulary (`VALIDITY: pre_execute = {deny, route}`). `RouteResult`/`DenyResult` are deleted: code policies return `Route`/`Deny`/a runtime name/the wire dict/`None`.
- The routing script becomes the `RoutingPolicy` built-in, registered on the chain by `Workspace(policy=)`; `decide_line` shrinks to entry-script willingness only.
- `PolicyContext` -> `ExecuteContext`, moved into `mirage/policy/types.py` beside `ParsedCommand` (typed against a narrow `RuntimeIdentity` protocol; wire schema unchanged). The two `PolicyError` classes collapse into `mirage/policy/errors.py`, restoring the single TS `PolicyError` export.
- Package rename: `mirage/runtime/policy/` -> `mirage/runtime/route/` (TS `executor/policy/` -> `executor/route/`), so "policy" now unambiguously means the policy package.
- Gate helpers split into `policy/gates.py` / `gates.ts` with a real `Policies` import.
- Integ: `deny_line` + `route_command` runner policies and `hook_pre_execute_deny`/`hook_pre_execute_route` cases in both languages; all pre-existing routing cases pass byte-identical through the new path.

Verified: python full suite + mypy clean; TS core/node/server suites green; integ policy suite 26 py / 27 ts; cli.sh 96 passed 0 failed; pre-commit x3 clean.


zechengz:
@codex review

zechengz:
> P2: Fail closed consistently on evaluator infrastructure errors

Real, and fixed in d38a771f3.

Confirmed the divergence: `runWithTimeout` only throws `CommandTimeoutError`, so the blanket `catch` arm in TS `evalSource` was converting every unexpected evaluator failure into `PolicyError`, which `Policies.fire` rethrows and `execute()` surfaces as a rejection. Python's `eval_source` only converts `asyncio.TimeoutError` and `EvalError`, so the same failure reached `_fire`, failed closed into a Deny, and came back as exit 126.

Normalized on the Python side, because it matches the documented split: `PolicyError` is for caller-fixable mistakes (no evaluator in the world, a script that hangs, a script that does not parse or raises) and those three arms are unchanged; an evaluator whose own machinery breaks (dead container, dropped connection) is not caller-fixable, so it now propagates unchanged and the policy chain fails closed on it.

Pinned in both languages with a `BrokenEvaluator` whose `eval()` throws, asserting exit 126 and the `RoutingPolicy failed` wording: `a broken evaluator fails the line closed` in `runtime_policy.test.ts` and `test_a_broken_evaluator_fails_the_line_closed` in `test_runtime_binding.py`. Verified the TS test fails against the old blanket wrap before restoring the fix.

Gates: TS core 6064 green, python policy/route/workspace suites green, integ policy 26 py / 27 ts, pre-commit clean.

Issue #609, opened 2026-07-22
Coreutils gaps (medium/hard, priority-tiered): real-semantics commands, symlinks, follow, metadata flags (part of #597)


zechengz:
Part of #597. The **medium + hard** GNU Coreutils slice: commands and flags with real semantics — symlink resolution, follow loops, provider-quota truth, background job tables, and flags that depend on real VFS metadata (mtime, inode, ownership, dereference).

Reorganized by **priority for agent workloads**, difficulty tag on each (`med` / `hard`). Priority = how often real scripts / agents hit it; difficulty = implementation cost at Python/TS parity. Partitioned **by command** (a command's whole flag set lives in one issue) so this and #610 never touch the same command. Check a box only when Python and TS match GNU coreutils 9.x and focused unit + integ tests cover success, error shape, exit status, and VFS semantics. Exhaustive per-flag lists per command are in #597.

---

## Tier 1 — high priority

- [ ] `ls` full flag set (`hard`, large) — `-i` inode, `-L` dereference, `-S/-t/-c/-u/-X/-v` sort orders, `--time-style`, `-R` recursive, `-g/-o/-n` owner columns, `--group-directories-first`, `-Z` context, `--block-size`, `--hyperlink` … (needs real metadata)
- [x] `stat -c` — support all GNU directives (`med`) — done (#626): full GNU directive set + agent-aware ownership overlay wired into stat.
- [ ] `cp` semantics flags (`med`) — `--preserve/-p`, `--update/-u`, `--no-clobber/-n`, `--dereference/-L`/`-H`/`-P`, `-t`/`-T`, `--backup/-b` — **partial** (#629): `--update/-u`, `--backup/-b`+`-S`, `-t`/`-T`, long spellings, `-f`/`-i`/`--strip-trailing-slashes` no-ops, GNU overwrite-type + arity errors done (py+ts, GNU 9.7-pinned, 26 integ cases). `--preserve/-p` + `-L`/`-H`/`-P` dereference deferred (need metadata-preservation + symlink-follow passes).
- [x] `mv` semantics flags (`med`) — `--update/-u`, `--no-clobber/-n`, `--backup/-b`, `-t`/`-T`, `--exchange`, `--strip-trailing-slashes` — done (#629): all listed spellings, plus `--no-copy` (cross-mount refusal); shares cp's transfer/backup engine.
[...]

PR #722, opened 2026-08-06
feat(du): Honor -S/--separate-dirs (py + ts)

ThomasHartDev:
Related to #609 (`du` flags checkbox): implements GNU `du -S` / `--separate-dirs`.

Directory lines now count only files that sit directly in that directory. Subdirectory sizes stay out of the parent total, including under `-s` and `-a`. Ancestors that only contain other directories still print (at size 0), matching coreutils 9.7.

Changes land in the shared `du` rollup on both runtimes, the command specs, and the Python generic/github wrappers. Unit tests cover rollup and end-to-end output; `integ/unix/du/separate_dirs.json` covers shell-visible short/long/`-Ss`/`-Sa` cases.

Verified with `bash /root/command-center/scripts/oss-playbooks/mirage-verify.sh` (exit 0), plus targeted `pytest tests/commands/builtin/generic/test_du.py` and vitest on `packages/core` `du.test.ts`.


zechengz:
@codex review

zechengz (approved):
Thanks :)

Issue #292, opened 2026-06-12
Migrate remaining generics from named-parameter flags to the bag + parse-once struct contract


zechengz:
## Current state
Only the grep family (generic grep/rg/zgrep) uses the flags-bag contract: wrappers forward `**flags` wholesale and the generic interprets them via a frozen `parse_flags` struct (PR #291), so adding a flag touches only the spec and the generic.

Every other generic (du, cut, uniq, head, tail, wc, ls, find, sed, ...) still receives flags as named keyword parameters in its signature, e.g. `du(..., s=False, a=False, h=False, max_depth=None, c=False)`. That style is typo-safe for free (TypeError/NameError) but couples every backend wrapper to the flag list: adding a flag means editing N wrapper signatures, the exact N-wrapper sweep CLAUDE.md forbids for the bag contract.

## Target
When a family is next touched, migrate it to the grep-family shape:
- wrappers: drop per-flag parameters, declare only injected params (stdin/index/prefix) plus `**flags`, forward wholesale
- generic: `(paths, texts, flags)` signature; `FlagView(flags, spec=SPECS[cmd])` + `@dataclass(frozen=True, slots=True)` struct built by a module-level `parse_flags()`
- mirror the TS generic's parseFlags field inventory where one exists

## Notes
- No urgency and no behavior change expected; do it family-by-family alongside other work, verified by the integ truths.
- Repeatable/list-valued flags and shared helpers benefit most; trivial one-flag commands can stay named-parameter if the wrapper count is low.

zechengz:
Migrated. Every family named in the issue now has a module-level `parse_flags` plus a `@dataclass(frozen=True, slots=True)` struct, built from a spec-bound `FlagView`: du, cut, uniq, head, tail, wc, ls, find, sed, sort, nl.

The contract is also now written into CLAUDE.md as the standing rule ("Generics parse flags once into a frozen struct"), with meta-tests pinning it: `tests/commands/test_no_dead_flag_params.py`, `test_no_raw_flag_reads.py`, `test_flag_query_names.py`.

PR #733, opened 2026-08-08
refactor(flags): delete the dead short-flag params, name the bag's type, spec-bind the TS write-side builders (#292)

bytecii:
Block B items 9 and 10 of the parity plan: delete the flag parameters the parser can never fill, and give the flag bag's value the name TypeScript has always used for it. Item 9 is complete; item 10 lands its core (the eight builders that *are* the implementation) with the rest scoped below.

## 1. The dead short-flag params (T2-2b)

Since #676 both languages compile each spec so that **every spelling of an option maps onto one canonical dest** — the long form whenever an option declares one (click/argparse `dest` semantics). So `mkdir -p` and `mkdir --parents` both arrive as kwarg `parents`, and a builder parameter named `p` can never be filled:

```python
async def mkdir(..., p: bool = False, v: bool = False, m: str | None = None, **flags):
    parents = p or fl.as_bool("parents")   # `p` is permanently False
```

**110 such parameters across 32 builders** are deleted, plus `js -m` and `github du -d`. The `FlagView` read that already sat beside each one is now the only read. This is not a bug fix — a reader "fixing" the short branch today changes nothing — but it is the thing that makes the code lie about what it handles.

Correctly scoped: an option with **no** long spelling keeps its short as the canonical dest, so those parameters are live and stay (`base64 -D`, `mktemp -p/-t`, `join -j`, `du -h/-s/-a/-c/-L`, and the ls/find/sed/tail shorts).

`python/tests/commands/test_no_dead_flag_params.py` is the gate: every registered handler's named parameters must be canonical dests of its own spec, or dispatcher context. It walks the registry, so it covers backend wrappers and provision/aggregate functions too, not just builders.

## 2. Shared provisions read by spelling — and three real bugs

The cost estimators are shared across commands, so they cannot name a dest the way a handler does: `-c` is `bytes` on head and `c` on tail. They took `**kwargs` and reached into the bag raw, which is also how they dodged the existing `test_no_raw_flag_reads` lint (it matches `flags.get(`, not `kwargs.get(`).

They now receive the invoked command's `spec` — the provision call site already had it — and resolve the *spelling* through the compiled dest map:

```python
cap = _flag_of(spec, "-c", flags)             # "the -c option of whichever command this is"
recursive = _walks_a_subtree(spec, flags)     # a *boolean* -r/-R only
```

Three defects fall out:

- **`touch -r REF` was counted as recursion.** `write_metadata_provision` serves ln/mkdir/mktemp/rm/touch and read `r` blindly; rm's `-r` is recursive, touch's `-r` names a reference file. A plain `touch -r` reported UNKNOWN precision instead of EXACT. Only a boolean `-r`/`-R` counts now — the value-typed option under the same letter cannot trip it.
- **`head_tail_provision`'s `n` was dead and never read at all.**
- Its `-c` cap worked only through a raw `kwargs.get("bytes")` fallback.

`test_no_raw_flag_reads` now also catches `kwargs`/`_extra` reads, so that hole is closed.

## 3. `object` is no longer a type (rule + gate)

181 annotations said `object`, which admits bytes where JSON was meant and a `PathSpec` where a flag value was meant — so each use site paid for it with an isinstance chain back to the set the author had in mind. Each recurring set now has a name:

- **`FlagValue`** (`commands/spec/types.py`) — what a command receives. **TypeScript has exported this exact name all along** (`commands/spec/types.ts`); Python spelling it `object` *was* the divergence. It replaces five anonymous copies of the same union (`FlagBag`, three inline dicts in `parser.py`, `ParsedArgs.flags`), so the repo has *fewer* type spellings after this, not more.
- **`ParsedFlagValue`** — the narrower parser stage. This split is load-bearing, not decoration: the parser works on argv and emits only `str | bool | int | list[str]`; the executor promotes PATH values to `PathSpec` in `workspace/executor/command/flags.py`. Collapsing the two into one alias cost 20+ mypy errors.
- **`JsonValue`** (`types.py`) — a decoded payload or API field. Recursive, so it is a forward-reference string until the floor is 3.12; a union with it must be quoted (`Awaitable["JsonValue | X"]`).
- plus `str | PathSpec` for paths, `StatFn` for a stat callable, `SortKey`/`VersionPart`/`DedupKey` for sort keys, and **one-member enums for the two `object()` sentinels** (`_Fd` in redirect, `_NoContent` in gws) so a sentinel can never collide with the real values sharing its variable.

`python/tests/commands/test_no_object_annotations.py` enforces it via AST with an explicit five-entry allowlist: four Python protocol methods whose signatures the language fixes (`__setattr__`, two `__contains__`, `Mapping.pop`), and one guard whose entire job is to catch a value the annotations already claim cannot arrive. CLAUDE.md carries the rule.

**Typing the bag found a latent crash.** `trello`'s shared text-input helper called `read_bytes(config, path)` against a three-argument reader, so any `--*-file` operand would have raised `TypeError`. Fixed with a module-level adapter.

One shape deserves calling out because it looks wrong and isn't: `list[str | PathSpec]`. It has exactly one source — a PATH-typed `pair` option, which accumulates `(name, value)` flattened with only the odd slot promoted. That is `jq --rawfile`/`--slurpfile`; `--arg`/`--argjson` are str pairs and stay `list[str]`.

## 4. TypeScript write-side builders adopt the spec-bound FlagView (T2-3)

TypeScript already had `FlagValue` *and* the canonical-only `specFlagNames` guard that throws on an undeclared name. It simply never used them: the builders read `opts.flags.x` directly, so a name the parser can never emit read as `undefined` with nothing to catch it.

The eight builders that **are** the implementation — no TS generic exists for mkdir/rmdir/touch/ln/rm/truncate — now build `new FlagView(opts.flags, specOf(name))`. Two behavioral tightenings ride along, both toward the Python side: `od` stops treating an explicitly empty `-j`/`-N` as absent (`asStr`, matching `builders/od.py`), and `sed`'s `-e`/`-f` presence test goes through `asList` instead of comparing the raw value against `undefined` and `false`.

A mirror of the Python dead-param scan (over `opts.flags.X` reads, checked against each file's compiled spec) found only **four** dead reads repo-wide — `bc l`, `date I`, `wget O`, `email/rg l` — and all four are the harmless `args_X === true || flags.X === true` residue, so **TypeScript had no live flag bug** where Python had three. `grep.ts`'s old `flags.l` (called out in the plan) is already gone, fixed in #706.

## Deliberately not in this PR

`cp.ts`/`mv.ts` and the ~50 backend wrappers stay on raw reads, so the eslint rule stays scoped to `generic/*.ts` for now — widening it to `commands/builtin/**` fails on every unconverted file, so it has to land last.

`cp.ts` is not a mechanical swap: it declares its own `export type Flags = Record<string, … | PathSpec>` (`generic/cp.ts:52`) because TS's `FlagValue` omits `PathSpec` while the single-mount dispatch path really does pass one. That is the same two-stage split this PR just had to make in Python, and the fix is to mirror it — `ParsedFlagValue` = today's TS `FlagValue`, `FlagValue` = that plus `PathSpec`, widen `CommandOpts.flags` (`commands/config.ts:46`), delete the local type, helpers take a `FlagView`. It ripples through the `CommandOpts` contract and deserves its own change. Verified first that cp/mv read only canonical names today, so nothing is silently broken while it waits (cp's spec declares `-r` and `--recursive` as separate options, so that merge is live, not residue).

## Verification

- pytest green (fuse excluded — no macFUSE on this host); all three lint gates green.
- vitest: core 582 files / 6879 passed, node 207 files / 2104 passed.
- `pnpm -r typecheck` clean; `pre-commit run --all-files` clean including mypy over 1735 source files.
- The four tests that pinned the dead path now exercise the real one: `du`'s `-d` alias is asserted at the parser (where the alias actually lives), the two provision tests pass a `spec`, and `shuf`'s output flag uses the canonical `output`. One test is added for the `touch -r` fix.

Part of #292 (Block B of the parity plan; #706 and #731 closed Block A).


zechengz:
@codex review

PR #735, opened 2026-08-09
refactor(flags): spec-bound FlagView across every command wrapper, and the three Python bugs it surfaced (#292)

bytecii:
Finishes item 10 of the parity plan: the 152 raw `opts.flags.x` reads under `commands/builtin/` go through a spec-bound `FlagView`, cp/mv stop declaring their own flag-bag type, and the eslint rule that made this permanent widens from `generic/*.ts` to all of `commands/builtin/**`. #733 landed the first eight builders; this is the rest.

Converting the reads is mechanical. What it *found* is not: a spec-bound view throws on a name the spec never declares, so every flag the code was asking for and never getting became a test failure instead of a silent `false`.

## 1. What the guard caught in TypeScript

**Four dead reads.** `bc -l`, `date -I`, `wget -O` and `email/rg -l` each read a key the parser cannot emit — those shorts are disambiguated to `args_l`/`args_I`/`args_O` (`AMBIGUOUS_NAMES`). Verified against the parser rather than the spec:

```
wget -O /tmp/x http://u  ->  {"args_O":"/tmp/x"}
bc -l                    ->  {"args_l":true}
date -I                  ->  {"args_I":true}
```

Each of these was the dead half of an `||`, so the shipped behavior was correct — but three test files passed the dead shape (`runWget(..., { O: ... })`), which means the suite was exercising a branch production never reaches. Delete the live half and those tests still pass. They now use the canonical dest.

**`seq -w`'s string branch was unreachable** — the spec declares `-w` boolean, so `typeof flags.w === 'string' ? flags.w : ...` could only ever take the else. Python passes the bool straight through; TypeScript now matches. GNU 9.7 pins agree with both (`seq -w 1 10 5` → `1`, not `01`: the width comes from the produced values, not the operands).

**Provisions now resolve by spelling.** `generic_bind/provision.ts` guessed at `opts.flags.c ?? opts.flags.bytes` because `-c` is `bytes` on head and `c` on tail, and read `r`/`R` blindly for recursion. They now receive the invoked command's spec and ask it, the shape Python got in #733:

```ts
const c = flagOf(opts, '-c')
const recursive = walksASubtree(opts)  // a boolean -r/-R only
```

*(Correcting an earlier draft of this description: this was **not** the TypeScript twin of #733's `touch -r` bug. Python compared truthily and a `PathSpec` reference file read as recursion; TypeScript compared `=== true`, which a path string never satisfies, so `touch -r` was already exact here. The change is robustness — a command spelling it `--recursive` would have slipped past — not a fix.)* `CommandOpts.spec` is new and mirrors Python's `spec=` provision keyword.

## 2. cp/mv, and why the type is *not* widened

`generic/cp.ts` declared `export type Flags = Record<string, … | PathSpec>` with a comment saying `-t` values "arrive as PathSpec on the single-mount dispatch path". They do not. Python's executor promotes PATH-typed flag values back to PathSpec (`workspace/executor/command/flags.py:96-135`); TypeScript's twin does the opposite at `command/flags.ts:91` — `flagKwargs[key] = match.virtual` — and no site in either runtime package writes a PathSpec into a bag.

So TypeScript needs **one** name, not Python's two-stage split: the already-exported `FlagValue`. It replaces 70 inline copies of `string | boolean | number | string[]` across 43 non-test files, plus the private `type Flags` in `cross_mount.ts`, and `CommandOpts.flags` names it. `parseCpFlags`/`parseMvFlags` and their four helpers now take a `FlagView` exactly like `parse_cp_flags` does, and the builders and crossmount relays construct it with `specOf('cp'|'mv')`.

cp's PathSpec-tolerant read stays, with the comment corrected to say which language actually produces that shape. **Porting the promotion to the TypeScript executor is filed as follow-up work** — 23 path-typed options across 18 commands each reconstruct a PathSpec at their read site today, and changing that is a behavior change needing per-command integ coverage, not a rename.

## 3. Three live Python bugs the same rule surfaced

Doing the TypeScript side made the Python violations visible, and each hid a real defect:

- **`--desc_file` / `--text_file` never worked on trello.** A PATH-typed flag arrives as a PathSpec, so `fl.as_str("desc_file")` read it as absent — while the guard beside it used the raw `_extra.get` and *did* see the value. The branch was entered and the file was never read. (See §5: covering this end to end found a second half to the bug.)
- **`gzip -1` compressed at zlib's default.** `extract_level` looked the bag up by the bare digit; `-1` is the one digit the parser disambiguates (`args_1`). `-2`..`-9` worked. The existing native test asserted only `len(-9) <= len(-1)`, which stayed true with the bug.
- **`cp --backup --suffix=` used an empty suffix**, so the backup name equalled the original's. Pinned against GNU 9.7 in `debian:stable-slim`:

  ```
  $ cp --backup --suffix= f g && ls
  f  g  g~
  ```

  TypeScript already had this right (`firstStr` treated `''` as absent), so this one goes the other way — Python is fixed to match, and both sides now pin it.

## 4. The gate that should have caught all of it

`test_no_raw_flag_reads.py` matched only `flags.get(`. The same bag reaches a command as `**kwargs` and `**_extra`, and every offender above hid under one of those names. It now matches all three, with `(?<![\w.])` so `FileStat.extra` stays out, and exempts `config.py` (the `@command` wrapper answers `--help`/`--version` off the bag as spec layer).

*(This corrects #733's PR body, which claimed the widening had already happened. It had not — only the provision's own raw read was fixed.)*

Rounding out the same rule: qdrant/lancedb declare `cwd` instead of fishing it out of `_extra`, and curl's `-o` / unzip's `-d` stop annotating a PathSpec as `str`. CLAUDE.md records the PathSpec-vs-string asymmetry, which is the trap that produced the trello bug.

## 5. Integ coverage for all of the above — and what it found

The fixes above were pinned by unit tests only, so nothing proved them across the 20 backends the battery sweeps.

**Coreutils.** `unix/gzip/1.json` mirrors `9.json` for the one digit that was broken; `cp`/`mv` grow an empty `--suffix=` case pinning the GNU fallback to `~`; `unix/provision/basic.json` grows `prov_touch_r`, which sits between `prov_rm_r` (unknown) and `prov_mkdir` (exact) and would have printed `precision=unknown` before #733. Its op count is 2 — the provision layer scopes every PathSpec in the command line, reference file included — and both hosts agree.

**Trello was the harder gap: no write was covered on either host.** Its fake also ran per-host — the Python runner started `server/trello_server.py` in-process while CI started a second copy of the same file for TypeScript, so neither host could observe what the other wrote. `server/trello.ts` replaces it, built the way `server/slack.ts` is: Prisma + SQLite, seeded from the same fixture, one shared process both hosts call over `TRELLO_ENDPOINT` with `POST /reset` rolling back. Verified by diffing both servers endpoint for endpoint — 27 requests covering every route, the 404 bodies, a full write sequence, an idempotent re-add, a unicode card name, and the reads after those writes — all byte-identical. The datasource env var is now `INTEG_DB_URL` since it backs two services.

With writes finally exercised, two more defects fell out:

- **The other half of the `--desc_file` bug.** Reading the flag works after §3, but the path handed to the backend was still the *virtual* one, and `core/trello/read` keys off the mount-relative path. `/board/workspaces/...` looked for a `board` collection that does not exist, so an in-mount operand died with ENOENT and an off-mount one with "Operation not supported" — the flag had never once landed a read. Python now carries the whole PathSpec through `file_operand` and uses `resource_path`; TypeScript strips the prefix with `mountKey(filePath, opts.mountPrefix)`.
- **The API error text diverged.** TypeScript named the failing endpoint, Python did not. Python now names it too.

## Verification

- eslint clean with the widened rule; **0** raw `opts.flags.` reads left under `commands/builtin/`.
- `pnpm -r typecheck` clean; vitest 586 core / 44 browser / 207 node / 24 server / 17 agents / 8 cli files, all passing.
- pytest green (fuse excluded — no macFUSE on this host); mypy clean over 1744 source files; all four lint gates green.
- Battery on both hosts: `ram` 2295/2295, `disk` 2260/2260, `trello` 90/90 — byte-identical between Python and TypeScript.

## Known gap this surfaced but does not fix

TypeScript cannot honor `gzip -1`..`-9` at all: `utils/compress.ts` uses the Web `CompressionStream('gzip')`, which has no level parameter, so every level is zlib's default. Python's `zlib.compress(level=…)` respects them. The new integ case therefore pins the roundtrip, not the level — a level-observable pin (size, or the XFL header byte) would fail on TypeScript today, and byte-exact gzip output is not stable across zlib implementations anyway. Fixing it needs a leveled deflate in `core`, which browser has no native source for; worth its own issue.

Part of #292 (Block B of the parity plan; #733 landed items 9 and 10's first half).


zechengz:
@codex review

PR #736, opened 2026-08-09
chore(runtime): report a root mount in prefixes, inline the runtime names, split the monty tests

zechengz:
- `RuntimeVFS.prefixes()` no longer drops a mount at `/`, and `mount_of` can now match one (its boundary was `prefix + "/"`, which for the root is `//` and matched nothing). The one place the exclusion genuinely belongs is `WasmVFS._prefixes` (the build directory has to stay visible beside the mounts), documented there.
- Conformance rows in both languages pin that an explicit root mount is served like any other prefix; pyodide is pinned to refuse it with its own warning, since Emscripten already owns `/`.
- The `workspace.ts` sandbox-mounts comment now states what is actually withheld (the history view and the synthetic root anchor), since an explicit root mount was always forwarded.
- The monty osaccess docstring no longer claims the upstream mkdir-over-file bug was reported; it reproduces on a bare `OSAccess`, so it belongs upstream once filed.
- Deleted both `interface.ts` files, inlining the runtime name as a `readonly name` field to match the Python class attribute.
- Split `tests/runtime/python/test_monty.py` into a `monty/` package mirroring the source split (`runtime`/`osaccess`/`vfs`), and added negative-cache coverage for `MontyVFS` that the TS side already had.


zechengz:
@codex review

PR #737, opened 2026-08-09
refactor(integ,provisions): make the integ runner fail on zero work, and single-source the tables that existed twice (#292)

bytecii:
Items 11, 12 and 14 of the parity plan, plus the backend half of item 11 that the history work exposed. The through-line is the same one #733/#735 had: a table that exists twice drifts, and a gate that can't fail doesn't gate.

## 1. The integ runner could report success having tested nothing (item 14)

Both runners already carry a floor guard, but it is all-or-nothing and facet-only — it fires only when *every* target skips. Measured on the two-target `project` facet with `TRELLO_ENDPOINT` unset:

```
161 passed, 0 failed          exit 0     # healthy
skip [trello]: TRELLO_ENDPOINT not set
 71 passed, 0 failed          exit 0     # 90 cases gone, still green
```

Linear still ran, so `ran != 0` and the guard stayed quiet. Those 90 cases include every trello write case #735 added.

**`--strict` closes it**, driven by a new top-level `services` table in `targets.json` that replaces both hand-rolled service→env if-chains (13 python / 21 typescript, already drifted):

```json
"s3":     { "python": [], "typescript": ["S3_ENDPOINT"] },
"trello": { "python": ["TRELLO_ENDPOINT"], "typescript": ["TRELLO_ENDPOINT"] }
```

An empty list means that host self-hosts a fake — python does for s3, ssh, hf, box, databricks, discord, linear and dify; typescript for none of them. That asymmetry was previously implicit in two diverging if-chains and is now stated once. Loading rejects a service with no entry, an entry naming no target, and an entry missing a host, so adding a backend forces declaring its env. CI passes `--strict` on all 16 invocations; local runs stay permissive.

**Two more silent-loss vectors**, both now rejected at load: duplicate case ids and `targets` naming an unknown target id. The first found a live bug — `unix/lookup/type.json` and `bash/type/{basic,t,error}.json` shared six ids, and `parity.py` keys rows by `(target, id)`, so one of each pair was being **silently dropped from the py↔ts diff**. The `unix/lookup` set is renamed `lookup_type_*`. `parity.py` also gains a zero-pairs floor, since a run that compared nothing agrees with itself trivially.

**`check` is now additive.** Both harnesses *overwrote* stdout with the stat line, so a case could never pin both what a command printed and what it left behind — which is precisely why `chmod -v`, `chown -v/-f/-h` and `chgrp -v/-f/-h` have zero integ coverage. It now lands in `expect.check`; the 15 existing cases are migrated, and `touch_d_then_mv_v` pins both at once via `mv -v`. (`chmod -v` still accepts `-v` and prints nothing — a real gap, left for its own change.)

**The gates are themselves tested.** There were no tests for the integ runners at all, which is a poor joke for a change whose deliverable *is* assertions. `integ/runners/tools/gate_selftest.py` asserts 12 properties across both hosts and runs in CI ahead of the battery. Writing it corrected me once: my first typescript assertion failed and the code was right — blanking both vars empties the whole facet there, so the older guard fires first. The tests now use `--target` to isolate the new gate from the old one.

## 2. Two more tables that existed twice (item 11)

**`Limit.aggr`.** Python walks `model_fields` for each field's `Aggr(rule)`; typescript had a hardcoded 4-field literal, so a fifth bound would silently not compose. It now reads a mapped-type table — adding a bound without its rule is a compile error, verified by adding one.

**Backend provisions — the item-11 pattern across every backend.** Python's history commands each named their own estimator, a second copy of the family catalog; `with_default_provisions` (mirroring typescript's `withDefaultProvisions`) replaces that. The same shape turned out to hold for 19 backends in *both* languages, so it is fixed the same way. Of 39 hand-named provisions per language:

- **21 were pure duplication** (s3, gridfs, mongodb, postgres, box, dropbox, dify) and now come from the catalog. Verified command-by-command that all 21 land on the identical estimator family — `21 checked, 0 mismatches`. `mongodb/_provision.py` and `postgres/_provision.py` became unreferenced and are deleted.
- **14 are genuinely bespoke** and stay: the API backends' own estimators (`github/du_provision`, `email/find_provision`, `grep_provision` on discord/slack/gmail/langfuse), `general`'s `pure_provision`, and `search` on lancedb/qdrant/mem0, which has no coreutils family.

**Three commands gained a provision** — they had none while their generic siblings on the same backend did:

| | before | after |
|---|---|---|
| `mongodb/rg` | `ops=0 unknown` | `ops=1 unknown` |
| `postgres/rg` | `ops=0 unknown` | `ops=1 unknown` |
| `dify/find` | `ops=0 unknown` | `ops=2 exact` |

Each is pinned by a new case (`mdb_prov_rg`, `pg_prov_rg`, `df_prov_find`), measured against real mongo and postgres rather than predicted.

*Correction to the plan:* it claims `history/cat.py` drops a `resolve_glob` that `default_provision` supplies, flooring globs to UNKNOWN. It does not — the parameter defaults to None, typescript passes only `histStat` too, and neither history implementation has a glob resolver. There was no behavioral difference; this change is anti-drift only.

The plan's third item-11 bullet (`loadOptionalPeer`) it marks refuted and low-value; skipped.

## 3. Dead code (item 12)

`cut`'s `**legacy_flags` — the only back-compat shim in the tree, against CLAUDE.md — and its four test call sites; `cp`'s `find_type` knob threaded through 7 sites and only ever `"f"` (typescript already hardcodes it); `grep`'s unreachable `elif f.files_only` arm (the early return at `:130` covers every operand shape, so the argument is now literal `False` with a comment); `OpRecord.rel_path` (zero readers) and `mount_prefix` (read only by `rel_path` and two test assertions, never serialized — typescript's `OpRecord` carries neither and folds the prefix into `path`); the two byte-identical private `chainStreams` in `generic/{cat,cut}.ts`, folded into the exported `asyncChain`.

`typescript/vitest.config.ts` is deleted: nothing loads it (`pnpm -r --filter './packages/*' test` runs vitest per package cwd, and there is no `typescript/src` for its `include` to match), and it carried `passWithNoTests: true` — the same green-on-zero-work hazard as §1. Its `--experimental-wasm-jspi` execArgv is therefore already not reaching core today; if it is needed it belongs in `packages/core/vitest.config.ts`.

The dead `FILE_MIME_MAP` parquet/orc/feather/hdf5 entries the plan lists were already removed by the filetype unification in #731.

## Verification

- **7906 passed, 0 failed** across s3 + gridfs + box + dropbox — the regression net for the 21 unchanged provisions.
- dify 72/72, mongodb + postgres 94/94, ram/disk 4569 each and byte-identical py↔ts, ram-history 40/40.
- `gate_selftest.py`: 12/12 on both hosts.
- pytest green (fuse excluded — no macFUSE on this host); core vitest 6963. Two pyodide tests failed only when the suite ran concurrently with pytest and the battery; 10/10 in isolation on both this branch and upstream.
- `pnpm -r typecheck`, `pnpm -r build`, eslint, pre-commit all clean.

Part of #292 (Block B items 11, 12, 14 of the parity plan).


zechengz:
@codex review

bytecii:
Both findings were right and are fixed in `7f688e93a`.

**P1 — database battery without `--strict`.** Correct, and a real miss: the sweep that added the flag matched single-line invocations, and `integ-database`'s two commands use a folded YAML scalar (`run: >-`), so they slipped through. Both now pass it. I re-scanned with a matcher that handles multi-line invocations — all 18 runner calls in the workflow now carry `--strict` except the two `--emit` ones, which `parity.py` gates with its own zero-pairs floor.

**P2 — the self-test's TypeScript half never ran.** This one is the sharper catch, because the step was reporting success having tested nothing — precisely the failure mode the file exists to prevent. Two changes:

- `--require-ts` turns an absent `tsx` into a failure instead of a skip. Verified both ways: with tsx present the run is 14/14 and exits 0; with tsx removed it prints `FAIL typescript gates ran: --require-ts given but .../tsx is missing` and exits 1. Without the flag it still skips, so a bare-checkout local run stays usable.
- The step moved from `integ` (python only) to `integ-shared-py`, which uses `integ-battery-setup` and therefore has the venv, node 24 and `integ/node_modules` — so both hosts are actually exercised.

The AGENTS.md pointer was apt: a gate that can't fail isn't a gate, which is the whole subject of this PR.

PR #741, opened 2026-08-09
fix(tar): drop the GPL bzip2 dependency, and parse tar with modern-tar

zechengz:
Two changes to the TypeScript archive path. Python is untouched: its stdlib `tarfile` / `bz2` already do more than either package here.

## 1. The GPL dependency

`compressjs` is **GPLv2-or-later** and sat in `dependencies` of `@struktoai/mirage-node`, which ships Apache-2.0. Anyone installing that package pulled GPL code into their tree.

It is replaced by `seek-bzip` (MIT). seek-bzip only decodes, so `CompressionCodec.compress` is now optional and a codec may be decompress-only. `tar -cj` exits 1 with `tar: bzip2 not supported`, which is the answer browser core already gave for an unregistered codec. Reading `.tar.bz2` still works.

Every permissively licensed JS bzip2 *compressor* is either GPL (`compressjs`, `archive-wasm`) or unmaintained 2022 source that mutates `globalThis` (`bzip2-wasm`), so this is a deliberate py/ts divergence: Python reads and writes `.tar.bz2`, TypeScript only reads one. It is written down in CLAUDE.md along with the exit condition, since adding `compress` to that one codec closes the gap with no other change.

`pnpm licenses list --prod` now reports **no GPL/AGPL/CDDL/EPL/MPL anywhere in the production tree**.

## 2. The hand-rolled tar

`tar_helper.ts` wrote and parsed ustar blocks by hand. Two real bugs, both confirmed against an archive written by Python's `tarfile` before changing anything:

- **Names past 100 bytes were silently truncated.** `deep/<120 chars>/b.txt` lost `/b.txt` entirely, so the member extracted to the wrong path. ustar needs the `prefix` field or a PAX header for those.
- **A PAX extension block was read as a member.** `tar -t` on any archive GNU or Python wrote listed a phantom `././@PaxHeader` row.

It now delegates to `modern-tar` (MIT, zero dependency, web streams, USTAR + PAX), kept behind the same `TarEntry` shape the two call sites already speak, so the dependency is named in exactly one file. `writeTar` / `readTar` became async.

## Tests

New `tar_helper.test.ts` covers the long name, the absent phantom member, and directory / symlink members.

A flag-coverage audit turned up two gaps that are now closed:

- **`-h` (dereference) had no test at all.** A link member carries the same *name* whether tar stored the link or followed it, so the existing symlink test could not tell the two apart. The new tests read the archive back and assert the member kind in both directions, plus the unreachable-target case.
- **The xz test created and listed**, which only proves headers survived. It now extracts too, checking decompressed bodies byte for byte.

`native_tar.test.ts` also swaps the bzip2 create+list test for list and extract against a fixture another tar wrote, plus the create refusal.

The other half of `tar_helper.ts` is `workspace/snapshot/tar_io.ts`, a cross-language format. `integ/cross.sh` (the "Cross-language snapshot interop" job) covers that directly: it snapshots a 4-mount workspace with one language's CLI, loads the tar with the other's, and requires the re-run command output to match byte for byte, in both directions. It passes on this branch.

- core 7012 pass, node 2282 pass, browser 182 pass, 0 fail
- `tsc --noEmit` clean on all five packages
- eslint, prettier, knip, and `pre-commit run --all-files` clean

## Known gap, deliberately not closed here

**Compression x `--strip-components` / `--exclude` combinations** are only covered for `-z`. These are orthogonal by construction: compression wraps the finished tar blob, while strip and exclude act on the plan before any bytes are written.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -429,6 +429,28 @@ Invoke the venv's `pre-commit` binary directly (not via `uv --directory python r
   `-x a -x b` says the same thing), and the `adding:` line carries no
   `(deflated N%)` suffix, since the ratio depends on the compressor and
   would differ between the two languages.
+- **Tar formats come from a library, not from hand-rolled block code, and
+  bzip2 is read-only in TypeScript.** Python builds archives with stdlib
+  `tarfile`; TypeScript uses `modern-tar` behind `tar_helper.ts`, which
+  keeps the `TarEntry` shape (`name`/`data`/`isFile`/`isDir`/`linkname`)
+  the two call sites already speak and is the only place the dependency
+  is named. Do not go back to writing ustar blocks by hand: the version
+  that did truncated any name past 100 bytes instead of using the ustar
+  `prefix` field or a PAX header (so a deep member extracted to the wrong
+  path), and read a PAX/GNU extension block as if it were a member (so
+  `tar -t` on any archive GNU or Python wrote listed a phantom
+  `././@PaxHeader` row). `writeTar`/`readTar` are async for this reason.
+  Compression is a registry (`registerCompressionCodec`): gzip is built
+  in via `CompressionStream`, and a codec may be **decompress-only**,
+  which is the one deliberate py/ts divergence here. Python reads and
+  writes `.tar.bz2` because `bz2` is stdlib; TypeScript only reads one,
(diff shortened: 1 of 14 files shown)


zechengz:
@codex review

PR #743, opened 2026-08-10
feat(python3): parse CPython's option table, and seed sys.path on pyodide

zechengz:
## Summary

- **`python3` had no option table.** Its spec carried one option (`-c`) plus a free-text rest, so every other switch fell through to the operand slot: `python3 -u s.py` ran a file called `/-u`, `python3 -zz -c 'x'` exited 0 with the flag swallowed, and `-V`/`-h` were read as paths.
- **The boundary is argparse's `nargs=REMAINDER`**, declared as `Operand.remainder`. An unknown dash word before the script is python3's usage error; after it, it is data the script must receive. One leniency switch cannot do both, and this is POSIX's own option order (GNU's permuting default is the extension, as `POSIXLY_CORRECT=1 ls a -1` shows). `CommandSpec` and `Option` are unchanged.
- **An option whose argument is a program is not in the grammar.** argparse cannot express it (`add_argument('-c')` beside `nargs=REMAINDER` answers `unrecognized arguments: -u`), and CPython parses its own command line in C for that reason. POSIX already spells the handoff `--`, so a command-name table in `workspace/route` inserts one after `-c`/`-m`'s value, beside the existing line rules. It is added unconditionally, because CPython passes a typed `--` through as data.
- **`-m`** runs through runpy, with a `find_spec` probe so a missing module is CPython's one line rather than a traceback.
- **argv[0]** now matches CPython for all four source doors. The local and wasi tiers hand CPython the program via `-c`, which hardcodes argv[0] to `-c` and names every frame `<string>`; `bootstrap()` re-compiles under the real name, fixing argv[0] and the traceback filename together.
- **Init switches** (`-O -B -W -X -E -I`) ride in `RunArgs.flags`, honored where the engine can and reported on stderr where it cannot.
- **Pyodide `sysPath`** glob-expands and prepends mount paths once the mounts are in place, so a vendored wheel imports without the `sys.path.append` incantation. A glob matching nothing is reported on the run's stderr (the seed runs inside `syncMounts`, before stderr is captured, so it cannot report it itself). Plus `packages` / `packageBaseUrl` / `lockFileURL` for a prebuilt distribution.
- **Separate commit — `fix(ids)`:** `uuid6.uuid7` ordered ids inside a millisecond by bumping the timestamp past the previous one, and the borrowed value became the next call's floor, so a burst of 5000 ids stamped 4991 ms into the future and stayed there. Replaced with RFC 9562's counter in `rand_a`; `uuid6` dropped. TypeScript was already correct.

## Test plan

- Differential harness against vanilla CPython 3.13.7: **14/14** lines match on stdout, stderr and exit code, plus the three `--`-handoff lines checked separately.
- Pyodide exercised live with a real wheel on a RAM mount: import through the glob, prepend order, argv[0], `-O`/`-B`/`-X`, re-seed on later runs, and the zero-expansion warning.
- Python suite exit 0; TS core 591 files / 7171 tests; spec parity 93/93; `git diff --exit-code spec/` clean.

## Notes for review

- `CLISpec` **is** a `CommandSpec` in both languages, so the CLI tier is not a way to escape the shared grammar. CLAUDE.md now states the field rule as a literal test (write the line in argparse and run it), with both python3 halves as the worked example.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -254,14 +254,22 @@ Both dataclasses are shared by every command in the repo, so a field
 added for one command is a field every other command's author has to
 read past, and a fourth one turns the grammar into a pile of per-command
 dialects. **Do not add a field to `CommandSpec`, `Operand` or `Option`
-unless POSIX or argparse already has the concept, and then name it after
-theirs, not after the mechanism it trips inside the parser.**
+unless POSIX *and* argparse both already have the concept, and then name
+it after theirs, not after the mechanism it trips inside the parser.**
 
-The test is literal, not a judgement call: write the equivalent line in
-argparse and run it. If argparse parses it, the concept is borrowed and
-the field is allowed. If argparse cannot express it, the behavior belongs
-to the one command that needs it and must be handled in that command,
-not in the shared grammar.
+Both, not either. The test is literal, not a judgement call: write the
+equivalent line in argparse and run it. If argparse parses it, the
+concept is borrowed and the field is allowed. If argparse cannot express
+it, the behavior belongs to the one command that needs it and must be
+handled in that command, not in the shared grammar.
+
+POSIX alone is not enough, and python3 is exactly why: POSIX specifies
+an option whose argument is a program (`sh -c command_string [command_name [argument...]]`, and python3's own synopsis
+`python [option] ... [-c cmd | -m mod | file | -] [arg] ...`), so an
+either-or rule would license the `Option` field this section exists to
+refuse. argparse is the narrower gate because it is the parser this
+spec layer is modelled on, so a concept it cannot express is one the
+grammar has no shape for.
 
 Both halves of python3's command line are the worked example, and they
 come out on opposite sides:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -248,6 +248,43 @@ never lists it. Two mechanisms follow from that, and they are separate.
   refusing one that happens to spell a mount root would deny an
   ordinary listing.
 
+## `CommandSpec` and `Operand` are not a scratchpad
+
+Both dataclasses are shared by every command in the repo, so a field
+added for one command is a field every other command's author has to
+read past, and a fourth one turns the grammar into a pile of per-command
+dialects. **Do not add a field to `CommandSpec`, `Operand` or `Option`
+unless POSIX or argparse already has the concept, and then name it after
+theirs, not after the mechanism it trips inside the parser.**
+
+The test is literal, not a judgement call: write the equivalent line in
+argparse and run it. If argparse parses it, the concept is borrowed and
+the field is allowed. If argparse cannot express it, the behavior belongs
+to the one command that needs it and must be handled in that command,
+not in the shared grammar.
+
+Both halves of python3's command line are the worked example, and they
+come out on opposite sides:
+
+- **Allowed: `Operand.remainder`.** It is `nargs=argparse.REMAINDER`,
+  which is POSIX's own option order (the first operand ends option
+  parsing; GNU's permuting default is the extension, and `POSIXLY_CORRECT=1 ls a -1` shows the difference). argparse spells
+  it on the positional slot, so mirage spells it on `Operand` too, and
+  `CommandSpec` does not change at all.
+- **Not allowed: an option whose argument is a program.** `python3 -c 'code' -u x` must hand `-u` to the code, but
+  `add_argument("-c"); add_argument("rest", nargs=REMAINDER)` answers
+  `unrecognized arguments: -u`. CPython's own command line is parsed in C
+  for exactly this reason. There is no concept to borrow, so this does
+  not become an `Option` field.
+
+A `CLISpec` **is** a `CommandSpec` (python: subclass; TypeScript:
+`extends`), and every level of a CLI tree parses with the ordinary spec
+machinery. Moving a command to the CLI tier therefore does not exempt it
+from this rule, because it still parses with the same `Option` and
+`Operand`. The CLI tier is for a program *tree* (a verb the line selects,
+like `git status` or `ntn api`), not for a program with an unusual
+option grammar.
+
 ## An option that chdirs: `operand_base`
 
 `tar -C` is not a flag the command reads once, it is a chdir for the path
--- a/integ/runtime/policy.json
+++ b/integ/runtime/policy.json
@@ -20,7 +20,7 @@
           "command": "python3 -c \"print(argv[0])\"",
           "expect": {
             "exit": 0,
-            "stdout": "main.py\n"
+            "stdout": "-c\n"
           }
         },
         {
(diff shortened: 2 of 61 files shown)


zechengz:
@codex review

zechengz:
All six resolved in 74e1620. Each was checked against real CPython (3.12.11 and 3.13.7) rather than taken at face value; all six reproduced.

**`-W` leaks across pyodide runs.** Real, and the worst of the six, since pyodide keeps one interpreter alive across lines. The wrapper now snapshots `warnings.filters` before `_setoption` and restores it plus `_filters_mutated()` in `finally`.

**`-OOO` must clamp.** Real: CPython reports `sys.flags.optimize == 3`, but `compile(optimize=3)` raises `ValueError: compile(): invalid optimize value`. Clamped to 2 so the extra Os saturate instead of failing the run.

**`-E`/`-I` hiding `PYTHON*`.** Real, and wrong in both directions. CPython's `-E` stops those variables *configuring startup*; it does not hide them from the program, which can still read `PYTHONPATH` out of `os.environ` (verified). The deletion is gone.

**`-S`/`-s`/`-E`/`-I` on pyodide.** They were never implemented, and they cannot be: they only change how an interpreter *starts*, and this one started long before the line was typed. Pyodide now reports them through `unhonoredNotice`, which grew a `honored` subset parameter (pyodide honors `B O W X`; monty still honors none). That turned up a latent bug in the same function: the check was `value === 1`, so `-OO` was reported as nothing while `-O` was reported. Now `Number(value) > 0`.

**Missing switches.** Real: `-b`, `-P`, `-x` and `--check-hash-based-pycs` are all accepted by CPython 3.12/3.13 and were being refused as unknown. All four declared. `-b` is a count flag (`-bb` escalates the warning to an error, which round-trips correctly), `-P` and `--check-hash-based-pycs` ride as init flags, and `-x` is answered by the source resolver rather than as an init flag: handing it to an engine that runs code through `-c` would honor it nowhere while claiming otherwise. It empties the first line rather than dropping it, because CPython preserves file line numbering under `-x` (a raise on physical line 2 still reports line 2).

The four excluded switches are now named in the spec comment instead of implied: `-i` (interactive prompt), `-d` (debug-build parser trace), `-v` (import tracing) and the `--help-env`/`--help-xoptions`/`--help-all` dumps, which document a CPython build rather than this command.

One deliberate divergence: CPython parses `--check-hash-based-pycs` by hand and so rejects `--check-hash-based-pycs=never`, which mirage's parser accepts. Left permissive, and noted in the spec.

**Clustered `-uc'code'`.** Real. The scan now walks a short cluster letter by letter instead of prefix-matching, so `-uc 'p' -v` and `-ucp -v` both hand `-v` to the program. Fixing it surfaced a second hole in the same scan: only *short* value spellings were stepped over, so `--check-hash-based-pycs never -c p -u` would have let `-u` be parsed as python3's own.

**Verification.** The vanilla differential grew to 25 cases and reports 24/25 MATCH against CPython 3.13.7 with `MIRAGE_LOCAL_HOME` pinned to the same interpreter. Every new switch matches exactly, including `-bb` escalating to an error and `-x` skipping the shebang line. The one DIFF is the already-documented extra bootstrap frame in file-mode tracebacks.

The flags module had no test in either language; it has one now, plus 6 route cases and 4 interpreter cases per language.

The previously red `runtime` / `integ-runtime` / `test-python-gate` jobs were the quickjs `qjs: missing file for -c` regression, from `resolve_source` starting to set argv0 unconditionally. `Argv0Rules` carries each interpreter's own answer now, including `names_file` for qjs, whose `scriptArgs` start at the arguments rather than the script path. That fix rides in this push.

zechengz:
@codex review

zechengz:
Second round resolved in 9192fa1. Three of the four were real; one was half right.

**Restrict the argv rewrite to the routed interpreter.** Real for a function, and the fix is in. I first read `_layers` and concluded it could not happen, because `SHELL_NAMES` is the *first* layer and `python3` is in it, so `route()` returns `SESSION` no matter what. That reading was wrong: `SESSION` means the shell handles the word, and within the shell bash's own order puts a function ahead of a builtin. Running it settled it:

```
$ python3() { echo "$@"; }
$ python3 -c payload -u x
-c payload -- -u x        # before
-c payload -u x           # after
```

The gate is the session function table rather than `route()`, since `route()` cannot see the shadowing. That also makes `command python3` fall out correctly for free: it masks the function for its inner run, so the rewrite applies again, which is exactly right because the interpreter is what runs. Covered in both languages, including the `unset -f` restore.

The CLI half is not reachable. `register_cli` refuses the name at install time:

```
ValueError: CLI name 'python3' collides with a shell builtin
```

so a CLI can never win routing for `python3` and never reaches this code. Noted in the comment rather than defended against twice.

**Ignore invalid warning filters without aborting.** Real, and worse than described: `_setoption` ran inside the outer `try` whose only handler is `finally`, so `_OptionError` escaped `PYTHON_WRAPPER` entirely, before `sys.stderr` was even swapped. Now caught, reported in CPython's own words, and the program runs. Byte-identical to CPython 3.13.7 on both the message and the exit code:

```
$ python3 -W nonsense -c "print(42)"
Invalid -W option ignored: invalid action: 'nonsense'
42                                                       # exit 0
```

**Implement known -X behaviors before marking them honored.** Real. `-X` is now reported per *name*, not per letter: a name CPython gives an effect beyond `sys._xoptions` is reported as unhonored, and an arbitrary name stays silent because landing in that dict is all it does on CPython either (verified: `python3 -X nosuchopt` exits 0 with `{'nosuchopt': True}`).

```
$ python3 -X dev -c "print(7)"
python3: warning: -X dev is ignored by the 'pyodide' runtime
7
```

The set is pinned against `python3 --help-xoptions` on 3.13.7 and covers both the startup-only names (`dev`, `utf8`, `warn_default_encoding`, `importtime`, `frozen_modules`, `cpu_count`, `no_debug_ranges`, `perf`, `perf_jit`, `showrefcount`) and the runtime-settable ones this wrapper does not call (`faulthandler`, `int_max_str_digits`, `pycache_prefix`, `tracemalloc`). Implementing that second group is a real option later; reporting them is the honest state today.

**Apply optimization to imported Python modules.** Real, and I took the second of the two remedies. Propagating the level would mean reaching into the import machinery for a value that is only settable at interpreter start, but reporting `-O` as ignored would be the larger lie, because the payload genuinely is compiled at the requested level. It is now written down as a bounded divergence beside the `sys.flags` one, since both have the same single cause: the resident interpreter's own level is 0 and cannot change. A module imported from `sys.path` compiles unoptimized, so its asserts run and `-OO` leaves its docstrings.

**One bug found while testing this, not fixed here.** `"prefix $@"` drops the literal prefix:

```
$ f() { echo "saw: $@"; }; f a b
a b                    # mirage
saw: a b               # bash 5, and mirage's own "saw: $*" and saw: "$@"
```

Unrelated subsystem and untouched by this PR, so it is not in scope here, but it is a real divergence and I did not want to leave it unmentioned.

**Verification.** Python suite full green; TypeScript 7200/7200 in core and green across all seven packages; integ runtime 83/83 python and 90/90 typescript; vanilla differential still 24/25 with the one known bootstrap-frame row. The `-W`, `-X` and shadowing behaviors are all asserted end to end on a real pyodide boot rather than at the unit level.

PR #746, opened 2026-08-10
fix(box,dropbox): wire the du slot in python, drop the fake find slot in ts (#609 T1-I)

bytecii:
Finishes **T1-I** — the `CommandIO` slots that are wired on one side only. With the `append` decision shipped in #745 and github's `max_glob_matches` already on main, this is the last of it.

## du — python gains it

`python/mirage/core/box/du/{size,entries,walk}.py` was written and unreachable: imported by its own `__init__` and its own tests, by nothing else. Dropbox had no python du at all. So both took `DU_BUILDER`'s readdir walk bounded by `max_du_entries`, which past 10 000 entries prints a **partial total**, a stderr notice and exit 1 — on two backends where a tree that size is ordinary.

Ports dropbox's du beside box's and wires both, matching the typescript tables.

Neither is a true pushdown — it is the same walk, uncapped, which is the choice typescript already made. Dropbox's `list_folder` does take `recursive: true` (present and unused in **both** languages) and a Box folder object carries a recursive `size`; either would make this one API sweep instead of one call per directory. Noted in the code, left as follow-up, since it needs both fake servers extended.

## find — typescript loses it

`core/box/find.ts` and `core/dropbox/find.ts` were not pushdowns either. Neither API has a search call to push down to, so both were the *same* readdir/stat walk the find and cp builders already run — minus three things the builders pass and an op cannot receive:

- **the mount's own index.** Both allocated a scratch `RAMIndexCacheStore`, so every walk started cold and warmed nothing for the next command.
- **the namespace stat overlay.** The builder's native branch applies it only when `ops.local === true`, and both backends are remote.
- **the symlink table** that `-empty` counts entries from.

The overlay half is a live divergence, not a theoretical one:

```
touch -d 2020-01-01 /data/f.txt && find /data -name f.txt -mtime +5
```

finds it in python and does not in typescript, because typescript's walk sees Box's `modified_at` (now) instead of the namespace time `touch` recorded.

The comment that justified the ops — *"the cp builder may call find without threading one (unlike Python, whose cp threads the resource index)"* — is refuted by [`cp.ts:47`](typescript/packages/core/src/commands/builtin/generic_bind/builders/cp.ts#L47), which reads `opts.index` and passes it to its own fallback walk.

## Verification

`find/mtime.json`'s four mtprobe cases now also run on `box` and `dropbox`. Proven **red before, green after, on both hosts** — reverted the sources and re-ran rather than assuming:

```
FAIL [box]     find_mtime_touch_backdate: stdout: expected "/data/mtprobe.txt\n", got ""
FAIL [dropbox] find_mtime_touch_backdate: stdout: expected "/data/mtprobe.txt\n", got ""
```

After: box 1859/0 and dropbox 2088/0 on typescript, 1859/0 and 2090/0 on python, `dropbox-root` 2074/0 on both.

du needed no new cases — box and dropbox already carry 13 between `basic/gnu/h/max_depth/error`, covering `-a`, `-s`, `-d`, `--max-depth`, `-c` and missing operands, each pinning `stderr` exactly. Those are the truncation-notice guard the plan asks for, and they still pass with the native walk underneath.

Also: 4 new python unit tests for the dropbox du; core 7160 / node 2118 / browser 182 vitest and the full python suite green; both spec trees regenerate with no drift; parity gate passes.

Box and dropbox drop from three `command_io` exceptions each to one — python range-reads, typescript's core read takes no window — which is the remaining item on the byte-window follow-up.


zechengz:
@codex review

PR #747, opened 2026-08-10
fix(notion): delete verb, one trash bit, row cells, and stale docs

zechengz:
## Summary

- `DELETE /v1/blocks/{id}` had no route in the fake, and `ntn api` had no `DELETE` method (`unsupported method: DELETE`). It is the only delete verb the public API has and the only one the MCP tool surface exposes (`API-delete-a-block`), so nothing could remove anything. Added on both sides, py + ts.
- `archived` is upstream's deprecated alias for `in_trash` and "always returns the same value". The fake stored two columns, so `PATCH {"archived": true}` returned 200 and left the row in database queries, while `ntn pages trash` (which sends `in_trash`) worked. Now one stored bit with both spellings on the wire; the `archived` column is dropped from the three prisma models.
- A database row's cells now ride in its `page.json` under `properties`, as Notion's own property objects. `jq '.properties.Priority.number'` on a row works.
- Both `ntn.mdx` files documented the pre-#740 grammar (`--page`/`--datasource` flags, `ntn blocks`/`comments`/`search`). Rewritten from the actual program.
- `resource/notion.mdx`, `typescript/setup/notion.mdx` and both `examples/*/notion/` predated the data-source split. The examples were broken: they picked the data source directory as if it were a row, then read a `page.json` that is not there, and called the removed `ntn search`.

Found while fixing the above:

- Trashing a child page moved only the `NotionPage` row, leaving the `child_page` block in the parent's listing, so the mount still showed it.
- `datasources query` derived its TSV columns from the data source schema; upstream derives them from the returned rows, so a row created from Markdown alone printed as one title among seven blanks instead of `<id>\t<title>`.
- Two MCP-parity cases (`cat_row`, `jq_db_props`) pointed at pre-data-source paths, so they compared two identical errors and asserted nothing.
- `extract_title` crashed with `AttributeError` on a non-object `properties`; the TS side already guarded.

Not fixed, separate defect: `ntn api` renders any server error as exit 1 / `ntn api: <msg>` where the real binary uses exit 5 / `error: Public API request failed (404 Not Found object_not_found): <msg>`.

## Test plan

- `notion:conformance` against the real pinned `ntn` 0.21.9: 54 checked, all match (2 skipped are mount-path cases no bare binary can run)
- integ battery `--target notion --target cli-ntn`: 99 passed / 0 failed on both hosts, byte-identical
- `notion:mcp-parity`: 33/33 (was 30/30 with 2 vacuous)
- TS unit: 7210 passed. Python notion-scoped: 192 passed. Full python suite clean apart from the pre-existing `test_uuid7_embeds_current_timestamp` suite-order failure, confirmed by re-running with every notion test excluded.
- `pre-commit run --all-files` clean, re-run to confirm the tree stops changing


zechengz:
@codex review

PR #756, opened 2026-08-11
fix(notion): add GET /v1/comments, version the database object, align ntn API errors

zechengz:
Three gaps in the shared fake Notion server and the `ntn` CLI, found while running the Toolathlon arms.

## `GET /v1/comments` did not exist

Only `POST` was routed, so a comment could be written and never read back. Two tasks read comments to grade, so nothing an agent did could pass. Added `listComments` with `block_id`, `page_size` and `start_cursor`, returning the same shape `POST` emits.

Found alongside it: comments were seeded with a single global `position++`, while `createComment` numbers by the parent's existing count. A newly written comment could therefore sort ahead of seeded ones. Seeding is now per parent, like blocks.

## The database object was version-blind

`databaseJson` always answered 2025-09-03, where the column schema lives on the data source. Toolathlon's notion MCP server sends `Notion-Version: 2022-06-28`, which real Notion answers with the schema under `properties`. The oil-price arm could not learn that `Regime` accepts `Neutral`, wrote `Normal`, and lost the graded cells, since Notion mints unknown select options rather than rejecting them.

`databaseJson` now takes the requested version, wired at `GET /v1/databases/{id}` and `POST /v1/search`. The MCP arm deliberately keeps the default: a tool call carries no version header, and that arm must render byte-identically to the REST arm.

`pageParentJson` is deliberately left on the 2025-09-03 shape; the reasoning is a comment at the function.

## `ntn` reported API failures in the wrong shape

Every verb let `NotionAPIError` reach the executor's generic fallback, which prints `ntn <verb>: <message>` and exits 1. That drops the status, the reason phrase and Notion's machine-readable code, which are the three fields a caller branches on.

```
real ntn    error: Public API request failed (404 Not Found object_not_found): Could not find block ...   exit 5
mirage      ntn api: Could not find block ...                                                            exit 1
```

`failure.py` / `failure.ts` render upstream's wording. The behaviour was probed against `ntn` 0.21.9 rather than guessed, which caught three things: 401 drops the parenthesis, adds a hint line and exits 4; a missing `code` collapses to `(404 Not Found)`; reason phrases come from the standard HTTP table.

Every leaf is registered wrapped (`partial(guarded, verb)` in Python, `guarded(verb)` in TypeScript), with a test per language that fails naming any unwrapped leaf, so a new verb cannot quietly miss it.

`datasources query` also reported the wrong 404 when an operand was neither kind of id: it surfaced the database fallback's miss rather than the data source miss plus upstream's hint.

## Tests

13 cases added to `integ/cli/ntn.json`, which is asserted twice: by the battery inside a mirage workspace, and by `integ/ntn_conformance.ts` running the same line through the real `ntn` binary against the same fake.

| gate | result |
| --- | --- |
| ntn conformance (real `ntn` 0.21.9) | 67 checked, all conformable cases match |
| cli-ntn battery, TS / Python | 69 / 69 |
| notion battery, TS / Python | 43 / 43 |
| MCP / REST parity | 33/33 byte-identical |
| full TS suite | pass |
| full Python suite | 13,989 passed, 0 failures |
| `pre-commit run --all-files` | pass |

## Known gaps, not fixed here

- A transport failure (refused connection, non-JSON body) is not a `NotionAPIError`, so it still prints the generic shape. Unreachable against the fake, so no test covers it.
- The `/notion` mount still has no comment surface; reading comments needs `ntn`.

### DIFF
--- a/integ/cli/ntn.json
+++ b/integ/cli/ntn.json
@@ -727,6 +727,175 @@
         "stdout": "a0000000-0000-4000-8000-000000000002\tRow page\n",
         "stderr": ""
       }
+    },
+    {
+      "id": "ntn_api_comments_empty",
+      "seq": 566056,
+      "targets": [
+        "cli-ntn"
+      ],
+      "command": "ntn api v1/comments block_id==aaaa1111-2222-3333-4444-555566667777 | jq -r '.results | length'",
+      "expect": {
+        "exit": 0,
+        "stdout": "0\n",
+        "stderr": ""
+      }
+    },
+    {
+      "id": "ntn_api_comment_create",
+      "seq": 566057,
+      "targets": [
+        "cli-ntn"
+      ],
+      "command": "ntn api v1/comments -d '{\"parent\":{\"page_id\":\"aaaa1111-2222-3333-4444-555566667777\"},\"rich_text\":[{\"type\":\"text\",\"text\":{\"content\":\"sheet ready\"}}]}' | jq -r '.object, .parent.page_id, .rich_text[0].text.content'",
+      "expect": {
+        "exit": 0,
+        "stdout": "comment\naaaa1111-2222-3333-4444-555566667777\nsheet ready\n",
+        "stderr": ""
+      }
+    },
+    {
+      "id": "ntn_api_comment_create_second",
+      "seq": 566058,
+      "targets": [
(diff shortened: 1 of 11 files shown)


zechengz:
@codex review

PR #760, opened 2026-08-11
shared FileHandle/FileTable for the buffered handle tables

zechengz:
R4 of the runtime/VFS arc: one shared handles package per language (`mirage/runtime/handles/`, `core/src/runtime/handles/`) now owns the buffered file-handle semantics the runtimes used to hand-roll.

## What

Split one module per concern, the same way in both languages, with the package door re-exporting:

- `flush`: `plan_flush`/`planFlush`, `FlushKind`, `NO_WRITE`. The close-time question every encoder has to answer, moved here from `runtime/vfs`: did this handle only extend the file (ship a tail) or rewrite it (ship the whole buffer)? Answering "write" always is what makes an append loop quadratic.
- `mode`: `OpenMode` and `parse_mode`/`parseMode`, the fopen-style r/w/a/x/+ string read once into five facts (writable, truncate, append, create, exclusive).
- `file_handle`: `FileHandle`, holding the whole-file buffer, position, writability and the flush facts (`base_len`, `low_write`, `dirty`), with `read`/`pread`/`write`/`pwrite`/`seek`/`truncate`/`eof`/`flush_plan`. Also `merge_writes`/`mergeWrites`, FUSE's flush expressed as what it is, a batch of pwrites over a base, implemented through the same `pwrite`.
- `file_table`: `FileTable`, the integer-id handle table (add/get/pop/set), generic over the entry type.

## Migrations

- python wasi: `WasiFs`'s stringly-typed fd dicts became typed `FdEntry` rows holding a shared `FileHandle` (stdin included, as a read-only one); ftruncate is `FileHandle.truncate`. Two lookup helpers (`_handle`, `_file_handle`) replace the entry/kind/None chain that every `fd_*` call repeated.
- ts quickjs: `GuestFile`, `writeAt` and `isWritable` deleted; `js/vfs.ts` keeps only the quickjs marshalling dialect.
- FUSE, both languages: handle ids through `FileTable`, and the sparse write merge that was duplicated between the no-handle write path and flush now goes through one `_apply_writes`/`applyWrites` helper over `merge_writes`.
- monty: `path_open` reads `parse_mode(mode).writable` (its handles belong to the upstream binding). pyodide re-imports `NO_WRITE`/`planFlush` only, exempt by design.

FUSE's `Handle` deliberately keeps its own shape instead of becoming a `FileHandle`: a handle read serves rendered bytes while its flush merges over the raw base (`raw: true`), so one buffer cannot do both without storing a rendering, and open must not hydrate known-size files (one API fetch per open).

## Verified

At the first commit, the full ladder: python suite green with mypy clean, integ runtime 88 passed against live wasi and quickjs builds at the CI pins; typescript core 7542, node 2284, browser 182, integ runtime 94 passed, typecheck/eslint/knip/prettier clean.

At the split commit, scoped to what moved: python runtime and fuse suites green, mypy clean on 1781 files, ts core runtime 277 and node fuse 117 passed, build/knip/eslint/prettier clean. Both commits: pre-commit twice byte-stable from the worktree root, yapf 0.43.0 bridge zero-diff.


zechengz:
@codex review

PR #763, opened 2026-08-12
test: put the python test tree on the source layout, and gate it (#609 item 13)

bytecii:
Item 13 of the cleanup plan — the last of the three items whose cost grows with every commit that lands before it (14 and 30 shipped in #737 and #759).

## The problem

`tests/` had drifted away from `mirage/` with nothing watching:

- **87 `__init__.py`**, against CLAUDE.md's flat prohibition and unnecessary — `pyproject.toml` already sets `--import-mode=importlib`, which resolves the same-basename collisions (34 different `test_find.py`) they would otherwise be needed for. All 87 were content-free: 81 license header only, 6 empty.
- **Three directories mirrored no source package**, and that is what hid two coverage gaps. `tests/resource/object_storage/` held 24 suites for 19 sibling packages, so `mirage/resource/s3/s3.py` looked untested and `tests/resource/s3/` did not exist. `tests/commands/builtin/helpers/` shadowed the flat `*_helper.py` sources — which is precisely why the `sed_helper` gap stayed invisible.
- **`control.py` (537 lines) and `jobs.py` (233) had no test at all** on the Python side, while `control.test.ts` and `jobs.test.ts` sit beside their sources. The only suite references were two `ReturnSignal` imports.

## What this does

**Layout** (`1503cded2`). Delete the 87. Split `object_storage/` one directory per source package, the five s3 suites into `tests/resource/s3/`, and the cross-alias prefix suite up to `tests/resource/`. Flatten `helpers/` — its conftest was a byte-identical copy of its parent's, and three of its five suites collided with a same-named flat suite and were merged, because each pair tests disjoint halves of one helper (`number_flag_error` vs `parse_counts`; the `sort` command vs `parse_keydef`/`build_config`; `grep_lines`/`grep_recursive` vs `compile_pattern`/`search_query`). Move `tests/commands/ssh/` under `builtin/`, flatten the ten single-file `{ram,redis}/{cat,head,ls,tail,wc}/` directories, and move `test_man.py` beside `man.py`.

**The gate** (`bfc34c713`), a meta-test in the style of `test_no_raw_flag_reads.py`. Three hard checks — no `__init__.py`; every test directory mirrors a `mirage/` package or is one of eight `UNMIRRORED_DIRS` entries with a written reason; no stale exemption — plus a ratchet on how many source modules no test file is named for. The ratchet counts the weaker of the two readings, and the comment says so: matching is by basename, so seven unrelated `test_history.py` files make every `history.py` look mirrored. Read strictly it would be 816 today. What it buys is narrow but real — a module whose name appears nowhere in the suite cannot be added silently.

**The missing suites** (`4b5dfde9d`). Port the TypeScript assertions for `control` and `jobs`, then cover what only Python has: `handle_cfor`, a multi-level `break` decrementing on its way out, the `_MAX_WHILE` cap warning, and `handle_fg`. Add fresh suites for the three `builtins/` modules neither tree tested — most usefully `history._parse_args`, whose bash-getopt corners were documented in a docstring and asserted nowhere (`-d3` attached, `-dc` reading "c" as the offset rather than another option, `history -1` as an invalid option, `--` and the first operand both ending option parsing). Also move `test_find_action_dispatch.py` beside its source.

## One thing deliberately not covered

`cd -P`. `_resolve_target` implements it correctly — called directly, `/link/..` gives `/deep` under `-P` and `/` under `-L` — but nothing can reach that branch: every candidate comes from `_cd_candidates` → `resolve_path`, which normpaths `..` away first, so `-L` and `-P` return the same answer for every input I could construct. The TypeScript twin is structurally identical (`resolvePath` then `resolveTarget`), so this is a shared divergence from bash, not a parity gap. Fixing it changes `cd`'s behavior and belongs in its own PR; asserting today's answer would pin the bug.

## Verification

- Collection reconciles exactly. 13863 tests before; 13935 after = +68 new tests +4 gate tests, with the moves and merges contributing zero.
- Full suite: **13476 passed, 458 skipped, 1 xfail, 0 failed**.
- The merge is AST-verified, not eyeballed: every test/class definition and every decorator from both halves is present in the merged file. The first attempt silently dropped `@pytest.mark.parametrize` and `@pytest.mark.asyncio` because `ast` node `lineno` points at `def`, not at `@` — caught by running the file, then fixed and re-checked programmatically.
- `pre-commit run --all-files` clean. The merged grep import triggered the yapf/isort deadlock (each formatter undoes the other forever); `5668ac0e4` pins yapf's form and takes the import out of isort's hands, the way `tests/commands/builtin/generic/test_split.py` and eight other sites already do.

Refs #609.


zechengz:
@codex review

PR #769, opened 2026-08-12
refactor(runtime): name the default python engine (#609 item 19 / T1-N)

bytecii:
## What

The default runtime world was an opaque positional triple on both sides:

```python
DEFAULT_ENTRIES = ("monty", "quickjs", VFSRuntime.name)      # python
```
```ts
export const DEFAULT_ENTRIES = ['pyodide', 'quickjs', 'vfs']  // typescript
```

The one fact the two implementations disagree on — which python engine a default world registers — was readable only by diffing two literals in two languages, and nothing said it was deliberate.

This lifts it to a named registration on each side and derives the triple from it:

```python
DEFAULT_PYTHON: str = MontyRuntime.name
DEFAULT_ENTRIES: tuple[str, ...] = (DEFAULT_PYTHON, QuickJsRuntime.name, VFSRuntime.name)
```
```ts
export const DEFAULT_PYTHON = 'pyodide'
export const DEFAULT_ENTRIES: readonly string[] = [DEFAULT_PYTHON, 'quickjs', 'vfs']
```

**No behavior change.** Python stays on monty, TypeScript stays on pyodide. The divergence is now one greppable constant per tree, carrying twin comments that name the reason (`@pydantic/monty` cannot answer builtin `open()` calls yet, while `pydantic-monty` can) and point at each other.

Also drops the second hardcoded copy of the same fact: the TS `Runtimes` constructor forwarded `options.python` on a literal `name === 'pyodide'` test, now `name === DEFAULT_PYTHON`.

The TS constant is a name rather than a class because `name` is an instance field there — `PyodideRuntime.name` would be the JS function name, not the registry key. Noted in the docstring so nobody "fixes" it.

## Recorded as a pin, not prose

`integ/runtime/defaults.json` reads the split back out of a live default world on each host, so the asymmetry is asserted rather than commented:

| host | `python3 -m nosuchmodule` |
| --- | --- |
| python | exit 1, `python3: -m is not supported by the 'monty' runtime` |
| typescript | exit 1, stderr contains `No module named` (real CPython under pyodide) |

The refusal names the runtime, so the python case is a direct read of which engine got registered. Both also run `python3 -c "print(6 * 7)"` so a case that skipped rather than bound would not pass silently. Suites are glob-discovered, so no registration was needed.

A `default_js_is_quickjs` case was written and dropped: `requires` is suite-level in this runner, and quickjs needs `MIRAGE_QUICKJS_HOME`, so it cannot be gated per case. The js slot is not the divergent one anyway.

## Verification

- `integ/runtime` — python host **74 passed, 0 failed**; typescript host **95 passed, 0 failed** (`defaults/default_python_is_monty` and `defaults/default_python_is_pyodide` both ok)
- `pnpm -r typecheck` — clean across all 7 packages
- `pnpm --filter @struktoai/mirage-core test` — **621 files, 7748 passed**
- `pytest python/tests/runtime/` — green
- `pre-commit run --all-files` — clean (mypy: no issues in 1795 source files)

The full `pytest` run was cut short locally by a 10-minute cap after the runtime tests passed; CI covers it.

Closes the last open item in Block C of the cleanup plan.

### DIFF
--- a/integ/runtime/defaults.json
+++ b/integ/runtime/defaults.json
@@ -1,0 +1,51 @@
+{
+  "suite": "defaults",
+  "cases": [
+    {
+      "id": "default_python_is_monty",
+      "hosts": [
+        "python"
+      ],
+      "world": {},
+      "steps": [
+        {
+          "command": "python3 -m nosuchmodule",
+          "expect": {
+            "exit": 1,
+            "stderr": "python3: -m is not supported by the 'monty' runtime\n"
+          }
+        },
+        {
+          "command": "python3 -c \"print(6 * 7)\"",
+          "expect": {
+            "exit": 0,
+            "stdout": "42\n"
+          }
+        }
+      ]
+    },
+    {
+      "id": "default_python_is_pyodide",
+      "hosts": [
+        "typescript"
+      ],
+      "world": {},
+      "steps": [
+        {
+          "command": "python3 -m nosuchmodule",
+          "expect": {
+            "exit": 1,
+            "stderr_contains": "No module named"
+          }
+        },
+        {
+          "command": "python3 -c \"print(6 * 7)\"",
+          "expect": {
+            "exit": 0,
+            "stdout": "42\n"
+          }
+        }
+      ]
+    }
+  ]
+}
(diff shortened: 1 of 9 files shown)


zechengz:
@codex review

PR #772, opened 2026-08-13
refactor(seam): dispatcher-built CommandOpts, typed adapter ops, named ParsedCommand, one DispatchFn home (#609 items 26+23)

bytecii:
## Summary

Item 26 of the cleanup plan (T2-8, seam typing — all four pieces) plus task 23 (static gate on FlagView query names), done together because the gate pins the surface the seam work reshapes.

### CommandOpts is built by the dispatcher (the T2-2 companion's endgame)

- `Mount.execute_cmd` constructs one `CommandOpts` per invocation and calls every handler as `fn(accessor, paths, texts, opts)` — the exact TS convention (`Mount.executeCmd` → `cmd.fn(accessor, paths, texts, opts)`). Flags no longer share a bag with injected context (`cwd`/`index`/`filetype_fns`/`exec_allowed` used to ride `**flags` into every command), and the `accepts_kwarg` opt-in machinery is deleted — a handler reads the `opts` fields it wants and ignores the rest, so there is no delivery registry to fall out of step.
- All 72 factory builders, 75 bespoke wrappers, and ~30 provision functions become pure wiring with the one shape (`BuilderFn` adds the leading bound `ops`). The provision path (`handle_command_provision`) builds the same bag with `command`/`spec` set, mirroring TS's `handleCommandProvision`; pushdowns that override flags pass `dataclasses.replace(opts, flags=bag)` down (dify/chroma find).
- Generics that took trailing fact parameters (`ls`, `find`×2, `tree`, `zip`, `tar`, `du`, `file`, `stat`) now read `opts.links` / `opts.stat_path` / `opts.mounts` / `opts.index` exactly like their TS twins. Two behavior alignments ride along, both in the TS direction: bespoke wrappers that never declared `cwd`/`stat_path` now see the dispatcher's real values (TS wrappers always did), and `find -empty` works on chroma/github because `find_generic` parses the full expression rather than whatever the wrapper re-declared.
- `test_no_dead_flag_params.py` now pins the exact four-positional signature for every registered handler, provision, and builder.

### Adapter op slots are typed (py)

`generic_bind/adapter.py` gains per-slot protocols (`ReaddirOp`/`ReadBytesOp`/`ReadStreamOp`/`StatOp`/`ReadRangeOp`/`WriteOp`/`ExistsOp`/`PathOp`/`RmTreeOp`/`MkdirOp`/`PairOp`/`TruncateOp`/`IsMountedOp`/`DuSizeOp`/`DuEntriesOp`/`ResolveGlobOp`/`BuilderFn`) mirroring `adapter.ts`'s generics, so wiring `readdir` where `stat` belongs stops type-checking. `find`/`set_attrs` stay loose like TS (`FindOptions` is spelled as kwargs in python and drifts per backend). The accessor parameter is deliberately `Any`: backends annotate their concrete accessor, and contravariance would reject all of them — the slot *shape* is the guard (mypy: 0 errors across 1800 files).

**Real drift the protocols surfaced, fixed here:** `hf_buckets` and `databricks_volume` `mkdir` declared `index` *before* `parents`, so a positional `parents` call would have landed in `index`; `wget -O` read its PATH-typed value with `as_str` and silently dropped the `PathSpec` (caught by `tests/commands/builtin/test_net.py`).

### ParsedCommand is named (ts)

`parseFlags` returns a `ParsedCommand` interface (twin of the python NamedTuple in `executor/command/types.py`) instead of a 15-slot positional tuple whose four identical `string[]` slots could be swapped silently; `optionError(cmdName, parsed)` matches python's `option_error(cmd_name, parsed)`. The `parsed[4]`–`parsed[14]` index soup in `command.ts` and `cli.ts` is gone.

### One DispatchFn, in the right home (ts + py)

- The workspace door type moves from `commands/builtin/generic/crossmount/types.ts` to `runtime/types.ts` — python's home for the same protocol (`runtime/types.py`), and the file that already held `BridgeDispatchFn` with a comment pointing at it. Seventeen executor/workspace importers stop reaching through the `executor/cross_mount.ts` re-export (runtime modules were importing a type from a command family's folder).
- The `CommandDispatch` duplicate in `commands/config.ts` is deleted; `opts.dispatch`, the CLI tier, and the git builtins all use `DispatchFn`. 38 python files stop spelling the same protocol `Callable[..., Any]` (executor builtins, crossmount relays, the git CLI family, monty/vfs).
- `CommandOpts.resource` had zero readers anywhere — deleted, along with the `CrossResourceStub` class that existed only to fill it and one fake resource per test file.

### Task 23: FlagView query names are gated statically

`tests/commands/test_flag_query_names.py` and `commands/flag_query_names.test.ts`: every literal `fl.as_bool("x")` / `fl.asBool('x')` must name a dest of a spec bound in the same module, so a typo'd spelling fails CI with zero coverage instead of surviving until the first runtime hit. Both gates verified to fire on a planted typo. CLAUDE.md's wrapper/flag sections are rewritten for the new contract.

### Obsolete tests, replaced not deleted

The signature-opt-in tests (`test_links_optin.py`, the `accepts_kwarg` cases in `test_run.py`) guarded a mechanism this PR removes. Their invariant survives in the new spelling: the link-aware generics must *read* `opts.links` (source-pinned), and every bespoke command shadowing a link-aware family must *delegate* to the family generic — which is also what TS has always relied on.

## Verification

- `cd python && uv run pytest` (minus fuse / the known-environmental redis deselect): full suite green, exit 0.
- Integ `ram`+`disk`, both hosts: **py 4823/4823, ts 4823/4823**.
- `pnpm -r typecheck`: clean; `pnpm -r test`: 7806 passed (7 Pyodide REPL timeouts under full-machine load re-run clean in isolation, 10/10).
- `mypy`: Success, 1800 files. `pre-commit run --all-files`: all hooks pass (the `Type check (mypy)` hook's "files were modified" is the usual `uv.lock` rewrite; mypy itself prints Success).
- The four CI-only gates reproduced locally: spec drift (both generators, byte-clean), `check_spec_parity.py` (93 commands), width-table drift (clean), `check_layout_parity.py --strict` (302 = baseline 302).

Refs #609, #292.


zechengz:
@codex review

PR #773, opened 2026-08-13
refactor(ts): derive config types from their zod schemas

zechengz:
Follow-up to #768, which collapsed `GhConfig` into its schema. The same duplication was in every other TypeScript config.

Each one declared its shape twice, as an interface and as the zod schema that actually validates it, plus a `*Redacted` twin that re-listed every field by hand. Two type helpers in `resource/secrets.ts` replace both:

- `ConfigOf<Schema>` reads a config's shape off the schema.
- `RedactedConfig<T, K>` builds the redacted twin from the config plus the names of its secret fields. The secret marker is runtime metadata on a plain `z.ZodString`, so the names still have to be written once; what goes is the half that re-listed every ordinary field (SSH repeated ten).

Neither is a bare `z.infer`. That renders an optional field as `k?: T | undefined`, which under `exactOptionalPropertyTypes` is wider than the `k?: T` these configs were written with, and it broke every S3-alias subclass against its base.

## Drift this uncovered

The twin is reached through an `as unknown as` cast, so TypeScript never checked it:

- `gcs`, `oci`, `r2`, `supabase`, `wasabi`, `hf_buckets` and node's own `s3` typed redacted credentials as plain `string`, when redaction replaces them with `'<REDACTED>'`.
- `gridfs` typed its `uri` as `uri?: string`, which after redaction is neither optional nor a string.
- `box` dropped `contentSearch` from its redacted type although the schema carries it.

## Where the config type stays hand-written

Ten files, each saying why in place: `s3` carries a `profile` its schema lacks; `google`, `box` and `dropbox` omit callbacks no snapshot can carry; the S3-alias family and `hf_buckets` describe the *resolved* config, with region and endpoint the redactor fills in. Those derive only the twin.

## Runtime

One change: browser `trello` and `linear` declared `readonly string[]` their schemas did not, so the schemas say `.readonly()` now.

## Verification

`pnpm -r typecheck` clean across all 7 packages, `pre-commit --all-files` clean and stable, and every suite green: core 7808, node 2149, browser 182, server 249, agents 175, cli 92, opencode 2. No Python files change.

### DIFF
--- a/typescript/packages/browser/src/resource/aliyun/config.ts
+++ b/typescript/packages/browser/src/resource/aliyun/config.ts
@@ -12,21 +12,14 @@
 // limitations under the License.
 // ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
 
-import { redactConfigWithSchema, secretSchema, z } from '@struktoai/mirage-core'
+import {
+  redactConfigWithSchema,
+  type ConfigOf,
+  type RedactedConfig,
+  secretSchema,
+  z,
+} from '@struktoai/mirage-core'
 import type { S3BrowserPresignedUrlProvider, S3Config } from '../s3/config.ts'
-
-export interface AliyunConfig {
-  bucket: string
-  presignedUrlProvider: S3BrowserPresignedUrlProvider
-  region?: string
-  endpoint?: string
-  defaultContentType?: string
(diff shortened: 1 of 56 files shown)


zechengz:
@codex review

PR #775, opened 2026-08-13
fix(glob): let glob expansion see nested mount roots and symlinks

zechengz:
## The bug

`du /base/*` printed only two of the four entries `ls /base` shows:

```
3	/base/f1
7	/base/sub
```

GNU coreutils 9.7 expands `base/*` to all four and prints a row for each.

## Root cause

Not `du`, and not the backend listing. `echo /base/*` printed the same
two words, so it was the expander itself.

Glob expansion enumerates candidates from one backend's readdir. A
nested mount's keys live in a different resource and no resource stores
a symlink, so neither was ever offered as a candidate. They were never
filtered out, they were never enumerated. `merge_readdir` already states
the rule for listings: "Child mounts and symlinks are namespace state no
backend can see, so a listing that stops at one backend misses both." A
glob is such a listing.

Boundary evidence:

```
BACKEND   resolve_glob('/base/*')  -> ['/base/f1', '/base/sub']
NAMESPACE namespace_names('/base/') -> ['inner', 'link']
```

Two disjoint halves of one answer.

## The fix

Both glob tiers now merge the same `namespace_names` union a listing
uses, so there is no second source of truth and no second filtering
rule. It stays session-filtered, so a scoped session still cannot learn
an ungranted mount's name from an expansion.

- Shell tier (`workspace/expand/globs.py`, `globs.ts`): merges the union,
  and the mid-path walk resolves each level against the mount that owns
  that parent, so a segment can match a mount root and keep descending.
- Command tier (`utils/glob_walk`): `opts.child_mounts` is stamped onto
  the adapter per invocation in the factory, so all ~30 builders keep
  calling `resolve_glob` unchanged.
- Matches sort, because bash sorts a pathname expansion and the two
  sources are enumerated separately.
- A glob whose directory holds a child mount is expanded before routing,
  since the mount root cannot be served by the backend being listed. That
  happens before the follow policy, so a matched symlink is followed
  exactly as a typed one is.

## Scope

The task named `du`, but the expander is shared, so `ls`, `stat`, `find`,
`wc`, `echo` and every other glob consumer had the same omission and are
all fixed.

## GNU pins

`debian:stable-slim`, coreutils 9.7, tmpfs at `base/inner`, same symlink:

```
echo base/*   -> base/f1 base/inner base/link base/sub
du -b base/*  -> 3 base/f1 / 7 base/inner / 12 base/link / 7 base/sub
wc -c base/*  -> follows the link, names each directory on stderr, exit 1
```

`du -b` is the byte mode mirage implements, and a symlink reports its
target string's length there, which is where `12` comes from.

## Left alone (pre-existing, reproducible with operands typed by hand)

- `stat` renders a mount root's name as `/`.
- `wc` omits GNU's `0 <dir>` row for a directory operand.

## Testing

Python suite, TS core (7820), node and browser suites, pre-commit and the
layout parity gate all pass. New tests in both languages; verified they
fail without the fix.

### DIFF
--- a/integ/crossmount/nested/basic.json
+++ b/integ/crossmount/nested/basic.json
@@ -207,6 +207,97 @@
         "stdout": "/ghost\n",
         "stderr": ""
       }
+    },
+    {
+      "id": "nest_glob_enumerates_child_mount_root",
+      "seq": 950017,
+      "targets": [
+        "ram-nested"
+      ],
+      "command": "echo /data/*",
+      "expect": {
+        "exit": 0,
+        "stdout": "/data/inner /data/top.txt\n",
+        "stderr": ""
+      }
+    },
+    {
+      "id": "nest_glob_matching_one_name_is_installed",
+      "seq": 950018,
+      "targets": [
+        "ram-nested"
+      ],
+      "command": "ls -d /data/i*",
+      "expect": {
+        "exit": 0,
+        "stdout": "/data/inner\n",
+        "stderr": ""
+      }
+    },
+    {
+      "id": "nest_du_glob_reaches_child_mount",
+      "seq": 950019,
+      "targets": [
+        "ram-nested"
+      ],
+      "command": "du /data/i*",
+      "expect": {
+        "exit": 0,
+        "stdout": "5\t/data/inner\n",
+        "stderr": ""
+      }
+    },
+    {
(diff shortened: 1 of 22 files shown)


zechengz:
@codex review

PR #776, opened 2026-08-13
fix(du): a namespace-only directory is not absent

zechengz:
## The bug

```
ws = Workspace({"/": RAMResource(), "/empty/hole": RAMResource()}, mode=MountMode.WRITE)
await ws.execute("du /empty")
```

```
stdout: 0	/empty/hole
        0	/empty
stderr: du: cannot access '/empty': No such file or directory
exit=1
```

The rows were right. The diagnostic and the exit code were not. `/empty` is a real directory (`ls /` shows it, `test -d` agrees), but no backend holds an entry for it: the content lives in the descendant mount's own resource.

## Root cause

Not du-specific. A command runs bound to one mount and decides existence from that mount's backend stat, so it cannot see either fact that makes such a path a directory. Both are namespace state:

- a mount nested below it (`/empty`)
- a symlink at or below it (`/ghost` after `ln -s x /ghost/deep/lnk`, since `namespace_names` synthesizes a link's ancestors)

The dispatcher already answers both correctly, and always did: it catches the backend's miss and falls back to `namespace_stat`. Probed live, `dispatch("stat", "/empty")` returns a DIRECTORY stat. That answer is offered to every handler as `stat_path`, and `find` was its only consumer, which is why `find /empty` was the one sibling that got this right.

## The fix

`du_operands` / `duOperands` now ask that probe when the backend stat fails, ahead of the `has_content` walk: one dispatch is cheaper than a full readdir walk and it is the authoritative channel. `has_content` stays behind it for a backend with no directory entry for its own mount root (redis).

## Why `stat_path` and not `registry.descendant_mounts`

1. The registry lookup is **not session-filtered**. `mount_allowed` is applied in `ops/namespace_view.py` and `fanout._allowed_descendants`, but not in the registry, and `MountView.descendants` reads it raw. Proving presence from the mount table would answer `0 /empty` for a session scoped to `/` alone and confirm a walled-off mount's parent. `namespace_stat` is filtered, so absence stays the answer there.
2. It misses half the bug class: the link-implied directory has no descendant mount at all.

Both are pinned by tests.

## GNU pin

coreutils 9.7 on `debian:stable-slim`, tmpfs mounted at `/empty/hole`:

| line | stdout | exit |
| --- | --- | --- |
| `du --apparent-size -B1 /empty` | `0 /empty/hole`, `0 /empty` | 0 |
| same, 6 bytes inside | `6 /empty/hole`, `6 /empty` | 0 |
| `-s` | `6 /empty` | 0 |
| `-a` | adds `6 /empty/hole/f.txt` | 0 |
| control `du /nope` | `cannot access` | 1 |

mirage now matches every row. GNU reserves that diagnostic for real absence. Behavior is identical with and without `-S`, so this is orthogonal to #722 (rebased on top of it).

## Not fixed here

The same false absence remains in `stat`, `wc`, `file`, `cat` and `tree` for such a path. Each needs its own GNU wording rather than the same channel read (`wc` prints `0 /empty` on stdout plus `Is a directory` on stderr; `file` prints `directory`; `cat` prints `Is a directory`), so they are a separate change. `ls`, `find` and `test -d` were already correct.

## Verification

Python full suite exit 0 - TS core 630 files / 7878 tests - `pnpm typecheck` exit 0 - `pre-commit --all-files` exit 0 with a stable tree - mypy clean - layout parity at baseline. The three new presence tests fail with the wiring reverted in both languages; the ungranted-session test passes either way, as a guard should.

### DIFF
--- a/python/mirage/commands/builtin/generic/du.py
+++ b/python/mirage/commands/builtin/generic/du.py
@@ -11,7 +11,7 @@ from mirage.commands.errors import UsageError
 from mirage.commands.spec import SPECS
 from mirage.commands.spec.types import FlagView
 from mirage.io.types import IOResult
-from mirage.ops.types import LinkView, MountView
+from mirage.ops.types import LinkView, MountView, StatPath
 from mirage.types import FileStat, PathSpec
 from mirage.utils.key_prefix import mount_prefix_of
 from mirage.utils.path import respell_raw
@@ -168,6 +168,7 @@ async def du_operands(
     stat: Callable[[PathSpec], Awaitable[FileStat]],
     has_content: Callable[[PathSpec], Awaitable[bool]] | None = None,
     links: LinkView | None = None,
+    stat_path: StatPath | None = None,
 ) -> tuple[list[PathSpec], list[str]]:
     """Split the operands into the ones du can read and the ones it cannot.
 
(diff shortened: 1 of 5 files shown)


zechengz:
@codex review

PR #780, opened 2026-08-13
fix(find): `!` is negation, not a start point

bytecii:
GNU spells `find`'s negation two ways and mirage only understood one:

```
$ find /data -name '*.txt' ! -empty
find: unknown predicate '/!'          # exit 1
$ find /data ! -name '*.txt'
find: '!': No such file or directory
```

`-not` worked. So did `\( ... \)`. The cause is one missing entry in a
frozenset: find's spec declares `rest=Operand(type="path")`, so every word
without a leading dash is classified as a start point, and `!` was
cwd-prefixed into the path `/!`. The parens were already exempted through
`ignore_tokens`; `!` simply was not in that set. Nothing downstream needed
teaching — `!` has always been in the expression parser's operator table.

## A second bug the fix made reachable

`find /data !` used to print the whole tree and then a bogus
`No such file or directory` for `/!`. With the phantom operand gone it
would have negated nothing, in silence. Both parsers now refuse it in GNU's
words:

```
find: expected an expression after '!'
find: expected an expression between '!' and ')'
```

The check runs where the operator is consumed rather than in the primary,
because by the time recursion reaches a primary the token that needed a
right-hand side is no longer on the line.

## A live Nextcloud bug the new cases caught

Nextcloud's `find` pushes the predicate tree down as a WebDAV SEARCH.
Nextcloud inverts a `<d:not>` by swapping the comparison inside it for
its opposite, so a `<d:not>` around anything but a single comparison
raises server-side and comes back as a 500. Probed against
`nextcloud:30-apache` (30.0.17), the image CI runs:

| condition | |
|---|---|
| `not(like)` | HTTP 207 |
| `not(is-collection)` | HTTP 207 |
| `or(like,like)` | HTTP 207 |
| `not(or(like,like))` | HTTP 500 `Binary operators inside "not" is not supported` |
| `not(and(like,like))` | HTTP 500 same |
| `not(not(like))` | HTTP 500 same |
| `not(not(is-collection))` | HTTP 500 same |

The compiler now tracks whether a compiled condition is a bare
comparison and refuses to negate one that is not, which falls that
`find` back to the scan walk — the same "cannot express it, return
None" contract the module already uses for `-path` and `-empty`. No
behavior is lost: the shapes it now declines were returning 500.

**`-not` reached this first.** `find /data -not -type f` compiles to
`not(not(is-collection))` and has been answering 500 on Nextcloud all
along — unnoticed because `find_not_name` was the battery's only
negation case and it negates a bare `-name`. `find_bang_type_f` pins
it now.

Deliberately not done here: De Morgan normalization
(`not(or(a,b))` → `and(not a, not b)`) would keep the push-down instead
of falling back. That is an optimization on a path that is currently
broken, with its own test surface, and belongs in its own PR.

## Two py/ts divergences closed on the way

- The TypeScript parser skipped an ignore token **silently**, leaving its
  `wordKinds` slot `null` — which means "guess from the shape", and the
  shape of a grammar token says nothing about it. Python stated TEXT.
  TypeScript states it now too.
- `spec_hints.ts` carried a by-value override that re-nulled ignore tokens
  after the fact. Redundant once the parser states the kind, and
  **position-blind**: it matched an option's *value* as readily as a
  grammar token, so `find /d -name '!'` lost the TEXT the parser had
  correctly given the pattern. Deleted.

## GNU pinning

Pinned against GNU findutils 4.10.0 on `debian:stable-slim` as a three-way
byte-identical differential — GNU vs python vs typescript — over 25 output
cases and 15 error cases on one shared seed tree.

One known divergence is normalized and commented: GNU quotes as `` `x' ``
where mirage quotes `'x'` repo-wide. That is pre-existing and already
pinned in `integ/unix/find/error.json`. Also left alone as pre-existing:
GNU's wording for a binary operator with nothing before it, the
paren-specific messages (mirage says `unbalanced parentheses`), and
`paths must precede expression`.

## integ

Five output cases in `find/not.json`, three error cases in
`find/error.json`.

The output cases carry **`seq` 295**, the same seq as `find_not_name` whose
expected output they derive from. This is load-bearing, not cosmetic: the
battery runs one shared session ordered by `seq`, and every fixture that
would otherwise intrude into `/data` is created after 295 and never removed
— `/data/fs` (302), `/data/dr` (317), `/data/rgm` (459), `/data/arch`
(467), `/data/roots1` (785), `/data/xm2ctree` (755), `/data/g4` (500200),
`/data/f4` (500300). A first attempt at 500318+ failed 8 cases for exactly
that reason. The one case already sitting at 295,
`find_empty_dir_holding_only_a_link`, removes its own tree in the same
line, which is what leaves the slot usable.

chroma and github each gain the `!` twin of the `-not` case #772 added,
placed immediately after it in both file order and seq:

| twin of | new case | seq |
|---|---|---|
| `chroma_find_empty_composes_with_name` | `chroma_find_bang_composes_with_empty` | 630056 |
| `github_find_empty_composes_with_name` | `github_find_bang_composes_with_empty` | 530014 |

Both verified against a locally run fake on both hosts. Note these two
cases prove that `!` parses and composes on a SaaS backend — they do not
discriminate on `-empty`, since neither fixture holds an empty `.md`, so
their expected output equals the plain `-name` case's.

## Verification

All of the following on this branch, rebased onto `main` at 0666c1666:

| | |
|---|---|
| python pytest (full) | exit 0 |
| mirage-core vitest | 631 files, 7902 tests, 0 real failures¹ |
| integ, python host, ram+disk | 4863 passed, 0 failed |
| integ, TS host, ram+disk+opfs | 7145 passed, 0 failed |
| integ, chroma+github | 85 passed, 0 failed on **each** host |
| integ, nextcloud, live server | see below² |
| `pnpm -r typecheck` | clean, 8/8 packages |
| pre-commit | all hooks pass; mypy `Success: no issues found in 1800 source files` |
| spec regen + `git diff --exit-code spec/` | no drift |
| `check_spec_parity.py` | 93 commands match |
| width table | no drift |
| `check_layout_parity.py --strict` | 302 unexcused, baseline 302 |

¹ Two Pyodide tests in `python_repl.test.ts` time out under parallel load
and pass in isolation (10/10). Nothing here touches the Python runtime.

² Against a live `nextcloud:30-apache` (30.0.17) per host, one container
each. Stopped after 838 (python) / 835 (typescript) cases with **0
failures** — two Nextcloud+SQLite instances sharing one Docker VM run
several times slower than CI's dedicated runners, and by that point every
case this change can reach had already run green on both hosts: all six in
`find/not.json`, both dangling-operator cases in `find/error.json`, and
`find_start_empty_mount_not_empty`. Those are every negated predicate in
the battery; a tree with no `Not` node compiles byte-identically to before.
The exhaustive pass is CI's.


zechengz:
@codex review

PR #785, opened 2026-08-13
fix(shell): don't unescape an already-expanded word in classification

zechengz:
## Summary

`ln -s "$(printf '/data/x\ty')" /data/tabby` stored a target with no tab.

`classify_word` / `classifyWord` ran a second POSIX quote-removal pass (`shlex.split`) over words the expansion layer had already unquoted. A backslash surviving expansion is a literal name character, so the pass corrupted it. `printf`'s format arg is path-shaped, so the tab was destroyed before printf ran.

`ln` and the namespace symlink write are innocent; both store verbatim.

- delete `_unescape_path` / `unescapePath` and both call sites, py + ts
- the relative-branch call site was unreachable (`_RELATIVE_PATH` admits no backslash)
- side effect: a file named `a\b` is addressable for the first time, matching GNU

## Test plan

- pinned against GNU coreutils 9.7 in docker `debian:stable-slim`
- unit: `test_heuristic.py` / `heuristic.test.ts` (the old tests pinned the bug), `test_quoting_coverage.py`, `test_ln.py`, `native_ln.test.ts`
- integ: new `integ/unix/sym/verbatim.json`, seq 820-825
- full py suite, all 8 ts packages, integ both hosts, pre-commit: green
- reverted just the two source files to confirm 3 of 5 new integ cases fail identically on both hosts

### DIFF
--- a/integ/unix/sym/verbatim.json
+++ b/integ/unix/sym/verbatim.json
@@ -27,7 +27,7 @@
         "gdrive-folder",
         "box"
       ],
-      "command": "mkdir -p /data/symv",
+      "command": "mkdir -p /data/symv && echo alpha > /data/symv/t.txt",
       "expect": {
         "exit": 0,
         "stdout": "",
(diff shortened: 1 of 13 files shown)


zechengz:
@codex review

PR #788, opened 2026-08-13
fix(shell): ambient session for nested evals, whole-body cmdsub

zechengz:
## What

Three related shell-core fixes, mirrored in Python and TypeScript.

**Nested evals run in the live ambient session.** A re-entrant execute (`$()`, backticks, `eval`, `source`, `xargs`, `timeout`, `command`, `env`, process substitution, or an embedder callback fired mid-line) used to re-resolve its session by id through the session manager. An id cannot name the ephemeral per-call cwd/env fork the outer line actually runs in, and TypeScript dropped the id entirely, so on a named session `$(pwd)` answered the default session's cwd and a restricted session could read forbidden mounts through a command substitution. `execute_line`/`executeLine` now continue in the ambient live session when the call names no session or the same id; every eval site inherits this from the one resolution point.

**Background jobs rebind the ambient session.** `&` forks the session but the fork keeps its parent's id, so a nested eval inside a background job resolved the outer session: `$(pwd)` read the parent's cwd, and `$(cd ...)` moved the parent. The job task now runs inside its fork's ambient scope in both languages.

**Command substitution executes its whole body.** The expansion filtered the parse tree to command-like nodes and ran only the first one, so `$(echo a; echo b)` ran only `echo a`, and declarations, assignments and control flow (`$(export X=1)`, `$(X=5; ...)`, `$(if ...)`, `$(for ...)`) ran nothing at all. The whole text between `$(` and `)` now goes to the evaluator.

One documented divergence is pinned in the tests: bash runs `$()` in a subshell, mirage runs it in the live session, so a `cd` inside a substitution on the unforked session persists.

## Tests

- Red-first unit coverage: `python/tests/workspace/workspace/test_execute.py` and `typescript/.../workspace/workspace/execute.test.ts` (12 cases each side: named-session, per-call fork, confinement, bg-job fidelity, whole-body cmdsub, divergence pin), plus two `test_expand.py` cases pinning the exact text handed to the evaluator.
- Six shared integ cases, byte-pinned against real bash: `integ/bash/jobs/bg.json` (+2) and new `integ/bash/cmdsub/body.json` (+4).

## Verification

- Python battery, facet core: green across all local targets (the two bg cases and four cmdsub cases pass on every target).
- TypeScript battery, facet core: 24479 ok, 0 failures.
- Parity (`parity.py ram disk redis ram-history`, the CI diff): 7264 case/target pairs, 0 mismatches.
- Full Python suite, TS core/node/browser/dsh suites, `pnpm -r typecheck`, pre-commit twice with a stable tree, layout parity at baseline.

### DIFF
--- a/integ/bash/cmdsub/body.json
+++ b/integ/bash/cmdsub/body.json
@@ -1,8 +1,8 @@
 {
   "cases": [
     {
-      "id": "job_bg_then_wait",
-      "seq": 502400,
+      "id": "cmdsub_body_runs_all_statements",
+      "seq": 930320,
       "targets": [
         "ram",
         "disk",
@@ -27,16 +27,16 @@
         "gdrive-folder",
         "box"
       ],
-      "command": "(sleep 0.2 & wait) 2>/dev/null; echo done",
+      "command": "echo $(echo a; echo b)",
       "expect": {
         "exit": 0,
-        "stdout": "done\n",
+        "stdout": "a b\n",
         "stderr": ""
       }
     },
     {
-      "id": "job_bg_write_visible_after_wait",
-      "seq": 502402,
+      "id": "cmdsub_body_control_flow",
+      "seq": 930322,
       "targets": [
         "ram",
         "disk",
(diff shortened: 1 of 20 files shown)


zechengz:
@codex review

PR #798, opened 2026-08-14
feat(ts): install a CLI from a file reference in yaml

zechengz:
## Summary

- `clis: {tally: {cli: ./tally.mjs:TALLY}}` now resolves in TypeScript. Previously `cli:` could only name a spec compiled into mirage, so a foreign CLI needed a host program calling `registerCli`.
- New `packages/node/src/resource/loader.ts` (`loadAttr`), the mirror of `mirage/resource/loader.py`. Closes that layout-parity divergence, so the baseline drops 301 -> 300.
- `buildCliEntries` resolves a ref and `absolutizeCliRef` rebases a relative one against the config file, matching `script:`. Resolution sits in the config layer, not `cliSpecFor`: that lives in core, which has no filesystem and is synchronous.
- `.mjs`, `.js` and `.ts` all work. A `.ts` spec rides Node's strip-only type stripping, so `enum` / parameter properties / runtime `namespace` are refused at load with a hint.
- `instanceof CLISpec` is the fast path, not the test: the referenced file may resolve its own copy of `@struktoai/mirage-core`. A plain object literal is still refused.
- Docs corrected; they previously said TypeScript had no string loader.

## Test plan

- `integ/cli_ref.sh`, a new battery wired into `test_cli.yml`. Goes through `mirage workspace create` with a relative ref, so the rebase is load-bearing. Python loads `.py`, TypeScript loads `.mjs` / `.ts` / `.js`, all four answer `total 9 kg`, `man tally` renders, `type -t tally` is `cli`, and a ref naming a missing export is refused at create time.
- `packages/node` 2349 passed, `packages/server` 258 passed, `pnpm -r typecheck` clean.
- Full Python suite exit 0.
- `pre-commit run --all-files` exit 0, `check_layout_parity.py --strict` exit 0.

Note: the loader tests spawn a child `node` for the `.ts` cases. Vitest imports `.ts` through Vite's transform, which compiles what Node only strips, so an in-process test cannot see the strip-only cliff.


zechengz:
@codex review

zechengz:
Both Codex findings were real and are fixed in acaa2ba18.

**P1, scoped and subpath specifiers.** Confirmed: `@scope/my-clis:JIRA` and `my-clis/specs:JIRA` were classified as file paths, rebased under the config directory, and then reported as missing files. My own test pinned the bug (it asserted `isModulePath('my-clis/specs') === true`) and the docs I added promised the specifier form worked.

Root cause is the parity trap: I copied Python's `"/" in source` test, which is sound there because a module dotpath is dotted, so a slash can only mean a path. In Node a bare specifier carries slashes all the time. `isModulePath` now uses Node's own rule, with the one Python-style convenience kept and narrowed:

| source | before | after |
|---|---|---|
| `@scope/my-clis` | path | specifier |
| `my-clis/specs` | path | specifier |
| `pkg/dist/index.js` | path | specifier |
| `./dist/index.js` | path | path |
| `tool.mjs` | path | path |

The suffix rule now applies only to sources with no slash, so a subpath export ending in `.js` stays a specifier and `./dist/index.js` says path by saying so.

**P2, constructor name is not proof.** Agreed on the substance: the walker reads `subcommands.find`, `aliases.includes` and `options.some` at every level, so a value passing a name check crashed on the first line an agent typed rather than failing the create. Replaced with a recursive shape check over exactly those fields plus the constructor's own "needs fn, subcommands, or script" invariant.

I kept the non-`instanceof` path rather than demanding a brand. The referenced file lives in someone else's project and may resolve its own copy of `@struktoai/mirage-core`, which is the case the feature exists to serve; a brand would need a core change on both language sides. Three new tests cover an impostor class named `CLISpec`, a malformed nested subcommand, and specifiers surviving the rebase.

Verification: node 2357, server 276, `pnpm -r typecheck` clean, `integ/cli_ref.sh` PASS on all four arms, python `tests/config` green, `pre-commit --all-files` and `check_layout_parity.py --strict` exit 0.

PR #799, opened 2026-08-14
Release 0.0.5

zechengz:
Release 0.0.5, cut from main at 5cfb4598b (170 PRs since v0.0.4).

- `python/pyproject.toml` + `python/uv.lock` to `0.0.5`
- the six published TypeScript packages (agents, browser, cli, core, node, server) to `0.0.5`
- `@struktoai/mirage-dsh` to `0.0.1`, its own line, first release
- `opencode` stays at 0.0.0

Cross-package deps are `workspace:*`, so pnpm rewrites them at publish time and no dependency edits are needed; `pnpm-lock.yaml` is unchanged for the same reason. Both `mirage.__version__` and the TS side read their version from package metadata, so there is no second place to bump.

### Validated as 0.0.5-alpha.1 first

The alpha was published to both registries and exercised end to end before this PR moved to the final version.

**PyPI** (`mirage-ai 0.0.5a1`, wheel + sdist, `twine check` clean): in a clean venv installing only from PyPI, `mirage.__version__` reported `0.0.5a1`, the `mirage` console script ran, the shell smoke test passed (redirect + `cat`, `sort | tr`, `find | wc`, `du -a`, `history`), and plain `pip install mirage-ai` still resolved to 0.0.4.

**npm** (all seven packages, `--tag alpha`): `latest` stayed on 0.0.4 for the six, `workspace:*` deps were rewritten to real versions in the published manifests, and a clean project installing only from the registry ran the workspace commands, the dsh shell seam (reporting `sandbox: workspace-write`), incremental background streaming, and spill-with-recovery from #796.

### One test fix

Five TypeScript `--version` assertions pinned the output to `\d+\.\d+\.\d+`, which cannot express a prerelease, so all five failed under `0.0.5-alpha.1`. Python asserts the same output with `startswith` and was unaffected. The shape is still asserted, now with an optional prerelease suffix, so a future alpha does not break CI again.

After merge: tag `v0.0.5`, publish the release page, then upload 0.0.5 to PyPI and npm.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "mirage-ai"
-version = "0.0.4"
+version = "0.0.5a1"
 description = "A unified virtual filesystem for AI agents. Mount S3, Google Drive, Slack, Gmail, GitHub, Linear, Notion, Postgres, MongoDB, SSH, and more behind one filesystem so agents read, write, and pipe across services with familiar shell commands."
 readme = "README.md"
 license = "Apache-2.0"
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "mirage-ai"
-version = "0.0.5a1"
+version = "0.0.5"
 description = "A unified virtual filesystem for AI agents. Mount S3, Google Drive, Slack, Gmail, GitHub, Linear, Notion, Postgres, MongoDB, SSH, and more behind one filesystem so agents read, write, and pipe across services with familiar shell commands."
 readme = "README.md"
 license = "Apache-2.0"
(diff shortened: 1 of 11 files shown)


zechengz:
PyPI is published and validated from this branch: **mirage-ai 0.0.5a1** (wheel + sdist, both `twine check` clean).

Validated in a clean venv installing only from PyPI:
- `mirage.__version__` reports `0.0.5a1`, imported from site-packages
- shell smoke test passes end to end: redirect + `cat`, `sort | tr` pipeline, `find | wc`, `du -a`, `history`
- `mirage` console script runs
- plain `pip install mirage-ai` still resolves to **0.0.4**, so the prerelease does not become the default install

npm is not published yet, pending this merge and the tag.

PR #801, opened 2026-08-15
docs: fix the README hero snippet, resync the mirrors, and give each CLI page its own icon

zechengz:
Three related documentation fixes, cut off latest main.

### The hero snippet did not work

`README.md` opened with `ws.command('summarize', ...)`. That method exists in neither language: `typeof ws.command === 'function'` is `false` on the published `mirage-node`, and `hasattr(ws, "command")` is `False` on `mirage-ai`. Registration is the standalone `command({...})` plus `mount.register(...)`.

It is replaced with a Python example that shows more of what 0.0.5 can do: ram, redis and slack mounted side by side under per-mount modes, monty capturing `python`/`python3`, one grep across sources, a script hosted in Slack writing into Redis, and a typed CLI installed under a head word. Every call in it was run against the published 0.0.5 packages before it was written down, including `MontyRuntime(captures=[...])`, `RedisResource(url=...)`, `SlackResource(SlackConfig(token=...))`, `ws.register_cli("slack", SLACK, {...})` and the real `send-message --channel --text` grammar.

Two smaller corrections: the filetype sentence promised parsed PDF pages, which the filetype removal took away, and DeepSeek Harness was missing from the coding agents row.

### The eleven mirrors had drifted

They are regenerated from the root rather than patched, so they pick up the fixes above and also the drift they had accumulated on their own: the `agents` mirror was missing GridFS and Qdrant from the backend list, every mirror still described coding agents as "via the lightweight CLI + daemon", none had Grok Build, and their Codex link pointed at `python/agents/codex` instead of the TypeScript page. Transforms verified intact afterwards: per-package npm badges, `raw.githubusercontent` asset URLs, the dynamic license badge, absolute readme links, and no Contributors section.

The five translations carry the same four fixes with the prose translated. **These translations are machine-generated and deserve a native-speaker pass before anyone relies on them.**

### Every CLI page had the same icon

All nine shared `icon: terminal`, so the sidebar was nine identical rows. Each now uses the icon its service already uses elsewhere in these docs: `slack`, `discord`, `github`, `google`, `envelope` for himalaya, `book` for ntn and `chart-gantt` for linear (matching `setup/notion.mdx` and `setup/linear.mdx`, since Font Awesome has no brand mark for either), and `git-alt` for git. `gws` and `himalaya` are now spelled GWS and Himalaya; the rest stay lowercase because that is the head word you type.

### One example

`examples/python/filetype/filetype.py` reached through `ws._registry.mount_for(...)` when `ws.mount(...)` is public and returns the same `MountEntry`. Re-ran it after the change: output identical, so the CI truth file still matches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,25 +34,29 @@
 
 Mirage is **a Unified Virtual File System for AI Agents**: it mounts services and data sources like S3, Google Drive, Slack, Gmail, and Redis side-by-side as one filesystem. Any LLM that already knows bash can read, grep, and pipe across every backend out of the box, with zero new vocabulary.
 
-```ts
-const ws = new Workspace({
-  '/data':  new RAMResource(),
-  '/s3':    new S3Resource({ bucket: 'logs' }),
-  '/slack': new SlackResource({ token: process.env.SLACK_BOT_TOKEN! }),
-})
+```python
+ws = Workspace(
+    {
+        "/tmp":   (RAMResource(), MountMode.EXEC),
+        "/redis": (RedisResource(url=redis_url), MountMode.WRITE),
+        "/slack": (SlackResource(SlackConfig(token=slack_bot_token)), MountMode.EXEC),
+    },
+    # monty captures python, so scripts run sandboxed inside the workspace
+    runtimes=[MontyRuntime(captures=["python", "python3"]), "vfs"],
+)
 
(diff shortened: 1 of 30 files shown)


zechengz:
@codex review

PR #812, opened 2026-08-15
chore(examples): typecheck examples/typescript in CI

zechengz:
## Summary

`examples/typescript` is a pnpm workspace member with a tsconfig but no `scripts` block, so `pnpm -r typecheck` skipped it. Adding a `typecheck` script wires it into the existing CI step (no workflow change needed) and fixes the 32 errors it reports.

- 21 FUSE examples used `MountBackend` without importing it. These failed at runtime with `ReferenceError: MountBackend is not defined`, so they could never have run.
- `hf_datasets`/`hf_spaces` imported `HfDatasetsConfig`/`HfSpacesConfig`, which do not exist. Both resources take `HfRepoConfig`.
- `redis/redis.ts` passed the provision helpers `{ command }` in the `texts` slot and omitted `opts`. Now matches the Python example's `(accessor, paths, [], CommandOpts(command=...))`.
- `postgres_fuse`/`postgres_vfs` read a module-level `dsn` inside `main()`, where the `undefined` guard does not narrow.
- `s3_browser`/`browser/scripts/presigner` passed `Record<string, unknown>` to `ListObjectsV2Command`.
- `lancedb`: `registry.get()` may return `undefined`, and `TextEmbeddingFunction.sourceField()` takes no argument (it overrides the base to hardcode `Utf8`, so dropping the argument is behavior-identical).

The tsconfig has no `include`/`exclude`, so this covers 155 files including `browser/scripts/*.ts`, which the browser package's own tsconfig (`include: src/**`) never checked.

## Test plan

- `pnpm -r typecheck` from `typescript/`: exit 0, examples included
- Negative control: deleting one `MountBackend` import makes that command exit 2 and name the file/line, so the gate is not vacuous
- `redis/redis.ts` (run in CI against a truth file): `OK: integ/truth/typescript/redis.txt`, provision output unchanged
- `lancedb.ts` runs end to end, exit 0
- `pre-commit run --all-files`: exit 0, no formatter churn
- `knip`: exit 0

Python is untouched, so the Python suite was not run.

## Not addressed

`ts-prettier` and `ts-eslint` are scoped `files: ^typescript/.*`, so nothing formats or lints `examples/typescript` either. Edits here follow each file's local style.


zechengz:
@codex review

PR #814, opened 2026-08-15
chore(ts): sweep the dead exports out of core (D28)

zechengz:
Now that #805 is merged, `packages/core/src/index.ts` is 102 lines instead of 1589, so the barrel no longer keeps unused core exports artificially alive and knip can see them. This is a single commit on top of main.

## Reproducing

Add `"packages/core": {"includeEntryExports": true}` to `typescript/knip.json` and run `pnpm exec knip`. Two caveats, both of which cost real time:

- **Build first.** `core/package.json` maps `./*` to `./dist/*.js`, so with no dist knip mis-resolves and reports 364 instead of 233. Run `pnpm -r build`.
- **Typecheck the way CI does.** `pnpm typecheck` in `typescript/package.json` filters to `./packages/*`; CI runs the unfiltered `pnpm -r typecheck` over 11 projects, `examples/typescript` among them. The filtered one hides real breakage.

## 74 of the 233 findings are false

They are the `index.ts` lines. knip's project root is `typescript/`, and the barrel's consumers live in `examples/`, `docs/` and `integ/`, outside its graph. That is why `scripts/check_barrel_surface.py` exists, and it still passes.

## Six more are false

knip strips a barrel line and then cascades into whatever that line alone kept alive, so these survive and stay exported:

| Symbol | Why |
| --- | --- |
| `CommandContext`, `ExecuteResultContext` | re-exported via `policy/index.ts` |
| `RAMWatchQueue`, `Watcher` | re-exported via `watch/index.ts` |
| `S3BrowserOperation` | re-exported by `packages/browser` |

`S3BrowserOperation` is the instructive one: `tsc --noEmit` **passed** while `pnpm build` failed on it, because typecheck was reading the stale dist.

## The other 153

| Outcome | Count |
| --- | --- |
| Dead re-export lines removed from sub-barrels | 61 |
| `export` dropped, symbol still used in its own file | 52 |
| Deleted outright | 40 |

71 files, -396 / +76.

Two worth spot-checking rather than trusting the tool: `numberLines` in `generic/cat.ts` is a dead generator that merely collides with the live `display.numberLines` flag (`cat -n` numbers inline), and gdrive's `readBytes` was a one-line pass-through to `downloadFile` that nothing called.

Re-running knip with `includeEntryExports` after this change leaves exactly the 71 barrel lines plus those 6, and nothing else.

`knip.json` ships unchanged: `pnpm exec knip` is a pre-commit gate and `includeEntryExports` would fail it on the barrel lines.

## Verification

Build (8 packages), `pnpm -r typecheck` (11 projects), eslint, prettier, knip under the committed config, `check_barrel_surface.py` (77 names, both directions), `check_layout_parity.py --strict` (300, on baseline), `check_case_targets.py --strict` (2297, on baseline).

Tests: core 8287, node 2382, browser 182, server 270, agents 175, dsh 87, cli 92, opencode 2.

## Noticed, not fixed

`listVersions` in `core/onedrive/index.ts` is dead while gdrive has live version wiring, so OneDrive versioning may never have been wired up. Left alone as out of scope.


zechengz:
@codex review

PR #825, opened 2026-08-16
feat(py): mirage mcp over stdio, and invert the mypy allowlist (items 29 + 28)

bytecii:
Cleanup-plan items **29** (`mirage mcp` for Python) and **28** (invert the mypy allowlist). Mostly Python; the review round below found three bugs the TypeScript side shared, so two `packages/agents` files change too.

## Item 29 — `mirage mcp`

TypeScript ships a six-tool stdio MCP server (`cli/src/mcp.ts` → `agents/src/mcp/server.ts`). Python had none: a `pip install` user could not point Cursor or Claude Desktop at a workspace, and `mirage --help` differed between distributions.

Adding just the entry point would have duplicated the tools, because this side kept them private inside the Claude Agent SDK integration (`_MirageTools` in `agents/claude_agent_sdk/server.py`). So the shared layer lands first, mirroring TS's `tool-descriptions.ts` / `tool-operations.ts` split:

| new module | TS twin | why |
|---|---|---|
| `agents/tool_descriptions.py` | `tool-descriptions.ts` | the six strings, one copy |
| `agents/tool_operations.py` | `tool-operations.ts` | `MirageToolOperations`, lifted out of the private `_MirageTools` |
| `agents/file_version.py` | `file-version.ts` | stale-write protection, which this side lacked entirely |
| `agents/mcp/server.py` | `mcp/server.ts` | the server |
| `cli/mcp.py` | `cli/mcp.ts` | `mirage mcp`, `MIRAGE_MCP_CONFIG` then `MIRAGE_CONFIG` |
| `server/workspace_config.py` | `server/src/workspace_config.ts` | config discovery, which Python had nowhere |

`claude_agent_sdk/server.py` now delegates instead of owning the logic; its 104 existing tests pass unchanged.

Three deliberate choices worth reviewing:

**Low-level `Server`, not `FastMCP`.** `FastMCP.__init__` takes no `version` and does not forward one to the server it builds, and TS advertises `VERSION`. The low-level API also lets the input schemas state the same constraints TS's zod objects do — `offset` min 0, `limit` min 1 — which the SDK then enforces.

**Stamps cover rendered bytes, not stored bytes.** TS's tracker fingerprints `readFile(path, {raw: true})`. Here that would let `edit` search bytes the agent never saw, because this side's read tool has always gone through `ops.read` (rendered). Same bytes for read and edit, or `old_string` comes from a view `edit` cannot see.

**`mcp` is a hard dependency**, as it is in TS (only the agent *frameworks* are optional there). The base install already carries fastapi/uvicorn/httpx/pydantic, so the added surface is small, and `mirage --help` is now identical across distributions — the defect the plan named.

Verified end-to-end over real stdio with an MCP client, not just in-process:

```
server: mirage 0.0.5
tools: ['edit', 'execute_command', 'grep', 'ls', 'read', 'write']
write -> Written: /a.txt | isError: False
read  -> '     1\tx\n     2\ty\n     3\tz\n'
exec  -> 3 /a.txt
miss  -> Error: file '/nope.txt' not found | isError: True
bad   -> Input validation error: -1 is less than the minimum of 0 | isError: True
```

## Item 28 — invert the mypy allowlist

54 modules opted *in* to annotation checking, against 1826. The default was unchecked, so every new file joined the unchecked side and the handles CLAUDE.md prescribes types for (`accessor: Accessor`, `index: IndexCacheStore | None`) went unannotated wherever the list did not reach.

The default is now `disallow_untyped_defs` + `warn_return_any`, with an explicit list of what is not yet annotated that **only shrinks** — the inverse of what was there. 166 annotations cleared here (595 → 429 before the remainder was listed); `mypy` reports **Success: no issues found in 1839 source files**.

Two real defects surfaced, neither of them about typing:

- **`Workspace._original_open` / `_original_os` were invented by assignment.** `lifecycle.patch_process` set them; nothing declared them. An `unpatch_process` without a prior `patch_process` raised `AttributeError` instead of restoring nothing. Declared on the class, and the restore is guarded.
- **`sed_generic` declared a non-optional writer.** Its own docstring said `Callable | None` and its body has a `write_bytes is None` branch; the builder passes `None` whenever the backend cannot write.

### On the `tests/` half — a scope call to review

The plan justified type-checking `python/tests` with "the PathSpec-not-str rule is unenforced across 1220 Python test files." Measured on this tree, full strict on `tests/` is **2374 errors**, of which:

- **634 (27%)** are `str` passed to a pydantic `SecretStr` field — which pydantic v2 coerces at runtime (checked, not assumed). Not defects.
- most of the remainder is `union-attr` / `attr-defined` / `method-assign` from the monkeypatched fakes CLAUDE.md explicitly sanctions.
- **19** are the PathSpec rule.

So full strict would buy 19 real findings for ~2350 mechanical edits that make the suite noisier. Instead all 19 are fixed and `scripts/check_test_pathspec.py` holds that one line; the mypy hook's file regex now also triggers on test changes, which it previously could not.

One of the 19 was a latent crash rather than a style nit: `tests/e2e/test_snapshot_drift_live.py` passes a `str` to s3 `write_bytes`, which reads `path_spec.mount_path`. The test skips without a live versioned bucket, so nothing ran it.

If you would rather have the full-strict treatment on `tests/`, say so and I will do the other ~2350.

## Review round (`cb1b2c28`)

Merged `upstream/main` first — two conflicts, both in signatures this PR annotated and #821 extended (`tar_generic`, `unzip_generic` gained `relay`, and unzip a `stat`); resolved by keeping both.

All four codex findings are real, and **three of them are shared with TypeScript**, so both sides are fixed and tested together rather than one side being nudged into divergence.

| finding | what actually happens | shared with TS |
|---|---|---|
| P2 write stamp | `_record_write` stamped the bytes handed in, while every later check reads them back. On a mount whose read op renders, the agent's own next write is refused as somebody else's change. Both sides now stamp what a later read returns. | yes |
| P2 alias keys | stamps were keyed by the caller's spelling, but read and write follow the namespace symlink table, so `/alias` and `/target` got separate stamps and an edit through the other name skipped the check entirely. Keyed through `namespace.follow` now. | yes |
| P2 grep exit | every exit code came back as success. Exit 1 is "no match" and stays a success; `>1` (verified: an unreadable path exits 2 on both runtimes) is now an error. | yes |
| P1 `Server[object, object]` | now `Server[dict[str, Any], Any]` — the default lifespan yields an empty dict and nothing here reads a request payload. | py only |

The P1 comment's second half was the more useful one: `test_no_object_annotations.py` swept function signatures only, so a type argument inside a variable annotation was invisible to it. It now covers annotated assignments, which immediately found one more — `sort_helper.py`'s `seen: set[tuple[object, ...]]`, where `DedupKey` is right there as the return type of the function that fills it.

Every new test was checked against the unfixed source: 5 of 5 fail on Python, the same 5 of 5 fail on TypeScript. `packages/agents` had no tests for either module before this, so `file-version.test.ts` and `tool-operations.test.ts` are new.

### `Resource.resolve_glob` takes `list[PathSpec]`

Separately, and prompted by review of the annotations this PR added: the base declared `list[str | PathSpec]`. That union is the **argv** type — a command line really does mix text words and paths, and `classify_parts` tags them per position — but it had leaked one layer past the split, into a function that only ever receives paths. TypeScript's twin is already `glob?(paths: readonly PathSpec[], ...)`, all 39 implementations take `list[PathSpec]`, and all three call sites pass one.

Nothing caught the mismatch because `make_resolve_glob` returned `Callable[..., Any]`. It now returns a `ResolveGlobFn` protocol, and `ResolveGlobOp` (the consumer-side twin in the generic_bind adapter) drops the same union. Re-widening any override now fails twice — once on Liskov against the base, once at the call into the factory. Verified by doing it.

Not touched: `_as_spec` still tolerates a `str` coming *back* from `resolve_glob`. That is a deliberate, freshly-documented allowance for a public hook, not the same leak.

### CI failure from the first push

`tests/cli/test_mcp.py::test_mcp_help_describes_stdio` asserted on rendered `--help` text. That box is laid out to the terminal's width, so the assertion passed locally and the flag pair wrapped away on CI. It reads the declarations instead, and now passes at `COLUMNS` 40/80/120/200.

## Verification

- `mypy` — Success, 1843 source files (post-merge)
- `pytest` — exit 0 (`tests/fuse` excluded locally: no macFUSE on this machine; CI runs it)
- `pre-commit run --all-files` — all hooks pass (needs Node ≥ 22.12 and `uv ≥ 0.9`: an older Node crashes `knip` and the `pi` suite inside `undici`, and an older `uv` rewrites the whole lockfile)
- gates: layout parity, spec parity, docs frontmatter, barrel surface, case targets — all exit 0
- `pnpm -r build`, `pnpm -r typecheck` — green; TS drift vs `upstream/main` is exactly the two agents fixes plus their new tests
- `pnpm --filter @struktoai/mirage-agents test` — 19 files, 189 tests, all pass
- new gate proven to fire: reintroducing a violation fails it, restoring it passes

`spec/layout_exceptions.json` baseline drops 299 → 296: the new modules are Python twins of TS modules that had none, so the ratchet required locking the improvement in.


zechengz:
@codex review

PR #829, opened 2026-08-16
fix: catch OSError when mfusepy cannot find libfuse

Jackpkn:
## Problem

`mfusepy` raises `OSError("Unable to find libfuse")` **at import time** when the package is installed but no libfuse driver is present on the system. The three guards around it caught only `ImportError`:

```python
try:
    import mfusepy as fuse
except ImportError:      # mfusepy raises OSError, not ImportError
    fuse = None
```

Because `mirage/__init__.py` reaches `fuse/mount.py` eagerly (`workspace` → `runner` → `kernel_mounts` → `workspace/fuse.py`), the result is not a degraded FUSE path — `import mirage` fails outright and the **entire pytest suite fails to collect**.

This means the two commands `CONTRIBUTING.md` gives contributors are unusable on any machine without macFUSE/fuse3/WinFsp:

```bash
uv sync --all-extras   # installs mfusepy via the [fuse] extra
uv run pytest          # OSError: Unable to find libfuse, 0 tests collected
```

Installing the extra is what breaks it: without `[fuse]` the `ImportError` guard works fine, so this only bites contributors who follow the documented setup.

## Fix

Widen the three guards to `except (ImportError, OSError)`. The degraded path they already implement is unchanged and correct — `fs.py` swaps in `object` as the base class, and `MirageFS.__init__` raises a clear `RuntimeError` naming the extra and the OS drivers.

`tests/fuse/test_darwin.py` imported `mfusepy` unguarded at module level and broke collection for the same reason. It now guards the import and skips only the three tests that actually touch `mfusepy`; the rest exercise pure-ctypes helpers (`changes_from_setattr`, `rename_flags_check`, `timespec_to_float`) and run anywhere.

## Python/TypeScript parity

No matching TypeScript change is needed. `packages/node` loads `@zkochan/fuse-native` through a **dynamic** `import()` inside `mount()` (`fuse/mount.ts:114`), so importing the package never touches it. Verified on a machine with no macFUSE: `import('@struktoai/mirage-node')` succeeds. The eager Python import is the divergence.

## Test plan

- [x] New regression test `test_import_survives_missing_libfuse` installs a `sys.meta_path` finder that raises `OSError` for `mfusepy`, then asserts in a subprocess that `import mirage` succeeds and all three modules degraded to `None`.
- [x] Verified the test is meaningful: it **fails** with the source fix reverted (`OSError: Unable to find libfuse`) and passes with it.
- [x] `uv run pytest --ignore=tests/fuse` → 15,244 passed, 0 failed.
- [x] `tests/fuse/test_darwin.py` now skips 3 tests instead of failing collection.
- [x] `pre-commit run` clean on all 5 changed files (yapf, isort, ruff, mypy).

Remaining `tests/fuse` failures on a driverless machine are pre-existing and out of scope: 89 tests construct `MirageFS`, which legitimately requires libfuse. Whether those should carry module-level skips is a separate call I left to maintainers.


zechengz:
@codex review

zechengz (approved):
Thanks for the contribution.

PR #830, opened 2026-08-16
fix(cache): make Redis file cache add atomic

### LINKED ISSUE #438: RedisFileCacheStore.add() check-and-set is not atomic; a late drain can clobber a fresher cache entry
## Problem

`RedisFileCacheStore.add()` (python `mirage/cache/file/redis.py`, ts `packages/node/src/cache/redis/file.ts`) implements its insert-only contract as two separate Redis commands: `EXISTS data:<key>`, then `SET` if absent. The check and the insert are not atomic, so the contract ("never clobber a live entry") is best-effort.

`add()`'s main caller is the background drain (`apply_io`), whose whole reason for using `add` instead of `set` is that a drain can finish late, after someone else already cached fresher bytes for the path.

## Failure interleaving

```
t0  drain A finishes draining v1 bytes (stream opened before the object changed)
t1  A: EXISTS data:/s3/f.txt -> 0
t2  B: set('/s3/f.txt', v2)              # fresh refetch caches v2
t3  A: SET data:/s3/f.txt v1             # v1 overwrites v2
```

End state: stale v1 bytes and v1 fingerprint in the cache. The RAM store cannot produce this ending: its per-key asyncio lock serializes check-and-insert, so the final state is always v2 regardless of ordering. Redis currently permits an end state RAM makes impossible.

## Severity

Mostly benign:
- within one process, `apply_io`'s `_drain_tasks` map already dedups drains per path, and the race window is one Redis round trip;
- if the remote content did not change, v1 == v2 and the overwrite is invisible;
- under `ConsistencyPolicy.ALWAYS` the stale fingerprint fails the next freshness check and the entry is evicted and refetched (self-healing).

It becomes real when the Redis cache is shared across processes (the per-process drain dedup does not apply) and consistency is LAZY (no freshness check ever runs): the stale entry is then served indefinitely, until an unrelated write invalidates the path. Low probability per read, silent and persistent when it lands.

## Proposed fix

Make the check-and-set atomic server-side: `SET data:<key> <bytes> NX`, and write the `meta:<key>` fingerprint only if the NX succeeded (small pipeline or Lua if we want both keys in one round trip). Either ordering then ends with the fresher bytes, restoring the same invariant the RAM store's per-key lock provides. Mirror in the TS node store.

Context: Redis drain support (the main `add()` caller) was added in the RAM/Redis cache alignment work following PR #434.

Beverly621:
## Summary

- make `RedisFileCacheStore.add()` atomic in Python and TypeScript with one Redis Lua execution
- write the data and fingerprint keys together and apply matching TTLs only when the insert wins
- add concurrent, binary-payload, fingerprint-consistency, and TTL regression coverage in both implementations

## Root cause

`add()` previously issued `EXISTS` and `SET` as separate Redis operations. A late background drain could observe an empty key, pause, and then overwrite fresher bytes inserted by another process. This broke the insert-only contract when Redis was shared across processes.

The Lua script keeps the existence check, both writes, and both expirations in one atomic Redis execution, so exactly one concurrent caller wins and its data/fingerprint pair remains consistent.

Closes #438.

## Validation

- Python Redis regression suite on Linux/Redis 7: 16 passed
- Python cache/file suite on Linux/Redis 7: 97 passed
- TypeScript Redis regression suite on Redis 7: 14 passed
- Node package TypeScript typecheck
- Prettier, ESLint, and Knip
- Python format, import, Flake8, Ruff, and PathSpec checks
- mypy: changed Python source passes
- `git diff --check`

## AI assistance

Implementation and test preparation were assisted by OpenAI Codex using GPT-5.6Sol. The resulting changes and validation output were reviewed before submission.

### DIFF
--- a/python/mirage/cache/file/redis.py
+++ b/python/mirage/cache/file/redis.py
@@ -13,13 +13,26 @@
 # ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
 
 import asyncio
-from collections.abc import Iterable
-from typing import Any
+from collections.abc import Awaitable, Iterable
+from typing import Any, cast
 
 from mirage.cache.file.mixin import FileCacheMixin, validate_max_drain_bytes
 from mirage.cache.file.utils import (default_fingerprint, glob_escape,
                                      parse_limit)
 from mirage.resource.redis.redis import RedisResource
+
+ADD_SCRIPT = """
+if redis.call('EXISTS', KEYS[1]) ~= 0 then
+    return 0
+end
+redis.call('SET', KEYS[1], ARGV[1])
+redis.call('SET', KEYS[2], ARGV[2])
+if ARGV[3] ~= '' then
+    redis.call('EXPIRE', KEYS[1], ARGV[3])
+    redis.call('EXPIRE', KEYS[2], ARGV[3])
+end
+return 1
+"""
 
 
 class RedisFileCacheStore(RedisResource, FileCacheMixin):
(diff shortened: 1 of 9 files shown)


zechengz:
@codex review

python/mirage/cache/file/redis.py
```diff
@@ -13,14 +13,27 @@
...
-from typing import Any
+from collections.abc import Awaitable, Iterable
+from typing import Any, cast
 
 from mirage.cache.file.mixin import FileCacheMixin, validate_max_drain_bytes
 from mirage.cache.file.utils import (default_fingerprint, glob_escape,
                                      parse_limit)
 from mirage.resource.redis.redis import RedisResource
 
+ADD_SCRIPT = """
```
zechengz: It will be appreciated if we can use a script such as python/mirage/workspace/session/cas.lua?

zechengz (approved):
Modified. Thanks for the contribution. Feel free to join our slack channel if you want to make some future contributions https://discord.com/invite/u8BPQ65KsS :)

Beverly621:
Thank you for taking the time to review and improve the PR. I've joined the Discord server and look forward to staying involved, learning more about the project, and contributing again in the future～🥳🤩

PR #837, opened 2026-08-17
refactor(layout): items 33+35 — module homes and workspace/config placement (34 already closed)

bytecii:
Block E items **33** (T3-3 module homes) and **35** (T3-2 workspace/config placement) from the cleanup plan. Item **34** (T3-1 runtime layout) is in the title because it was in scope — it turned out to be already closed upstream, and the verification is recorded below rather than a diff.

No behavior changes except the three noted under "Bugs found on the way". Everything else is a module moving to the home its twin already occupies.

The measure is `scripts/check_layout_parity.py`, the item-30 gate: **286 → 269 unexcused divergences**, baseline lowered to lock it in. Every one of the 17 closed lines is listed below, so the diff can be reviewed against data rather than taste.

## Item 33 — module homes

| Move | Why that home | Gate |
|---|---|---|
| py `utils/fingerprint.py` → `watch/fingerprint.py` | all six consumers are watch machinery (`core/{s3,ssh,dropbox,opendal}/watch.py`, `core/disk/watch/walk.py`, `watch/walk.py`); TS is already at `watch/fingerprint.ts` | −2 |
| py `workspace/provision/rollup.py` → `provision/rollup.py` | imports nothing from `workspace`; TS is at the leaf `provision/rollup.ts`. Now reads `provision/types.py` directly, as `rollup.ts` reads `./types.ts` | −2 |
| py `workspace/provision/builtins.py` deleted | a 7-line module for one function with one caller; TS keeps `handleBuiltinProvision` module-level inside `provision_node.ts`, so python does too | −1 |
| py 9 snapshot key enums → new `workspace/snapshot/keys.py` | mirrors `core/src/workspace/snapshot/keys.ts`, the module python lacked | −1 |
| py `IndexType` → `cache/index/config.py`, `CacheType` → `cache/file/config.py` | their TS homes (`cache/index/config.ts:29`, `cache/file/config.ts:15`) | 0 |
| ts `node/cache/redis/file.ts` → `node/cache/file/redis.ts` | python is `cache/file/redis.py`; the family dir is the kind, the module is the backend, as everywhere else in `cache/` | −2 |
| ts new `core/src/concurrency/limiter.ts` | twin of `concurrency/limiter.py`; replaces the private `Semaphore` in `accessor/dify.ts` | −1 |

`NodeMetaKey` deliberately did **not** go to `keys.py`. It names `NodeMeta`'s own fields, its only consumer is `mount/namespace/namespace.py`, and importing the snapshot package from there closes a real cycle (`namespace → snapshot.keys → snapshot/__init__ → api → state → namespace`, ImportError on a partially initialized module). Snapshot serializes the namespace, not the reverse, so the key sits beside `NodeMeta`.

New test `tests/workspace/snapshot/test_keys.py` diffs the five shared tables against `keys.ts` and asserts every value is its member name lowercased. Verified it discriminates: renaming `StateKey.CLIS`'s value fails both assertions.

## Item 34 — already closed, no diff

The plan asked for `runtime_config.ts`, a `HomeConfig` twin, and deleting `RuntimeTable.candidates()`. All three are done on `main`:

- `core/src/runtime/` mirrors `mirage/runtime/` module-for-module (`base`, `config`, `errors`, `handles/`, `js/`, `language`, `mixin`, `policy/`, `python/`, `resolver`, `table`, `types`, `vfs`). The `runtime_*`-prefixed siblings under `workspace/executor/` the plan describes are gone.
- `runtime/config.ts:38` exports `HomeConfig`, `:42` `HOME_CONFIG_KEYS`, `:55` `coerceRuntimeConfig`.
- `candidates` has no hit anywhere in the TS tree as a `RuntimeTable` member.

## Item 35 — workspace/config placement

| Move | Why that home | Gate |
|---|---|---|
| ts `core/workspace/workspace.ts` → `core/workspace/workspace/workspace.ts` | beside its own helper package, as `workspace/workspace/workspace.py` is (139 importers rewritten) | −2 |
| ts `node/workspace/mount_spec.ts` → `core/workspace/mount/spec.ts` | zero node APIs; twin of `workspace/mount/spec.py` | −2 |
| ts new `core/workspace/workspace/guard.ts` | twin of `workspace/workspace/guard.py`; three call sites moved onto it | −1 |
| ts new `node/workspace/workspace/kernel_mounts.ts` | twin of `workspace/workspace/kernel_mounts.py`; `NodeWorkspace` 211 → 169 lines, now delegations | −1 |
| ts `server/src/config.ts` → `node/src/config.ts` | python's YAML workspace-config loader is the top-level `mirage/config.py`; `yaml` moved with it, and the config API is exported from the node barrel so a `mirage-node` consumer can finally load a YAML | −2 |

`rejectConfigScript` "returning zero hits" was literally true but substantively wrong: TS already had the guard as `scriptStringError` in `runtime/base.ts`, called at the same three sites, each doing its own `typeof === 'string'` test. `guard.ts` gives it python's shape and home and drops the three copies; the message stays in `runtime/base.ts`, which raises it for a RuntimeEntry built with a string script and must not reach up into the workspace layer for its wording.

## Bugs found on the way

1. **The dify accessor's semaphore could exceed its capacity.** `release()` returned the permit to the pool *before* waking the queued waiter, so a caller arriving in the same tick took it while the woken waiter also decremented — two holders at capacity 1, and a count that drifts negative under load. `ConcurrencyLimiter.release` hands the permit straight to the waiter. Pinned by `a caller arriving during a release cannot barge past the queue`, verified to fail against the pool-first version.
2. **Double release inflated the count.** The old release function had no guard. The new one is idempotent, matching what `async with` gives python for free. Pinned, verified to fail without the guard.
3. **`maxConcurrency < 1` was silently clamped** (`Math.max(1, max)`) where python raises. Both reject now. Unreachable for real configs — `PositiveInt` and `normalizePositive` already gate it — so no behavior change in practice.

## Plan premises that were stale

Recorded so the plan doc can be corrected:

- **Item 34** entirely (above).
- Item 35's `cache.ts` "missing from TS `workspace/workspace/`" — shipped in #751.
- Item 35's "python lacks config auto-discovery" — `server/workspace_config.py` has `WORKSPACE_CONFIG_CANDIDATES` and `resolve_workspace_config`, shipped in #825.
- Item 35 says to move `workspace_config.ts` to node alongside `config.ts`. That would **create** a divergence: python has `server/workspace_config.py`, so the pair already matches. Left in place.
- Item 36's one behavioral clause ("`qdrant/search.py` raises `UsageError` where `lancedb/search.py` raises `ValueError`") is already closed — both raise `ValueError`. Item 36 is now pure DRY.

## Deliberately deferred

- **onedrive/sharepoint 18-module collapse** (item 33's largest clause). The plan already notes those python modules are not pure delegation — `onedrive/mkdir.py` owns the `-p` parent loop and the invalidation calls. That is a merge of behavior, not a move, and does not belong in a layout PR.
- **Widening `browser/src/workspace.ts` to accept a per-mount mode** (item 35's last clause). That is a feature.

## Rebase onto main, and the three CI reds it explains

Rebased onto `c35fb1cf0`. Two of the moves collided with work that landed after the original base — both are semantic, not textual, which is why the first CI run went red on a branch that was locally green:

- **#830 added `node/src/cache/redis/add.lua`** into the directory this PR empties. The script is read relatively by its only consumer (`readFileSync(new URL('./add.lua', import.meta.url))`) and copied into `dist` by an explicit tsup path, so it moves with the file: now `cache/file/add.lua`, and `tsup.config.ts` copies from `src/cache/file/`. This was the one textual conflict, and git's rename detection suggested exactly this resolution.
- **#833 added `core/src/workspace/shell_bash_builtins.test.ts`**, which imports `./workspace.ts` — the path this PR moves to `./workspace/workspace.ts`. That single line is the whole of the `test`, `pre-commit` (eslint + knip) and `test-typescript-gate` failures. I swept the tree for every other reference to the four moved paths; the remaining `./workspace.ts` hits are `packages/node` and `packages/browser`, whose own `workspace.ts` this PR never touched, plus the three tests already inside `workspace/workspace/` that resolve correctly.

The fourth red, `audit`, is unrelated to the moves and is a latent bug in the job I added in #831. It runs `pnpm audit` and never installs, so the pnpm store does not exist; `cache: pnpm` then fails post-run path validation on every cache **miss** and silently skips the save on a **hit** — which is why it is green on main and red here. Dropped the cache from that job, since it has nothing to cache. It is `continue-on-error: true` and outside the gate, so it never blocked a merge either way.

## Verification

- `scripts/check_layout_parity.py --strict` — 269, baseline lowered from 286
- `scripts/check_barrel_surface.py` — 77 names, all consumed
- `scripts/check_spec_parity.py` — 93 commands match
- `pnpm -r build`, `pnpm -r typecheck`, `pnpm run lint`, `pnpm run format:check`, `knip` — clean
- `pnpm -r test` — 11,943 passed, 0 failed
- `pytest` (minus `tests/fuse`, no macFUSE on this host) — green
- `pre-commit run --all-files` — clean
- integ battery, both languages — see the checks


zechengz:
@codex review

PR #838, opened 2026-08-17
refactor(watch): let the cache manager walk the ancestor chain

zechengz:
The watcher built its own ancestor list and evicted each entry with `invalidate_after_write`, which is what `CacheManager.invalidate_ancestors` already does.

Probed before changing anything: the two produce the identical set of dropped directory keys.

```
watcher loop   dirs: ['/m', '/m/', '/m/data', '/m/data/', '/m/data/sub', '/m/data/sub/']
invalidate_anc dirs: ['/m', '/m/', '/m/data', '/m/data/', '/m/data/sub', '/m/data/sub/']
```

So the watcher's copy added a second stopping rule to keep in step, plus one file-cache `remove` per ancestor *directory* key, which can never hit.

- `invalidate_ancestors` / `invalidateAncestors` joins the `CacheInvalidator` protocol in both languages.
- `Watcher._ancestors` / `Watcher.ancestors` deleted.
- The reach is now pinned against a real `CacheManager` (seeded `RAMIndexCacheStore`, assert every listing up to the mount root is gone) rather than only through a fake, since the walk lives there now. Confirmed non-vacuous by removing the call: 6 of 9 Python and 5 of 8 TypeScript watcher tests go red.

## Gates

| | |
| --- | --- |
| pre-commit (all files) | exit 0, stable on re-run |
| layout parity | 286 = baseline |
| Python suite | exit 0, 0 failures |
| `pnpm -r typecheck` | all 11 packages |
| TS core | 8912 passed, 10 skipped |
| TS node | 2448 passed, 4 skipped |
| TS browser | 184 passed |

### DIFF
--- a/python/mirage/watch/base.py
+++ b/python/mirage/watch/base.py
@@ -34,6 +34,9 @@ class CacheInvalidator(Protocol):
     async def invalidate_subtree(self, path: PathSpec) -> None:
         ...
 
+    async def invalidate_ancestors(self, path: PathSpec) -> None:
+        ...
+
 
 class WatchMount(Protocol):
     """What the watch runtime needs from one mount entry.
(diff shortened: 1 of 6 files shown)


zechengz:
@codex review

PR #840, opened 2026-08-17
fix(s3): refuse a version-pinned read on the browser presigned client

bytecii:
The presigned-fetch S3 client's `Get` case read only `Key` and `Range`, so a `VersionId` was **silently dropped** and the request returned whichever version was current. No error, no log — just the wrong bytes.

That is reachable today: `core/s3/read.ts:65` and `core/s3/stream.ts:50` both set `input.VersionId = pinnedRevision` when `revisionFor()` returns a pin, so every snapshot-pinned browser read resolved against the live object instead. It affects all **15** browser resources that route through `createBrowserS3Client` — s3, r2, gcs, minio, ceph, backblaze, wasabi, digitalocean, scaleway, supabase, aliyun, tencent, oci, qingstor, seaweedfs. The node/SDK path is correct, so it never reproduces outside a browser runtime.

## The fix

Refuse, exactly as #742 did for `Range`. A presigned URL is signed for a key, not a version: the provider is `(path, operation, options)` and has no slot to carry one, so there is nothing to pass through and no way to answer honestly.

`Head` is guarded through the same helper. `stat` does not pin a revision today (`stat.ts:101` sends a bare `Bucket`/`Key`), but a version-pinned stat would be wrong in precisely the same way, and the shim is the chokepoint where that has to be caught.

## The rule this does *not* over-apply

Only fields that change the **answer** may refuse. `stat`'s directory probe sends `MaxKeys: 1` (`stat.ts:88-92`) and the shim ignores it — ignoring it returns *more* data but the *same* answer, and refusing it would break `stat` outright. A test pins that counter-example so a later reading of "never accept-and-ignore" does not turn every ignored hint into an error.

## Behavior change

Browser snapshot restores that were silently returning current-version bytes now fail loudly. That is the point of the change, but it is a visible difference for anyone who had a pinned read quietly "working".

There is no python twin: python has no browser runtime and no presigned shim.

## Verification

- `client_browser.test.ts` — 22 passed; whole `core/s3` suite 67 passed across 8 files
- `tsc --noEmit`, eslint, prettier — clean
- **Discriminating check**: with the two `refuseVersionPin` calls commented out, exactly the 2 new refusal tests fail and the MaxKeys counter-example still passes. The tests assert that neither the provider nor `fetch` runs, which is what separates *refusing* from *answering* — the pre-fix path reached both and resolved.

### DIFF
--- a/typescript/packages/core/src/core/s3/client_browser.test.ts
+++ b/typescript/packages/core/src/core/s3/client_browser.test.ts
@@ -169,6 +169,56 @@ describe('createBrowserS3Client — command dispatch', () => {
     ).rejects.toMatchObject({ $metadata: { httpStatusCode: 404 } })
   })
 
+  // A presigned URL is signed for a key, not a version. Before this refusal the
+  // shim read only Key, so a snapshot-pinned read fetched and returned whichever
+  // version was current -- no error, no log, just the wrong bytes. Asserting
+  // that neither the provider nor fetch ran is what distinguishes refusing from
+  // answering: the old code reached both and resolved.
+  it('GET refuses a version-pinned read instead of returning the current object', async () => {
+    const { provider, calls } = makeProvider(() => 'https://signed/get')
+    fetchHandle = installFetch(() => new Response(new Uint8Array([104, 105]), { status: 200 }))
+    const client = createBrowserS3Client(cfg(provider))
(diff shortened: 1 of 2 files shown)


zechengz:
@codex review

PR #845, opened 2026-08-17
fix(deps): override deepmerge-ts to >=8 for GHSA-ggr8-5vv4-36mx

zechengz:
## Summary

- `@prisma/config` pins `deepmerge-ts` to exactly `7.1.5`, which is vulnerable to stack exhaustion on recursive object graphs (GHSA-ggr8-5vv4-36mx, high).
- Latest prisma (7.9.1) pins the same exact version, so a prisma bump does not fix it. Added a `pnpm.overrides` entry instead, matching the existing entries.
- v8 is a superset of v7's exports; `@prisma/config`'s only use is `deepmerge` as c12's `merger`.

## Test plan

- `pnpm audit` -> No known vulnerabilities found (was 1 high)
- `pnpm install --frozen-lockfile` -> exit 0
- `prisma generate --schema prisma/schema.prisma` (the integ CI step) -> exit 0

### DIFF
--- a/typescript/package.json
+++ b/typescript/package.json
@@ -48,6 +48,7 @@
       "nanoid@>=3.0.0 <3.3.18": ">=3.3.18 <4",
       "find-my-way@<=9.6.0": ">=9.7.0 <10",
       "postcss@<=8.5.17": ">=8.5.18 <9",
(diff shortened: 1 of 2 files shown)


zechengz:
@codex review

PR #847, opened 2026-08-18
fix(ls): list a nested mount's root in its parent's -R listing

bytecii:
`ls -R` on a directory holding a nested mount omitted the mount's own row while still printing its group:

```
$ ls -R /base          # mounts: /base/, /base/nested/
/base:
top.txt                # <- `nested` missing

/base/nested:
real.txt
```

GNU lists a mountpoint as an ordinary directory entry of its parent, `-R` or not. Pinned on coreutils 9.7 over a real tmpfs in `debian:stable-slim`:

```
base:
nested
top.txt

base/nested:
real.txt
```

Non-recursive `ls /base` was always right, which is what makes this specifically the recursive case.

## Root cause

The generic conflated two different things: **listing** a mount root as a row, and **descending** into it. `-R` withheld the whole `child_mounts` merge on a backend-served listing, on the theory that the cross-mount fan-out contributed the nested mount whole. It contributes the *group*, not the parent's *row*, so the row survived only by accident — whenever the parent's own backend happened to hold a key of that name (which is why the existing shadowed-mount fixture never caught it).

Splitting the two:

- **The merge is unconditional.** A mountpoint is an ordinary entry of its parent.
- **Descent stops at a mount root**, read from the mount table (`MountView.is_root`) instead of inferred from `child_mounts` being empty. A namespace-only directory *above* a mount root is still descended, because nothing else renders it.

## Three more shapes, same cause

- `ls -R /` dropped `ghost` **and** both the `/ghost:` and `/ghost/very:` groups whenever a backend served `/` (mount at `/ghost/very/deep`).
- `/base/sub`'s group dropped a mount at `/base/sub/deep` — the merge is per directory listed, not per operand.
- A relative operand **leaked shadowed content**: `ls -R base` printed `base/inner:` twice, once with the parent's hidden `leftover.txt` and once with the mount's own listing, because `_drop_shadowed_ls_groups` only recognizes an absolute header. That group is now never produced rather than produced and filtered.

## A mountpoint is not always a directory

Follow-up from review. The merged row was synthesized as `FileType.DIRECTORY`, but a mount root can serve a single file — every workspace mounts `/.bash_history` that way — so `ls -aF /` printed `.bash_history/` and `ls -l` rendered it `drwxr-xr-x`.

No backend can answer that question: the parent's cannot see into the child mount, and the child's own names its root `/`. So the row is stat'd through the dispatcher (`opts.stat_path` / `opts.statPath`), which routes to whichever mount owns the path, and is renamed the way the relay renames one. Directory stays the fallback for a caller with no dispatcher — the only thing that absence can mean.

The same fact settles what `-R` descends: the `ls` fan-out no longer emits a block for a mount that is not a directory, which otherwise printed the name a second time as its own group. Only a **confirmed** non-directory is dropped, so a root the dispatcher cannot stat keeps its block rather than vanishing on a failed probe.

GNU is pinned the same way, over a `mount --bind` of one file onto another: one ordinary row in the parent, no group of its own.

```
$ ls -aRF /
/:
.bash_history          # <- no '/', and no `/.bash_history:` group below
dev/
top.txt

/dev:
null
zero
```

## The contract, and why the relay differs

`ls` takes `mounts` beside `links` and `child_mounts`, and it names the boundaries **this walk's readdir cannot cross**, not the boundaries that exist. That distinction is the whole contract, because the two entry points differ:

| entry point | readdir | mount root under `-R` |
| --- | --- | --- |
| `ls_generic` | bound to one backend | stop — the fan-out renders the group |
| cross-mount relay (#844) | routes per path | descend — nothing else will |

Python already said this by omission, calling the generic with explicit keywords. The TypeScript relay hands the generic a whole `CommandOpts`, so it now builds one carrying every name-plane fact **except** the boundaries. Without that, `ls -R /base /other` dropped `/base/inner:` entirely — caught by the existing `fanout.test.ts` spanning-mounts case, and now pinned directly in both relay suites too.

## Tests

- 4 end-to-end cases in `test_fanout.py` and its TS twin, one per shape above. All four fail on the pre-fix source (verified by reverting).
- 1 generic-level unit test per side: a mount root is listed but not descended out of a backend-served directory.
- 1 relay case per side: handed a **full** namespace, the relay still descends the mount root. Both existing relay suites passed while broken because they defaulted `ns` to none.
- 3 more per side for the row's type: a mount root is listed but not descended; a mount serving one file is not a directory row; the row falls back to directory with no dispatcher wired. Plus one end-to-end case pinning `ls -aRF /` byte-for-byte, which is the one that fails on both counts before the fix (`.bash_history/` in the parent, then a `/.bash_history:` group).
- Two unit tests that passed a bare `child_mounts` now supply a mount table, since the fact moved to its real source.
- `integ/crossmount/nested/basic.json` seq 950006 **pinned the bug**; it now pins GNU's answer, with a `note` recording the tmpfs pin.

## Verification

Both hosts, both languages, byte-identical output on every probe.

- Full Python suite green (`--ignore=tests/fuse`; `test_darwin.py` needs mfusepy, absent on this machine).
- Full TS suite green: core 674 files / 8990 tests, plus browser, node, server, dsh, agents, cli, opencode.
- `pnpm -r typecheck`, eslint, prettier, `pre-commit --all-files`.
- Integ battery on `ram`, `ram-root`, `ram-alias`, `ram-nested`, `ram-history`, `disk`, `statvfs`, `git-disk`, `git-ram` — **5871 passed, 0 failed on each host**.

## Out of scope

Group *ordering* across mounts is still the fan-out's concatenation rather than GNU's pre-order DFS: `ls -R /` emits `/`, `/ghost`, `/ghost/very`, then `/dev`, `/ghost/very/deep`. That is the fan-out's shape, not this row's.

### DIFF
--- a/integ/crossmount/nested/basic.json
+++ b/integ/crossmount/nested/basic.json
@@ -67,6 +67,7 @@
     },
     {
       "id": "nest_ls_recursive_renders_both_groups",
+      "note": "A mount root is an ordinary directory entry of its parent, so `inner` is a row of `/data` as well as a group of its own. Pinned on coreutils 9.7 over a tmpfs mounted at `base/nested`: `ls -R base` prints `nested` then `top.txt`, then the mounted directory under its own header. `-R` used to withhold the namespace merge and leave the whole nested mount to the cross-mount fan-out, which contributes the group but not the parent's row, so the row went missing here where `/data`'s backend holds no `inner` key. Non-recursive `ls /data` (seq 950002) always listed it.",
       "seq": 950006,
       "targets": [
         "ram-nested"
@@ -74,7 +75,7 @@
       "command": "ls -R /data",
       "expect": {
         "exit": 0,
-        "stdout": "/data:\ntop.txt\n\n/data/inner:\nleaf.txt\n",
+        "stdout": "/data:\ninner\ntop.txt\n\n/data/inner:\nleaf.txt\n",
         "stderr": ""
       }
     },
--- a/python/mirage/commands/builtin/generic/crossmount/relay/ls.py
+++ b/python/mirage/commands/builtin/generic/crossmount/relay/ls.py
@@ -89,7 +89,12 @@ async def run_ls(scopes: list[PathSpec], flag_kwargs: dict[str, FlagValue],
         ns (NamespaceView | None): Name-plane facts no backend can see.
             The attr overlay matters most here: without it a relayed row
             would report the raw backend mode and silently lose a chmod
-            the namespace holds.
+            the namespace holds. Every fact but ``mounts``, which names
+            the boundaries a walk's readdir cannot cross: this one's
+            does cross them, so the relay descends a nested mount
+            itself, and it has to, because nothing runs behind a relay
+            to contribute that group the way the fan-out does for a
+            single-mount run.
     """
     p = functools.partial
     stat_fn: Stat = p(relayed_stat, dispatch)
--- a/python/mirage/commands/builtin/generic/ls.py
+++ b/python/mirage/commands/builtin/generic/ls.py
@@ -10,7 +10,7 @@ from mirage.commands.config import CommandOpts
 from mirage.commands.spec import SPECS
 from mirage.commands.spec.types import FlagValue, FlagView
 from mirage.io.types import IOResult
-from mirage.ops.types import ChildMounts, LinkView
+from mirage.ops.types import ChildMounts, LinkView, MountView, StatPath
 from mirage.types import FileStat, FileType, LsSortBy, PathSpec
 from mirage.utils.errors import fs_strerror
 from mirage.utils.key_prefix import rekey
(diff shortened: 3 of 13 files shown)


zechengz:
@codex review

bytecii:
> **P2 — Preserve mounted files' types in recursive listings.** When a child mount is a file rather than a directory, the unconditional synthesis misclassifies it. [...] every workspace mounts the file view `/.bash_history`, so `ls -aR /` now emits a synthetic `.bash_history` directory row in the parent block and then the fan-out emits `/.bash_history` again from the history mount; with `-F`, the first row is also incorrectly suffixed with `/`.

Correct on both counts, and reproduced exactly as described. Fixed in `88ef9901`.

**The row is stat'd, not synthesized.** No backend can answer this — the parent's cannot see into the child mount, and the child's own names its root `/` — so `_mount_row` / `mountRow` asks the dispatcher through `opts.stat_path` / `opts.statPath`, which routes to whichever mount owns the path, and renames the row the way the relay renames one. `FileType.DIRECTORY` stays only as the fallback for a caller with no dispatcher wired, which is the one thing that absence can mean.

**The duplicate group is the same fact, one layer up.** `_ls_block_mounts` / `lsBlockMounts` drops a non-directory descendant before the `ls` fan-out runs a sub-run for it. Only a *confirmed* non-directory is dropped: a root the dispatcher cannot stat keeps its block rather than vanishing on a failed probe.

GNU is pinned the same way, over a `mount --bind` of one file onto another — one ordinary row in the parent, no block of its own:

```
$ ls -aRF /
/:
.bash_history
dev/
top.txt

/dev:
null
zero
```

That exact string is now asserted end-to-end in `test_fanout.py` and its TS twin, byte-identical on both hosts, plus three generic-level cases per side (listed-not-descended, a one-file mount is not a directory row, and the no-dispatcher fallback).

Re-verified after the fix: full Python suite, full TS suite, `pnpm -r typecheck`, eslint, prettier, `pre-commit --all-files`, and the integ battery on `ram`, `ram-root`, `ram-alias`, `ram-nested`, `ram-history`, `disk`, `statvfs`, `git-disk`, `git-ram` — **5871 passed, 0 failed on each host**.

<sub>Unrelated to this PR: the red `cli-gate` on the previous run is a rollup of a `cancelled` job. "Cross-language snapshot interop" stalled inside `sudo apt-get update` in its **first** step (last log line 04:10:38, cancelled at the 30-minute timeout 04:39:51), before any of this branch's code ran. The push above re-triggers it.</sub>

PR #851, opened 2026-08-18
fix(search): gate the grep/rg push-down on operand count

bytecii:
## The bug

A search push-down asks the backend one whole-container question and prints its entire answer, so it can only stand in for a line naming exactly one operand. Given two it answered for the first and dropped the rest **in silence**:

```
$ rg checkout-flow /lf/traces /lf/sessions
traces/trace-alpha.json:{"id":"trace-alpha", ...}      # sessions never searched, nothing said so
```

## The fix, and the alternative that was rejected

Running the push-down once per operand is not the fix. Several scopes map to the same container search — langfuse routes both `sessions` and one `session` to "search every session" — so two operands in one family would print that container twice. A multi-operand line therefore **defers to the generic scan**, which searches each operand in turn the way GNU does, and which every backend without a push-down already used.

The gate is written once rather than per wrapper:

| helper | family | backends |
|---|---|---|
| `pushdown_operand` / `pushdownOperand` | regex push-down | langfuse, mongodb |
| `literal_pushdown_operand` / `literalPushdownOperand` | LIKE/ILIKE substring | postgres |

Both return the sole concrete operand or `None`, folding operand count in beside the glob, pattern-list and shaping-flag rules that the six wrappers were each open-coding as a five-clause boolean. Twelve gates collapse to twelve calls, so they cannot drift apart again.

## Scope

**postgres was not in the original report** and has the identical bug through the same `search_pushdown_ok` gate. It is fixed here because it shares the helper.

**Five more backends are affected and deliberately left alone**, because each needs its own decision rather than this one:

- **slack** and **discord** already carry a `coalesce_scopes` mechanism, consulted only when `paths[0]` is non-native — there "coalesce every operand" is the likely answer, not "defer". Their TS sides never call `coalesceScopes` at all.
- **email**'s `rg` returns exit 1 (no match) instead of falling through, so a line the push-down cannot answer reports "nothing matched" rather than searching.
- **gmail** open-codes a partial shaping check and has no glob check at all.
- **email**'s `find` has the same `paths[0]` walk, and its TS twin has no push-down — a py/ts divergence to settle.

These are filed as follow-up work.

## Verification

Pinned **empirically on both hosts** against a real Langfuse v3 stack, MongoDB and Postgres — 13 new integ cases, each captured from actual output rather than predicted:

- `integ/resources/observability/langfuse/search.json` (seq 602107–602111)
- `integ/resources/{mongodb,postgres}/{grep,rg}.json`

**All 5 langfuse cases and 6 of the 8 database cases fail on the unfixed tree**, showing the exact drop (`got 'traces/trace-alpha.json\n'` where both operands were expected). The two `lone_operand` cases pass either way on purpose: they guard against over-deferring, proving the push-down still fires for a single operand.

Multi-operand shapes were checked against the real tools, not assumed — **GNU grep 3.11** (`debian:stable-slim`) and **ripgrep 14.1.1**:

- two directory operands without `-r` → two `Is a directory` lines and exit 2 (matches mirage exactly)
- a repeated operand is searched **twice**, not deduped (both tools agree)
- `rg` parallelizes and so emits operands in nondeterministic order, which is why those cases normalize with `sort -u`

Also fixes a **misdirected test**: `test_unresolved_glob_defers_to_generic` built its `PathSpec` without a `pattern`, so `has_unresolved_glob` returned `False` and the deferral it asserted actually came from `/traces/*` routing to `invalid`. It now uses `/sessions/*`, which routes to a searchable scope, so only the glob half of the gate can defer it.

### Checks

- `pre-commit run --all-files` — all hooks pass (converged; no yapf/isort oscillation)
- `pytest` (python, excl. `tests/fuse` — no macFUSE on this host) — exit 0
- `pnpm -r typecheck` — clean
- `@struktoai/mirage-core` vitest — 691 files, 9088 passed
- `check_case_targets.py --strict` — 2295, on baseline
- langfuse / mongodb / postgres batteries — python **154/111 passed**, typescript **154/111 passed**

### DIFF
--- a/integ/resources/mongodb/grep.json
+++ b/integ/resources/mongodb/grep.json
@@ -103,6 +103,32 @@
         "stdout": "2\n",
         "stderr": ""
       }
+    },
+    {
+      "id": "mdb_grep_two_collections_cover_both",
+      "seq": 620058,
+      "targets": [
+        "mongodb"
+      ],
+      "command": "grep ada /mongodb/mirage_integ/collections/books /mongodb/mirage_integ/collections/authors",
+      "expect": {
+        "exit": 2,
+        "stdout": "",
+        "stderr": "grep: /mongodb/mirage_integ/collections/books: Is a directory\ngrep: /mongodb/mirage_integ/collections/authors: Is a directory\n"
+      }
+    },
+    {
+      "id": "mdb_grep_r_two_collections_cover_both",
+      "seq": 620059,
+      "targets": [
+        "mongodb"
+      ],
+      "command": "grep -r ada /mongodb/mirage_integ/collections/books /mongodb/mirage_integ/collections/authors | cut -d: -f1 | sort -u",
+      "expect": {
+        "exit": 0,
+        "stdout": "/mongodb/mirage_integ/collections/authors/documents.jsonl\n/mongodb/mirage_integ/collections/books/documents.jsonl\n",
+        "stderr": ""
+      }
     }
   ]
 }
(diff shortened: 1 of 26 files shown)


zechengz:
@codex review

PR #856, opened 2026-08-18
fix(search): gate the remaining grep/rg/find push-downs on operand count

bytecii:
## What

The grep/rg search push-downs in **gmail, slack, discord and email** answered
from `paths[0]` and silently dropped every additional operand, plus **email
`find`**, which did the same thing through `extract_folder`. So:

```bash
grep -rw foo /slack/channels/alpha /slack/channels/beta
```

searched `alpha` only and printed the result as though it covered both. This
routes all five through the shared gate the companion PR (#851) added for
langfuse/mongodb/postgres — `pushdown_operand` / `pushdownOperand` — instead of
adding a second one.

Each backend got its own decision, because their gates differed in ways that
matter. They are written down below.

## The `honored` parameter

`has_search_shaping_flags` refuses a push-down whenever a flag would alter the
match set or the output shape. But it lists `w`, and gmail/slack/discord
*require* `-w` to push down at all — provider search is word-based, so `-w` is
precisely the flag that makes a verbatim provider answer faithful rather than
approximate. Rather than fork the predicate, it grew a `honored` sequence of
flag names the caller has already accounted for.

Two families use it differently, and the distinction sets each backend's list:

- **Verbatim-answer** (gmail, slack, discord — and langfuse/mongodb/postgres):
  the provider's search result is printed directly, so the only honored flag is
  the one that shapes the query itself, `w`.
- **Narrow-then-scan** (email): IMAP search only picks candidate messages;
  `grep_lines` then runs the real compiled pattern locally. So a flag may be
  honored whenever it *cannot make a message the search did not return
  contribute output* — `n`, `args_l`, `w`, `o`, `m`.

`-v` and `-c` are deliberately **not** honored for email even though the local
scan implements both: `-v` reports lines that do not match, which needs every
message rather than the candidates, and GNU `-c` prints a `path:0` row for
files with no match. Both need the full set, so both defer.

### gmail — `paths[0]` → gated operand

Was reading `paths[0]` for both the scope and the output `file_prefix`,
open-coding a partial shaping check, and had **no `has_unresolved_glob` check
at all**, so an unexpanded glob reached the native search. Now uses the shared
gate. The py/ts drift is closed by making `SEARCH_HONORED` / `SEARCH_MAX_RESULTS`
module constants that `rg` imports from `grep` on both sides, so the two commands
cannot disagree again. Dropping the open-coded check also means `-m` now defers
rather than being half-honored.

### slack — `coalesce_scopes` deleted, not promoted

This one already had a multi-operand mechanism, consulted only when
`detect_scope(paths[0])` was *not* native. The tempting fix is "always
coalesce". **It is unsound**, for three reasons:

1. `build_query` emits only `in:#channel`. `search_guild` takes no date. So
   folding `.../alpha/2026-01-08` into its channel **drops the date** and widens
   the search to the whole channel.
2. It widened a lone leaf operand the same way — a single `chat.jsonl` path was
   being answered with its channel's entire history.
3. It never actually fixed the cross-channel drop it looked like it addressed,
   because two different channels cannot fold into one `in:#` term.

Coalescing was producing a *wider* answer while looking like a narrower one, so
it is removed in both languages and multi-operand lines take the generic scan.

Also closed the real py/ts divergence: TS never called `coalesceScopes` at all
and was missing python's `target != "files"` and `search_available` terms.
`SlackTransport` gained an optional `searchAvailable?()` — the node transport
answers from its token (Slack rejects a bot token on `search.*`, so a doomed
call is worth not making), and the browser transport leaves it unimplemented,
which callers read as "try it".

### discord — same shape, same answer

`coalesce_scopes` was consulted only when `scope.level == "messages"`, so any
guild or channel directory operand was decided from `paths[0]` alone. Same
reasoning as slack, same removal.

### email grep/rg — the exit-1 bug

`email/rg.py` called `extract_folder(paths)` and, on a falsy folder,
**returned exit 1 (no match) instead of falling through to the generic scan** —
so a line the push-down could not answer reported "nothing matched" for a search
it never ran. It also had no shaping gate and no `use_native` term, unlike
`email/grep.py`. Both now share one gate and one `SEARCH_HONORED`.

`extract_folder` had no callers left and is deleted.

While rewriting the TS side to use `patternArg` + `FlagView` (it carried a
bespoke `parseFlags`/`FlagSet`/`getPattern`), a TS-only bug surfaced: it greped
each message as **one line**, so `-n` reported `1` for every hit and printed the
entire message as the matching line. `messageLines()` now splits the way
python's `splitlines()` does.

### email find — python is right, TS gains nothing

`_is_folder_level(paths)` read only `paths[0].mount_path` and `_find_server_side`
called the same `paths[0]`-only `extract_folder`, so
`find /email/INBOX /email/Sent -name '*x*'` searched INBOX and dropped Sent.
Replaced with `_folder_operand`, built on the shared `lone_operand`.

The TS counterpart has no push-down at all. **Python is the right side** — the
IMAP subject search is a real saving over walking every message — so TS is left
as-is rather than having the push-down removed from python; the gate makes the
two agree on every line python now declines to push down.

A second bug fell out: the mount root reached `_find_server_side` and answered
with an **empty result and exit 0**, reporting "nothing matched" for a search it
never ran. It now takes the walk like any other path the push-down cannot serve.

## Tests

Unit tests mirror `langfuse/test_grep.py`: a multi-operand invocation must not
reach the native fetch and must land in the generic scan. Added/updated across
gmail, slack, discord, email (grep, rg and find), the shared helper, and both
scope modules in both languages.

`tests/e2e/test_search_pushdown_iteration.py` was pinning the coalescing
behavior; it is rewritten to assert the opposite — a glob expanded to 60 day
paths now reads those 60 days, makes no native search call, and every output
line names an operand the line actually carried.

13 integ cases pin it byte-identically on both hosts, captured empirically with
`--emit`:

| file | cases |
|---|---|
| `integ/resources/discord/scope.json` | `dc_grep_two_channels_cover_the_second`, `dc_grep_w_second_day_file_only`, `dc_grep_wn_defers_to_scan`, and `dc_grep_w_day_file_native` re-pinned as `dc_grep_w_day_file_scans_only_that_day` |
| `integ/resources/slack/grep.json` | `slack_grep_two_channels_cover_the_second`, `slack_grep_n_defers_to_scan` |
| `integ/resources/gmail/basic.json` | `gm_grep_two_labels_cover_the_second`, `gm_grep_glob_operand_scans` |
| `integ/resources/email/basic.json` | `em_grep_two_folders_cover_the_second`, `em_rg_root_scans_instead_of_no_match`, `em_grep_c_defers_to_scan`, `em_find_two_folders_cover_the_second`, `em_find_root_name_walks` |

## Verification

- Python suite green (`--ignore=tests/fuse`, no macFUSE on this host); e2e green.
- `pnpm -r typecheck` clean; core 9086 passed, node 2370 passed, browser 231 passed.
- `pre-commit run --all-files`: all hooks pass, mypy `Success: no issues found in 1892 source files`.
- Integ battery, discord + slack + gmail + email, **320 passed / 0 failed on
  each host**, byte-identical.
- Both ratchet gates unchanged at baseline: case-targets 2295, layout parity 259.

### DIFF
--- a/integ/resources/discord/scope.json
+++ b/integ/resources/discord/scope.json
@@ -14,7 +14,7 @@
       }
     },
     {
-      "id": "dc_grep_w_day_file_native",
+      "id": "dc_grep_w_day_file_scans_only_that_day",
       "seq": 690201,
       "targets": [
         "discord"
@@ -22,7 +22,7 @@
       "command": "grep -w standup '/discord/Mirage HQ__100000000000000001/channels/general__200000000000000001/2026-06-02/chat.jsonl'",
       "expect": {
         "exit": 0,
-        "stdout": "/discord/Mirage HQ__100000000000000001/channels/general__200000000000000001/2026-06-02/chat.jsonl:[bob] second day standup\n",
+        "stdout": "{\"id\":\"1511278195507200000\",\"channel_id\":\"200000000000000001\",\"author\":{\"id\":\"300000000000000002\",\"username\":\"bob\",\"global_name\":\"Bob\",\"bot\":false},\"content\":\"second day standup\",\"timestamp\":\"2026-06-02T08:00:00.000000+00:00\",\"edited_timestamp\":null,\"tts\":false,\"pinned\":false,\"mention_everyone\":false,\"mentions\":[],\"mention_roles\":[],\"attachments\":[],\"embeds\":[],\"type\":0}\n",
         "stderr": ""
       }
     },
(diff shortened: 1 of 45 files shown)


zechengz:
@codex review

PR #860, opened 2026-08-19
fix(dropbox): bound the emptiness probe's walk, not just its page

bytecii:
## The defect

`list_folder` / `listFolder` took a `limit`, passed it to `/files/list_folder` as **that endpoint's page-size hint**, and then looped `while has_more` calling `/files/list_folder/continue`, accumulating every entry.

So `limit` capped the page, not the walk. Four callers used it as a bounded emptiness probe and got the exact opposite of what they wanted — **one request per child of the folder** to answer a yes/no:

- `python/mirage/core/dropbox/rmdir.py:47` — `list_folder(..., limit=1)`
- `python/mirage/core/dropbox/rename.py:56` — `limit=1`
- `typescript/.../dropbox/rmdir.ts:38` — `{ limit: 1 }`
- `typescript/.../dropbox/rename.ts:46` — `{ limit: 1 }`

This is the same defect fixed for Google Drive and Microsoft Graph in #854.

## The fix

Modelled on Drive's, including the docstring that explains the page-vs-walk distinction. The page hint keeps its own name and the new cap is `limit`:

- page hint renamed `limit` → `page_size` / `pageSize` (still spelled `limit` on Dropbox's wire, where the field genuinely is a page hint)
- new `limit` / `limit` is the **walk cap**: the wire hint becomes `min(page_size, limit)` and the continuation loop breaks once `len(out) >= limit`

Renaming rather than introducing a third name means `limit` denotes the same thing in Drive and Dropbox, and it makes the four probe call sites correct **as already written** — `limit=1` now means the bound they always intended, so none of the four needed an edit.

`readdir` and `watch` pass no cap, so `limit` stays `None`/`null` and the walk still runs to completion. No entries dropped.

## Pinning it — and why the obvious pin is a dud

The pins record the cap each caller asked for, as #854 does for Drive. But a recording-only assertion **does not work here**, and the difference is worth stating:

Because the call sites still read `limit=1` verbatim, the bug lives entirely *inside* `list_folder`. Any fake substituted **for** `list_folder` answers in one call whatever it is handed, so it cannot distinguish a bounded probe from an unbounded one. Confirmed empirically: `assert rpc.list_limits == [1]` **passes against the unbounded form**.

So the fakes stand in for the **transport** (`dropbox_rpc` / `dropboxRpc`) and let the real `list_folder` run, paging the way the live API does (the cursor retains the original request's parameters). They record both the cap and the request count — and it is the count that bites.

Each pin was verified to fail before the fix and pass after:

| pin | unbounded | fixed |
|---|---|---|
| py `test_rmdir_probe_is_bounded_to_one_entry` | `assert 3 == 1` | pass |
| py `test_rename_conflict_probe_is_bounded_to_one_entry` | `assert 3 == 1` | pass |
| ts `rmdir.test.ts` → is bounded to one entry | `expected 3 to be 1` | pass |
| ts `rename.test.ts` → is bounded to one entry | `expected 3 to be 1` | pass |

New fakes: `python/tests/core/dropbox/conftest.py` and `typescript/.../dropbox/_test_util.ts` (named after gdrive's). The TS rmdir/rename tests are new files because the existing `write.test.ts` mocks `./api.ts` wholesale, which is precisely the substitution that cannot see this bug; as a side effect the TS test layout now mirrors python's `test_rmdir.py` / `test_rename.py`.

## One reviewer note

The new `_test_util.ts` lands in a source directory (TypeScript colocates tests), pushing the layout gate 259 → 260. I added a 5-line excuse to `spec/layout_exceptions.json` and **left the baseline at 259**.

Worth knowing: `core/gdrive` carries an identical `_test_util.ts` that sits *unexcused* in the baseline. Excusing both would be more coherent but would mean moving the baseline number, so I left it and said so in the exception's reason text. Happy to do that as a follow-up.

## Verification

- Python **16,129 passed** (`tests/fuse` excluded — no macFUSE on this machine; untouched by this change)
- TS core **9,188 passed** / 702 files · node **2,386** · browser **231**
- `pnpm -r typecheck`, eslint, prettier, knip, `check_spec_parity.py`, `check_layout_parity.py --strict` — all exit 0
- `pre-commit run --all-files` passing; mypy clean on 1,971 files

### DIFF
--- a/python/mirage/core/dropbox/api.py
+++ b/python/mirage/core/dropbox/api.py
@@ -115,17 +115,37 @@ async def list_folder(
     tm: DropboxTokenManager,
     path: str,
     recursive: bool = False,
-    limit: int = 2000,
+    page_size: int = 2000,
+    limit: int | None = None,
 ) -> list[dict[str, Any]]:
+    """List a folder's entries, following every continuation cursor.
+
+    Args:
+        tm (DropboxTokenManager): Dropbox token manager.
+        path (str): folder to list; "/" and "" both mean the account root.
+        recursive (bool): list every descendant, not just the children.
+        page_size (int): entries per request.
+        limit (int | None): stop once this many entries are in hand and do
+            not request another page. An emptiness probe wants one entry,
+            and ``page_size`` alone cannot express that: it caps the page,
+            not the walk, so a small page turned a listing of a large
+            folder into many requests instead of fewer.
+
+    Returns:
+        list[dict]: entry metadata dicts.
+    """
     api_path = "" if path in ("/", "") else path
     out: list[dict[str, Any]] = []
-    resp = await dropbox_rpc(tm, "/files/list_folder", {
-        "path": api_path,
-        "recursive": recursive,
-        "limit": limit,
-    })
+    resp = await dropbox_rpc(
+        tm, "/files/list_folder", {
+            "path": api_path,
+            "recursive": recursive,
+            "limit": page_size if limit is None else min(page_size, limit),
+        })
     out.extend(resp["entries"])
     while resp.get("has_more"):
+        if limit is not None and len(out) >= limit:
+            break
         resp = await dropbox_rpc(tm, "/files/list_folder/continue",
                                  {"cursor": resp["cursor"]})
         out.extend(resp["entries"])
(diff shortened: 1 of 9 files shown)


zechengz:
@codex review

PR #863, opened 2026-08-19
test(cache): service-free pins for the recursive-delete subtree evict

bytecii:
Follow-up to #864, which fixed this bug and pinned it in
`integ/resources/cache/subtree_evict.json`. That battery needs a live minio or
mongo, so the guarantee only holds when the integ job has a container. These
three tests add the same guarantee to a plain `pytest` / `vitest` run.

**No production code** — #864 owns the fix. This PR is coverage only.

## What each one holds

`python/tests/workspace/test_cache_invalidation.py` (+2) — the two shell-visible
symptoms end to end against moto S3, matching the file's existing pattern. A
warm `ls` caches the nested listing, `rm -r` takes the parent, and the reads
that follow used to be answered from cache without touching the bucket:

```
ls /data/a/b         -> 0, "f.txt"      # stale listing
cat /data/a/b/f.txt  -> 0, "hi"         # deleted bytes served
find /data           -> /data           # the backend disagreed, same session
```

`python/tests/core/ram/test_rm.py` and
`typescript/packages/core/src/core/ram/rm.test.ts` — a mirrored pair one layer
down: the real `rm_r` driven through a real `CacheManager` over a RAM index and
file cache, asserting the nested listing and the nested body are both gone. RAM
needs no service, so the path from a backend mutator through `cache/context` to
the manager is exercised on every run.

## Falsifiability

Each was checked against the **merged** implementation, not my own:

- reverting `make_remove_prefix` to `invalidate_after_unlink` fails both moto
  tests
- reverting `core/ram/rm.ts` to `invalidateAfterUnlink` fails the vitest one

## Not included

An earlier revision of this branch carried its own `rm_r` sweep and an
`integ/unix/consistency/subtree.json` case. Both are dropped: #864's sweep is
broader (it covers `rename` too) and its 16-case battery pins the same
behaviour across the same four targets, so the integ case was redundant and its
case-target baseline bump is reverted.

### DIFF
--- a/python/tests/core/ram/test_rm.py
+++ b/python/tests/core/ram/test_rm.py
@@ -1,0 +1,75 @@
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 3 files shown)


zechengz:
@codex review

PR #864, opened 2026-08-19
fix(cache): a recursive delete or directory rename evicts its subtree

bytecii:
## Summary

A recursive delete or a directory rename left **everything cached below the path
still being served**. `ls` printed a deleted directory's contents and `cat`
returned a deleted file's bytes, both exit 0, until the index TTL expired.

The cause: `invalidate_subtree` existed on the cache manager but had exactly one
caller — the watcher. No `rm -r`, no recursive prefix delete, and no `rename`
reached it, in either language. Every mutation site reported
`invalidate_after_unlink`, which evicts the path's own listing and its parent's.
That is complete for a file and blind for a subtree, because each listing and
body beneath the path was cached under its own key and nothing above them evicts
one.

Reproduced against a live minio, with the bucket verified empty through boto3 so
the reads are provably cache and not survival:

```
$ rm -r /s3/a/b                       # bucket now holds zero keys
$ ls /s3/a/b/deep                     # (0) c.txt
$ cat /s3/a/b/deep/c.txt              # (0) hello
$ mv /s3/mv/a/b /s3/mv/a/z            # keys really moved to mv/a/z/
$ ls /s3/mv/a/b/deep                  # (0) c.txt
```

Blast radius is every backend that indexes listings — API backends serve
`readdir` straight from `index.list_dir` — and 30 resources set
`caches_reads = true`, so the body half applies to them too.

## The fix

`cache/context` gains an `invalidate_subtree` free function alongside the two it
already had, and the `CacheInvalidator` protocol/interface gains the method.
Unlike `invalidate_ancestors`, this one **delegates to the manager rather than
walking**: ancestors can be assembled from `invalidate_after_write` calls
because the caller knows the chain, but only the caches know which keys lie
*beneath* a path.

Then one uniform rule, applied in both languages: **an op that destroys or moves
a whole subtree evicts that subtree.**

- every `rm_r` / `remove_prefix` site swaps `invalidate_after_unlink` →
  `invalidate_subtree`
- every `rename` swaps **both** endpoints, because the destination may have
  replaced something that was cached

`invalidate_subtree` is a strict superset of `invalidate_after_unlink`, so the
swap degrades to exactly the old behavior on a file path — the extra prefix
eviction matches nothing. `under_path` boundary checks mean `/a/b` never evicts
`/a/bc.txt`.

Untouched on purpose: `unlink`, non-recursive `rmdir` (the keyed-store one
deletes a marker and nothing else), and `copy`, whose destination-side staleness
is a different symptom I did not reproduce and did not want to fix on a guess.

## The sweep this came out of, and the plan correction

This started as the kit plan's PR-K1b item *"delete-side `invalidate_ancestors`
asymmetry sweep — nextcloud, box, onedrive, sharepoint"*. **That premise was
wrong and the item should not be reopened.**

`invalidate_ancestors` exists because one op can materialize *several* levels at
once: `mkdir -p a/b/c`, and on a keyed store any write, since a put creates every
missing prefix. Nothing on the delete side removes an *ancestor* level on a
real-directory backend, so `invalidate_after_unlink` is already complete there.
The one delete that does need it is the keyed-store one, and
`object_store/remove.py` already had it. The manager's own docstring says as
much: *a backend with real directories cannot gain a level that way, so there
this is a handful of spare evictions.*

box and gdrive only *looked* like they were missing it — their multi-level
`mkdir -p` invalidates per created level in a bespoke loop instead of calling the
shared helper. databricks needed no special re-derivation; its delete side is an
ordinary real-directory delete.

## `emulate_truncate`

The other half of this plan item was the five dead python-only `emulate_truncate`
flags. **#858 already removed them**; only dropbox's live one remains. All that
was left was two docstrings still naming `s3/ssh/ram/redis` as the emulating
backends, corrected here.

## Verification

- **Both hosts, byte-identical, against the committed goldens.** The 16 new
  cases were run as literal shell lines through a python `Workspace` and a
  TypeScript `Workspace` over the same minio bucket: 16/16 agree with each other
  and with `integ/resources/cache/subtree_evict.json`.
- **Prefixed mount checked separately**, since the two `*-prefix` targets are
  pinned here and an unprefixed mount cannot see a key-prefix bug: a mount
  carrying `key_prefix="team/reports/data"` produces the same 16 rows, so the
  manager's cache-key derivation is carrying the prefix through.
- **The pin is falsifiable.** Reverting just the two `object_store` files and
  rebuilding makes all four assertion cases return the stale content at exit 0.
  The first draft of the case file was *not* falsifiable on the rename half — it
  warmed the body but not the listing, so `ls` never had a stale entry to serve.
  Both warms are now present, with a comment saying why.
- Unit tests updated where the assertion moved: a rename and a prefix delete now
  report a subtree rather than an unlink, which is the stronger guarantee.
- `pnpm -r typecheck` clean; widening the interface made tsc enumerate every TS
  fake, python's structural Protocol needed the test run to find its own.
- **Call-site parity checked per backend, not in total.** Python has 36 swapped
  calls and TypeScript 40; every backend matches three-for-three except `ram`
  and `redis`, whose TS `rename` has two return branches to python's one. Both
  branches are covered — the delta is the pre-existing shape, not a gap.
- Both ratchets unchanged: case-targets 2289/2289 (the new file reuses
  `resources/cache/`'s existing 4-target modal set), layout parity 259/259.

## Known coverage gap

The integ pin targets `s3`, `s3-prefix`, `gridfs`, `gridfs-prefix` — the same set
`ancestor_evict.json` uses, which is also what keeps the case-target ratchet flat.
A hierarchy backend with `caches_reads` (nextcloud) exercises the same code path
and is fixed by the same change, but is not pinned here.

### DIFF
--- a/integ/resources/cache/subtree_evict.json
+++ b/integ/resources/cache/subtree_evict.json
@@ -1,0 +1,260 @@
+{
+  "cases": [
+    {
+      "id": "cache_subtree_seed",
+      "seq": 942000,
+      "targets": [
+        "s3",
+        "s3-prefix",
+        "gridfs",
+        "gridfs-prefix"
+      ],
+      "command": "mkdir -p {mount}/sub_dir/a/b/deep",
+      "expect": {
+        "exit": 0,
+        "stdout": "",
+        "stderr": ""
+      }
+    },
+    {
+      "id": "cache_subtree_seed_body",
+      "seq": 942001,
+      "targets": [
+        "s3",
+        "s3-prefix",
+        "gridfs",
+        "gridfs-prefix"
+      ],
+      "command": "printf 'hello' > {mount}/sub_dir/a/b/deep/c.txt",
+      "expect": {
+        "exit": 0,
+        "stdout": "",
+        "stderr": ""
+      }
+    },
+    {
+      "id": "cache_subtree_warm_listing",
+      "seq": 942002,
+      "targets": [
+        "s3",
+        "s3-prefix",
+        "gridfs",
+        "gridfs-prefix"
+      ],
+      "command": "ls {mount}/sub_dir/a/b/deep",
+      "expect": {
+        "exit": 0,
+        "stdout": "c.txt\n",
+        "stderr": ""
+      }
+    },
+    {
+      "id": "cache_subtree_warm_body",
+      "seq": 942003,
+      "targets": [
+        "s3",
+        "s3-prefix",
+        "gridfs",
+        "gridfs-prefix"
+      ],
+      "command": "cat {mount}/sub_dir/a/b/deep/c.txt",
+      "expect": {
+        "exit": 0,
+        "stdout": "hello",
+        "stderr": ""
+      }
+    },
+    {
+      "id": "cache_subtree_remove_recursive",
+      "seq": 942004,
+      "targets": [
+        "s3",
+        "s3-prefix",
+        "gridfs",
+        "gridfs-prefix"
+      ],
+      "command": "rm -r {mount}/sub_dir/a/b",
+      "expect": {
+        "exit": 0,
+        "stdout": "",
+        "stderr": ""
+      }
+    },
+    {
+      "id": "cache_subtree_listing_gone_not_stale",
+      "seq": 942005,
+      "targets": [
+        "s3",
+        "s3-prefix",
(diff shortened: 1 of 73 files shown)


zechengz:
@codex review

Issue #861, opened 2026-08-19
Design: what an agent's command actually fetched — closing the gaps in the op ledger


Jackpkn:
**Type:** Feature request / design &nbsp;·&nbsp; **Area:** Python package + TypeScript packages

_Measured against `main` @ a7cb96de5. Suggested label: `design`._

## Context

`observe/` is a well-factored op ledger, mirrored in both languages: `OpRecord`
(op, virtual path, source, bytes, timestamp, duration, fingerprint, revision),
a `RecordingScope` / `runWithRecording` opened once per top-level typed line,
`push_mount_prefix` so records name virtual paths, and two sinks —
`ws.ops.records` in memory and the `Observer` → `ObserverStore` (RAM / Disk /
Redis) as JSONL. Snapshot drift detection already rides it.

But as instrumented today it answers "what bytes did the agent read or write on
*some* backends", not "what did this command do" or "what did it fetch from an
API". Measured on `main` @ a7cb96de5:

**Metadata ops are invisible on the shell path.** Three identical operations, two
entry points:

```
ws.ops.stat(); ws.ops.readdir(); ws.ops.read()   → 3 records (stat, readdir, read)
$ stat /disk/a.txt ; ls /disk ; cat /disk/a.txt  → 1 record (read)
```

The cause is structural, not an oversight: shell commands go
`executor → mount.execute_cmd` (a `cat`/`ls`/`grep` handler), while FUSE and
`ws.ops.*` go `Ops._call → dispatcher.dispatch → mount.execute_op`. `OpReport`,
the generic "who served this, how many bytes" mechanism, is constructed in
exactly one place — `ops/ops.py:194` — so it only ever serves the op path. Every
`record()` call site across all backends, counted:

```
read 22 · write 12 · create 8 · truncate 6 · unlink 3 · append 3
setattr 1 · rmdir 1 · rm_r 1 · rename 1 · mkdir 1 · copy 1
```

No `stat`, `readdir`, `search`, `find`, or `du`. So `ls` on a bucket, or a `find`
costing dozens of API calls, leaves no trace. (`OpRecord`'s own docstring lists
`"stat"`/`"readdir"` as op types, which is true only via the facade.)

**14 of 37 backends emit records at all.** Instrumented: `box`,
`databricks_volume`, `disk`, `dropbox`, `gdrive`, `gridfs`, `hf_buckets`,
[...]

PR #867, opened 2026-08-19
feat(observe): surface op events and join them to their line

Jackpkn:
Refs #861.

The recorder already wrote an op event per byte transfer, and nothing could read one. Every view filtered to `type="command"` — `/.bash_history` (`core/history/read.py`), the `history` builtin, `ws.history()`, and snapshot capture — so a session that recorded six ops answered with nine commands and zero ops from every public surface. The data was being persisted and thrown away.

Two additions, mirrored in both languages.

**`op_events()` / `opEvents()`** is the missing view, and **`ws.op_history()` / `ws.opHistory()`** is its workspace door, spelled the way `history()` already is. Nothing else about the store changes: the same JSONL, read through a second filter.

**`line_id`** is what makes the two views joinable. An op could only be attributed to its command by append order within a session file, which is a contract that holds until anything writes concurrently. The id is minted once per `RecordingScope` / `runWithRecording`, carried through `push_mount_prefix` / `runWithMountPrefix` so a mount frame cannot drop it, and stamped on both the op records and the command entry. **`line_events()` / `lineEvents()`** is then a filter rather than an inference.

Three consequences, each pinned by a test:

- A nested eval's ops carry the **top-level** line's id. The recorder already scopes them that way (an inactive `RecordingScope` joins its enclosing one), so reporting the inner eval would contradict the scoping that already exists.
- An op raised outside a line carries **none**. A FUSE callback and a direct `ws.ops` / `ws.fs` call have no line behind them, and a synthesized id would claim otherwise.
- Guest code inside a line **does** get the line's id: the runtime's patched `open()` reaches the ops facade, so `Ops._record` reads `active_recorder()` / `activeLineId()` rather than assuming no line is running.

The wire format gains one optional key, `line_id`, absent when there is no line — `to_json_line` already drops `None`, and the TS twin only sets it when present, so an old entry still parses.

Snapshots deliberately keep capturing `COMMAND`/`CLEAR`/`DELETE` only. A snapshot pins what was read through fingerprints and revisions, so the transfer log would grow it without adding a fact it restores from. Documented rather than left as a silent filter.

## Scope, deliberately narrow

This is questions 1 and 2 of #861 only, and it is purely additive — new read paths over data already being recorded, plus a join key. Nothing is removed and no existing behaviour changes, so it does not foreclose the open questions in that issue (metadata ops on the shell path, HTTP-layer instrumentation, the `is_cache` naming, bounding `ws.ops.records`).

Two things I did **not** do, on purpose:

- **No `/.mirage/ops.jsonl` view mount.** #861 question 1 raises whether letting an agent grep its own I/O is a useful capability or a self-observation hazard. That felt like a maintainer call, so this adds the programmatic surface only.
- **No new instrumentation.** Coverage is what it already was — 14 of 37 backends, read/write family only — so `op_history()` is honest about what the ledger currently holds, not more.

`line_id` is my coinage and I have no attachment to it; happy to rename if there is a convention I missed.

## Test plan

Run on `main` @ 8e18c5687:

- `uv run pytest` — 15,913 passed, 495 skipped, 1 xfailed, 0 failures (7 new tests in `tests/observe/test_op_events.py`)
- `pnpm test` (core) — 9,252 passed, 10 skipped, 0 failures (12 new tests across `observe/op_events.test.ts` and `workspace/history_view.test.ts`)
- `pre-commit run` — all 20 hooks pass, including mypy, ruff, PEP8, PathSpec discipline, Prettier, ESLint and Knip
- `tsc --noEmit` — clean
- `scripts/check_layout_parity.py --strict` — 259, unchanged from baseline

The TS tests are split on purpose. The observe unit tests build no shell parser: an earlier draft did, and as the 28th test file to load tree-sitter WASM it pushed the pyodide REPL tests past their 5s timeout and made the suite flaky (a different set of tests failed each run). The plumbing is covered directly in `observe/op_events.test.ts`, and the four end-to-end assertions live in `workspace/history_view.test.ts`, which already builds a parser and already tests views over recorder events.

Docs updated: `docs/home/observer.mdx` gains a "Reading the events" section and states the snapshot exclusion; `CLAUDE.md` gains a History bullet.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -42,6 +42,18 @@ Packages split by role, one module per concern, the same way in both languages:
 Command history is a recording, not a command log. A hidden `Observer` records every top-level command as timestamp-ordered events (`COMMAND`, `CLEAR`, `DELETE`, op events); the user-facing surfaces are just views of those events.
 
 - **Observer + ObserverStore.** The `Observer` owns a storage-agnostic `ObserverStore` (`append`/`write`/`readAll`/`readMatching`/`clear`/`close`), not a mount. Stores: `RAMObserverStore` (core, default), `DiskObserverStore` and `RedisObserverStore` (node). RAM is just the default, history can persist to disk or Redis.
+- **Op events are readable, and joined to their line.** Every event the
+  recorder writes is reachable: `ws.history()` / `ws.op_history()`
+  (`history()` / `opHistory()`) are the two views, and `Observer.line_events`
+  / `lineEvents` returns one typed line's ops followed by its command entry.
+  What joins them is `line_id`, minted once per `RecordingScope` /
+  `runWithRecording` and stamped on every `OpRecord` and on the command
+  entry, so a nested eval's ops carry the *top-level* line's id. An op
+  raised outside a scope (a FUSE callback, a direct `ws.ops` / `ws.fs`
+  call) carries none, which is what "no line caused this" looks like.
+  Snapshots still capture COMMAND/CLEAR/DELETE only: a snapshot pins reads
+  through fingerprints and revisions, so the transfer log would add size
+  without adding a fact it restores from.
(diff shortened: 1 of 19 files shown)


zechengz:
@codex review

PR #871, opened 2026-08-20
fix: measure the gmail excerpt and daemon description budgets in code points

bytecii:
Two more of the code-unit/code-point divergences [#869](https://github.com/strukto-ai/mirage/pull/869) fixed for `sanitizeLabel`/`sanitizeName`, found by sweeping from the python side. Both measure and slice `String.length`, which counts UTF-16 units, where the python twin measures `len` and slices, which count code points.

### `extractExcerpt` (gmail grep), live

`core/gmail/search.ts` builds a grep excerpt with four unit-counted offsets: the `EXCERPT_MAX` cut, `indexOf`, `pattern.length`, and `flat.length`. A 200-emoji body is 200 code points and 400 units, so python returns all 200 and typescript returned 117 — ~40% of the excerpt dropped on the same message. The windowed branch is worse than short: `slice(286, 532)` on an emoji run cuts *inside* a surrogate pair at both ends, and the lone halves encode as U+FFFD in the grep line.

Every offset now runs over `Array.from(flat)`. `indexOf` still finds the hit in units, so its result is converted once (`Array.from(lower.slice(0, hit)).length`) rather than the whole search being rewritten.

### `describeResource` (daemon workspace detail), latent

`server/summary.ts` ellipsizes a resource's prompt at 120. Every `PROMPT` mirage ships today is ASCII, so nothing diverges in practice; fixed for consistency rather than left as a known-divergent twin — a prompt of astral characters would have ellipsized at half the budget.

### Pins

Both sides get parity pins mirroring `sanitize.test.ts` / `test_sanitize.py`: an input python leaves whole must survive intact, and an over-budget astral input must cut on a code-point boundary with no U+FFFD. The excerpt pins go through the exported `formatGrepResults` / `format_grep_results`, so they run the real function rather than a copy of it, and every expected string was read off the python implementation before being written down. Reverting either fix fails exactly the two new astral cases per language.

`describeResource` is exported for its pin; the equivalent python name is already importable. Knip is clean.

Two new python suites (`tests/core/gmail/test_search.py`, `tests/server/test_summary.py`) close mirror gaps; neither moves `MIRROR_BASELINE`, which already counted both basenames.

### Verification

- `pnpm -r typecheck`, `pnpm exec knip`
- vitest: core 9251, browser 212, node 2399, server 182 — all pass
- `pytest --ignore=tests/fuse` — 15741 passed, 491 skipped, 1 xfail (no macFUSE on the dev host)
- `pre-commit run --all-files` — every hook passes; the mypy hook's "Failed" line is its `uv run` rewriting `uv.lock`, and it reports `Success: no issues found in 1974 source files`

### DIFF
--- a/python/tests/core/gmail/test_search.py
+++ b/python/tests/core/gmail/test_search.py
@@ -1,0 +1,79 @@
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+
+from typing import Any
+
+from mirage.core.gmail.scope import GmailScope
+from mirage.core.gmail.search import format_grep_results
+
+SCOPE = GmailScope(use_native=True,
+                   label_name="INBOX",
+                   date_str=None,
+                   resource_path="/")
+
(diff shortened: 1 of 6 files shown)


zechengz:
@codex review

PR #872, opened 2026-08-20
refactor(browser): the resource layer inherits BaseResource instead of restating it

bytecii:
## What

`packages/browser`'s resource layer never received the `BaseResource` rollout that `packages/node` and `packages/core` have. This aligns it — browser adopts the base class that already exists in core, rather than a new design.

| axis | before | after |
|---|---|---|
| extends `BaseResource` | browser **0/19** (node 28/28, core 8/8) | 19/19 |
| `deltaHook` | missing on 5 browser backends node has | present |
| config from core | 1 duplicated module + 2 inlined normalizers | taken from core |

## Why it isn't cosmetic

**`index:` mount config was silently ignored in the browser.** Core's `Workspace` hands it to the resource:

```ts
// workspace/workspace/workspace.ts
resource.setIndex?.(options.index)
```

`setIndex` exists only on `BaseResource`. Optional-chained against 19 resources that didn't extend it, that call did nothing — so a browser mount with `index: {type: 'redis'}` or a custom `ttl` quietly got a default RAM store, where node and python honored it. Browser's `Workspace` only forwards to core's constructor, so nothing else intercepted it.

Three more consequences of the same cause:

- Each hand-rolled `close()` was `return Promise.resolve()` and never closed the index it built — the exact leak `BaseResource.close` was written for: *"a mount configured `index: {type: redis}` holds a client that nothing else closes"*.
- Every constructor eagerly built a `RAMIndexCacheStore`, including for resources that never read one. The base builds it lazily.
- `workspace/mount/storage.ts` carries a WeakMap serial-number fallback whose comment names this gap. It stays for third-party resources; no first-party browser resource needs it now.

I checked the one thing that would have justified the split — whether `BaseResource` drags Redis into a browser bundle — and it does not: `cache/index/redis.ts` reaches redis through `loadOptionalPeer`, and redis is not a core dependency.

## The two riders

**`deltaHook` on box, dropbox, gdrive, github, s3.** Node had it, browser didn't, so watch/delta was silently absent. All five core `buildDeltaHook`s take only an accessor and import nothing runtime-specific. Before wiring s3 I confirmed its presigned-URL shim can actually serve a listing walk: `client_browser.ts` implements `ListObjectsV2Command`, honors `ContinuationToken`, and returns `IsTruncated`/`NextContinuationToken`.

**linear + github config.** Browser kept a `linear/config.ts` duplicating `core/linear/config.ts` field for field, and `registry.ts` inlined the rename maps core exports as `normalizeLinearConfig` / `normalizeGitHubConfig`. Both now come from core, as node takes them, which also retires the restated `LinearBrowserCtorConfig`. Python keeps linear's config at `core/linear/config.py` with no `resource/linear/config.py`, so this matches python's layout as well. `linear.ts`'s import block is now byte-identical to node's.

## Deliberately not changed

Three configs looked like the same drift and are not:

- browser **discord**/**slack** take `{proxyUrl, getHeaders}` against core's `{token}` — a browser cannot hold a bot token or reach those APIs directly, so it proxies
- browser **notion** takes an MCP `authProvider` where core takes an `apiKey`
- browser **s3**'s config is the presigned-URL variant

Repointing any of them at core would have broken browser auth outright.

## Test

`base_contract.test.ts` pins both halves — a sweep asserting every exported `*Resource` inherits `BaseResource` (so a future resource cannot re-open this), plus a behavioral pair on the mount index config. Restoring the pre-change `trello.ts` makes it fail with `TypeError: r.setIndex is not a function`; 44 pass after.

Python needed no change: all its resources already extend `BaseResource` (the 14 s3 aliases inherit through `S3Resource`).

## Verification

- `pnpm -r typecheck` clean; `pnpm -r test` green — core 9436, browser 256 (+44 new), node 2417, dsh 156, server 178, agents 189, cli 93
- every pre-commit hook passes
- gates: layout **257 → 256** (dropping browser's duplicate linear config closed a divergence, and the gate fails on improvement too), barrel 77, spec parity 93, case targets 2289/2289

Follow-up left out of scope: 26 of browser `registry.ts`'s rename entries merely restate what `normalizeFields`' default `snakeToCamel` already does (the 15 real ones are all `endpoint_url → endpoint` for S3 aliases, which are load-bearing), and 7 more `*BrowserCtorConfig` interfaces restate config shapes where node declares none.


zechengz:
@codex review

PR #874, opened 2026-08-20
feat(hierarchy): move discord, slack, gmail and email onto the kit

zechengz:
Migrates the four wave-2d backends (discord, slack, gmail, email) onto the core/hierarchy kit, in both languages. The hand-rolled dataclass scope classifiers, readdir ladders and stat/read plumbing are replaced by kit scope tables, listers, guards and readers; grep/rg/head wrappers gate on a per-backend NATIVE_KINDS set plus match slots instead of scope.use_native.

Kit extensions (tested in tests/core/hierarchy and the .test.ts twins):

- DirListing(entries, seeds): a lister can return descendant listings the same fetch proved (discord/slack day fetch seeds the files dir; gmail label listing seeds date and attachment dirs), with a post-resolve re-check so a resolve that ran the seeding parent is not refetched.
- parent_entry_listers: a third lister table whose existence proof is the parent dir entry, for day directories that must list beyond the synthesized window.
- make_read_range: byte windows; ranged kinds (discord/slack file blobs) push offset/size to the source, everything else slices the full read.
- DATE codec: shape-only ISO date segment validation.

Behavior changes, pinned by unit tests; integ goldens needed zero changes (374/374 per host, both hosts, targets discord/slack/gmail/email/argerr-*/cli-discord/cli-slack):

- bare name segments without __id are now ENOENT everywhere (the tree always mints name__id)
- static container dirs under a nonexistent guild are ENOENT (was an unconditional dir)
- discord/slack files dir under a sealed day is ENOENT (was a dir)
- cat of an existing gmail label/date/attachment dir is ENOENT, not EISDIR (matched shape is not existence; the mount root stays EISDIR)
- gmail readdir bootstrap failures propagate instead of collapsing to ENOENT
- discord/slack cold reads work (old read required a warm index)
- deliberate divergence kept: discord stats a sealed day's chat.jsonl as size-unknown, slack answers ENOENT (pinned in both)

Verified: full python suite + mypy clean, pnpm -r build/typecheck/test, layout parity at baseline, pre-commit fixpoint.

### DIFF
--- a/python/mirage/commands/builtin/discord/grep.py
+++ b/python/mirage/commands/builtin/discord/grep.py
@@ -31,7 +31,7 @@ from mirage.core.discord.entry import channel_dirname
 from mirage.core.discord.formatters import format_grep_results
 from mirage.core.discord.read import read as discord_read
 from mirage.core.discord.readdir import readdir as _readdir
-from mirage.core.discord.scope import detect_scope
+from mirage.core.discord.scope import NATIVE_KINDS, detect_scope
 from mirage.core.discord.search import search_guild
 from mirage.core.discord.stat import stat as _stat
 from mirage.io.types import ByteSource, IOResult, materialize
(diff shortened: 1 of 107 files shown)


zechengz:
@codex review

PR #875, opened 2026-08-20
feat(hierarchy): move lancedb and qdrant onto the kit with per-mount routes

zechengz:
Migrates lancedb and qdrant onto the core/hierarchy kit, in both languages. These are the two backends whose route table is a function of mount config (pinned table/collection, group_by depth, config-gated leaf set), so their scope tables cannot bind at module import like the other 17 consumers.

The per-mount mechanism is one new kit helper, hierarchy/bind per_accessor (perAccessor): a weak map keyed on the accessor that caches one build of make_detect_scope(scopes_for(config)) plus the three factories per mount. The kit factories themselves needed no change for this; they were already pure functions. Listers, guards and readers stay module-level functions that read accessor.config at call time, so the existing patch-the-client-fns test convention is unchanged and the module-level readdir/stat/read signatures (and all io.py/ops/resource wiring) are untouched.

Scope shape, identical for both backends: a pinned table/collection drops the leading table slot; group slots are named positionally (g0, g1, ...) so a column literally named "table" cannot collide with the table slot, with table_of/filters_of mapping back to column names; every partial depth shares one "group" kind whose lister derives the depth from the slots, so the lister table stays static while the scope table varies per mount. Leaf codecs are config-built (.md and blob for lancedb; .json, .txt and blob for qdrant).

One kit fix rode along: make_readdir and make_stat now normalize index=None the same as NULL_INDEX (a call-local RAM store). Bare commands built outside a workspace bind None; the old backends tolerated it and the kit crashed. The TS twin was already safe via ??. Pinned by new kit tests.

Behavior changes, pinned by unit tests; integ goldens needed zero changes (95/95 per host, both hosts):

- cat of the mount root is EISDIR (was ENOENT), matching the rest of the kit family
- ls of a nonexistent table/collection is a clean ENOENT instead of a leaked driver error
- dot-led segments are refused and dropped from listings
- every listing level now seeds the index (old code seeded only the rows level)
- qdrant .json leaves keep FileType.TEXT, the old behavior in both languages

Verified: full python suite + mypy clean, pnpm -r build/typecheck/test, layout parity at baseline (bind lands on both sides), pre-commit fixpoint.

### DIFF
--- a/python/mirage/core/hierarchy/bind.py
+++ b/python/mirage/core/hierarchy/bind.py
@@ -1,0 +1,50 @@
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+
+from collections.abc import Callable
+from typing import TypeVar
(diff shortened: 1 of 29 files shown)


zechengz:
@codex review

PR #879, opened 2026-08-21
Serve symlink, readlink and setattr to guests

zechengz:
A guest could not create a symlink, read one, or stamp a mode or a time. The three verbs reach the name plane rather than a backend, so they work on a mount whose store holds none of them: wasi, quickjs, pyodide and monty all route through the same ops.

## Transport

`RuntimeVFS` gains `symlink`, `readlink` and `setattr` in both languages, and `BridgeDispatchFn` gains the three cases, where `dst` carries both a rename's destination and a symlink's target. `SetAttrFields` moves to the shared root `types.ts`, because the ops facade and the runtime bridge both take it and neither may import the other.

TS gains the `posixMode` and `LINK_MODE` twins of the python helpers, and `VFSStat.mode` is now required. That closes a real gap rather than adding a field: a quickjs guest's `os.stat().st_mode` was a synthesized constant, so a `chmod 600` the shell had made read back as `0o666`.

## Per engine, and the shim mimics the engine

Measuring the engines first overturned two plan items, so nothing here invents a verb its host does not have.

- **wasi**: `path_symlink`, `path_readlink` and `path_filestat_set_times`. `path_filestat_get` reads `LOOKUP_SYMLINK_FOLLOW` and lstats when the bit is clear. `fstflags` splits each stamp into "write the argument" and "write the host clock", and neither bit set is UTIME_OMIT. `path_link` answers the verb table's refusal for `link`, rendered in preview1 numbers, rather than a hardcoded errno.
- **quickjs**: qjs-wasi has no `symlink`, `readlink` or `lstat`, so the bootstrap gains exactly `os.utimes` and `os.S_IFLNK`, which is what the real binary has.
- **pyodide**: an `S_IFLNK` node, a `readlink` node op, and journal `markSymlink` / `markSetattr`. The preload now copies a link as a link instead of skipping it, one readlink paid only for entries the namespace already marked; before this, a link in a mount tree was simply absent from the guest.
- **monty**: `Path.is_symlink` in both languages. TS reads the readdir mark; python asks readlink, because a python row carries no mark yet.

## Emscripten finalizes a create through setattr

pyodide's journal grew a phantom `setattr` per created file and, worse, split two writes that `markWrite` would have coalesced, because a create is finalized through `node_ops.setattr` with the same shape a guest chmod has. Comparing the fields is not enough on its own: the create genuinely lowers 0o777 to 0o666. `MirageFs.fresh` marks the node `mknod` just made and `changedAttrs` drops a write that moves nothing, so a create journals one write and a no-op utime journals nothing.

## readlink splits its two misses

`readlink` answered EINVAL for every miss. POSIX splits them, and a caller's `except FileNotFoundError` depends on it. Pinned against Linux (`python:3.13-slim`) for all six cases: a link resolves, a file, a directory and a mount root are EINVAL, and a missing path, or one under a missing parent, is ENOENT.

Absence is probed on the failure path only, and it takes both channels a backend can answer on: on a prefix store a directory is not an object but the set of keys under it, so `stat` misses what `readdir` lists, and a listing has to be non-empty because those stores answer a missing prefix with `[]`. The namespace is asked first, since a directory that exists only because a mount or a link sits below it is structure no backend can see. `resolve_path_stat` could not be reused: it dispatches, and this door is what dispatch is inside of.

Three things the probe got wrong first, all now pinned by tests:

- It passes `pre_ops_gate`. It reads on the caller's behalf but not at its request, so a policy denying `stat` must not become reachable through a readlink.
- A refused channel is not absence. It collapses back to EINVAL, which asserts nothing the policy is withholding, where ENOENT would assert a fact this door was not allowed to check.
- The TS dispatcher does not route through `MountEntry.executeOp`. A MountEntry built by the Workspace constructor has no ops registered at all, so the probe calls `opsRegistry.call` and stamps index and filetype the way the dispatcher does.

`os_patch._link_target` swallowed every `OSError` as "not a link", which would have hidden the new ENOENT from `os.lstat`; only EINVAL is swallowed now. The `readlink` command already funnelled every failure into GNU's silent exit 1, and FUSE reads the node table directly, so neither of those changes.

## Also

- `isMissError` moves from `links/probe.ts`, where it was private to one file, to `utils/errors.ts` beside python's `MISS_ERRORS`.
- `docs/typescript/python-fs.mdx` carried a row saying symlinks are refused with EPERM, which was false before this PR too. `os.utimes` added to both JS runtime pages.

## Known gaps, deliberate

- A python readdir row carries no link mark, so a wasi guest's `scandir().is_symlink()` is False and a link to a directory reads as a directory. TS marks it because its bridge holds the namespace; python's door holds only dispatch, so closing it means giving the runtime tier a name-plane door. Next PR.
- pyodide's seed carries no mode and no mtime, so a guest stat there does not reflect a shell chmod (quickjs's now does). Same next PR: both fields ride the row each door already builds.
- Guest hard links stay EPERM, which is what link(2) documents for a filesystem that cannot hold a second name for one inode, and what the verb table already decided.

## Verification

- python suite green, 0 failures
- TS core 728 files, 9573 tests; every TS package typechecks
- pre-commit clean in one pass, nothing reformatted
- layout parity `--strict` at baseline
- integ runtime: 116/0 python, 122/0 TS, including 10 new cases across wasi, pyodide, quickjs and monty

### DIFF
--- a/docs/python/runtime/javascript.mdx
+++ b/docs/python/runtime/javascript.mdx
@@ -58,7 +58,7 @@ The engine runs under WebAssembly capability isolation: it sees only what
 the run passes it. Host files and the network are invisible. Workspace
 mounts are visible with no setup: mirage intercepts the sandbox's
 filesystem calls and bridges them through the workspace dispatch, so
-`std.open('/data/f.txt', 'r')` reads the mount live, and `os.stat`/`os.mkdir`/`os.rename`/`os.remove` stat and mutate it (0 or -errno in WASI numbering, like the real engine) — RAM, Redis, S3, or
+`std.open('/data/f.txt', 'r')` reads the mount live, and `os.stat`/`os.mkdir`/`os.rename`/`os.remove`/`os.utimes` stat and mutate it (0 or -errno in WASI numbering, like the real engine) — RAM, Redis, S3, or
 a virtual backend — and writes are immediately visible to every command,
 with the same cache, write modes, and per-session mount narrowing as
 `cat`. A read-only mount (or a session narrowed to read) fails the open,
--- a/docs/typescript/python-fs.mdx
+++ b/docs/typescript/python-fs.mdx
@@ -66,6 +66,23 @@ open('/ram/notes.txt', 'w').write('hello')              # writes flush back to R
     ```
     Anything that goes through Python's `open()` works (PIL, numpy, pandas, json, pickle).
   </Tab>
+  <Tab title="Symlinks & metadata">
+    ```python
+    import os
+
+    os.symlink('report.json', '/ram/latest.json')
+    print(os.readlink('/ram/latest.json'), os.path.islink('/ram/latest.json'))
+
+    os.utime('/ram/report.json', (1767225600, 1767225600))
+    os.chmod('/ram/report.json', 0o600)
+    ```
+    A link is namespace state, so no backend has to be able to store one:
+    the op reaches the name plane and a link lands on an S3 or Notion mount
+    the same way it lands on a RAM one. Links the shell made are collected
+    with the prefix, as links rather than as copies of their targets, so a
+    cyclic one is safe to walk past. A `chmod` or `utime` a backend cannot
+    keep is stored in the namespace overlay and reported back by `stat`.
+  </Tab>
   <Tab title="Cross-mount">
     ```python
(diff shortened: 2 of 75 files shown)


zechengz:
@codex review

PR #880, opened 2026-08-21
feat(policy): decision ledger, explain, and whole-line judging

zechengz:
Folds the remaining permissions work into one PR.

## Outcome is the document's three verbs

`ALLOW | ASK | DENY`, nothing invented beside them. `RUN` and `NOT_ALLOWED` are gone. `Ruling.rule` tells the two allow-list refusals apart, and the exit code separates 127 (head word not visible, so an unlisted tool never leaks that it exists) from 126 (head visible, line not covered).

## ws.explain

Answers what every command of a line would do, without running it. It spends no answer, records no question and never reaches the host, so a host can read a line and an agent can type it and they cannot be told different things. The gate is split into a judging half and an enforcing half so both halves read the same tables.

## A line is judged before any of it runs

The reason is replay: an ask is only meaningful if approving it runs the line exactly once, and that requires nothing to have run. A head word the session cannot see is a routing miss, not a verdict, so it stays bash and a typo cannot cost an agent the work the line already did.

The hold reaches as far as the text does. The pass reads a line's text while the gate reads its values, so a computed path (`cat $S`, `$( )`, `cd $VAR`) is judged at the gate, after the earlier commands ran. That limit is documented in place and pinned by a test rather than left implied. Closing it would mean asking whenever a word cannot be read, which over-asks with no way out for a deny, so it is left to a future policy script that can resolve the value itself.

A literal `cd` earlier in the line moves what later rules read, with subshell and pipeline scoping pinned against bash (`( )`, `|`, `&`, `$( )` do not let a `cd` escape; a brace group and an `if` body do).

## ws.decisions replaces the approvals machinery

Ask state used to live in two stores: pending questions in `RecordApprover`'s private map, which did not survive a restart, and answers on the session, which did. A host could see a question that no longer existed or miss one that did.

Now there is one record type and one store. A `Decision` with no `outcome` is a question waiting, one with an `outcome` is settled, `scope` says how far the answer reaches, and all of it persists with the session.

Deleted: `Approver`, `RecordApprover`, `CallbackApprover`, `Approvals`, `Grant`, `ApprovalRequest`, `ApprovalDecision`, `GrantScope`, `SessionGrantsQuery`, `EXACT_LINE_DECISIONS`, `request_id`, `Session.grants`, the `approver=` knob.

`grant(id)` and `deny(id)` collapse into `answer(id, outcome, scope=, note=)`, since they were one operation with two names. The `Approver` protocol with two implementing classes becomes one optional coroutine:

```python
async def allow(record: Decision) -> Decision:
    return dataclasses.replace(record, outcome=Outcome.ALLOW, scope=Scope.SESSION)

Workspace(..., on_ask=allow)
```

`policy/config.py` and `config.ts` held one frozen dataclass and no configuration, so `Ruling` moved into `types.py` where the layout rule puts data shapes, and the module is gone from both languages.

## Validation, internal only

Two load errors, raised where the document compiles so a deployment learns at startup rather than the first time an agent types the line the operator thought was covered:

- a rule naming a builtin the allow list never installs (it reads as a guard and is not one)
- an ask an outranking deny already covers at the same anchor

Across anchors the deeper rule leads and which one that is depends on the line, so nothing is reported there rather than guessed at; `conflict_a_later_tiers_deny_outranks_an_earlier_tiers_ask` is exactly that document and a test names it as the reason.

A rule naming a verb its CLI does not declare is refused at `create_session`, where the CLI exists to be checked against.

## Behavior change worth calling out

`echo $(rm /scratch/z)` where the `rm` asks used to exit 0 with the removal silently not happening. It now refuses the line at 126, because the substitution's own command is a command of the line.

## Verified

| gate | result |
| --- | --- |
| python full suite | exit 0 |
| python mypy | clean, 2002 files |
| TS core suite | 729 files, 9552 tests |
| `pnpm -r typecheck` | 0 |
| pre-commit | clean at a fixpoint |
| python integ `ram` / `ram-commands` | 2845 / 132 |
| TS integ `ram` / `ram-commands` | 2845 / 132 |
| layout parity `--strict` | 256, baseline |
| case targets `--strict` | 2289, baseline |

Integ case files keep `answer: allow_once | allow_session | deny` as the battery's own vocabulary; both harnesses map it onto an outcome and a scope, so no goldens moved.

---

## Review round

Three P1s, all real, all fixed in both languages with tests that fail against the previous commit and pass after.

- **Coded policies were never prejudged.** The pass returned early when the session carried no document, and `MountRootPolicy` is always registered. Removing the guard was not enough on its own: a coded policy denies with no document rule to point at, and the pass read "no rule" as "no verdict".
- **The dry run never passed a statement's redirect targets to the gate**, so `explain("echo x > /protected")` answered ALLOW.
- **An ask the host answered inline ended the scan**, so a later command's deny ran behind an approval.

### Two latent bugs the fix exposed

Both broke integ once the pass applied to every workspace, and both are fixed here.

- Redirects were attributed to the statement's first child, but `a && b > f` parses as one redirected statement wrapping the whole list, so the target belongs to the **last command of the chain**. A compound (`{ }`, a loop, a subshell) redirects every command inside it, which is not a chain, so none is claimed and the op door judges the write.
- Fork tracking read a subshell as "no `cd` applies" where bash means "no `cd` **escapes**", which judged `(cd d && tar -c ..)` at the session cwd and made `..` read as a mount root. The walk is now scoped and returns the session each scope ends in; a pipeline is deliberately not a fork scope, because it forks per segment.

### One deliberate call

A line with one command to judge is left to the per-command gate. Not an optimization: nothing runs before it, so there is no hold to buy, and the gate refuses from inside the shell where the line's own redirections still apply. Refusing `rm -rf /mnt 2>&1` in the pass wrote to stderr where bash puts it on stdout.

### Integ now cross-checks `explain`

Every case on a target carrying a document reads the line through `ws.explain` first and asserts three things: the dry run records no question, a refusal it predicts is the refusal that arrives, and **a refusal that arrives was predicted**. The last is found by looking for one of the document's own rule reasons in what the run printed, and is the direction a prediction cannot check on its own. Against the previous commit it catches the redirect bug and two more; it passes here.

Seven cases carry `explain_blind` with the reason the command plane cannot reach their verdict (a runtime-expanded glob, a refusal from the op door, a function the same line defines), named rather than silently skipped, following the `conformance_skip` precedent.

The battery's three answer words are a real type in both harnesses now. They were compared as open strings, so anything that was not `allow_once` fell through to a session-wide allow and a misspelled word passed the case while testing the most permissive answer there is.

### Re-verified after the fixes

python suite exit 0 · TS core 729 files / 9558 tests · `pnpm -r typecheck` 9 packages · integ `ram` 2845 and `ram-commands` 132 on **both** hosts · layout parity 256 and case targets 2289, both at baseline · pre-commit clean at a fixpoint.

The TS integ battery imports `@struktoai/mirage-core` from `dist`, so those TS numbers are post-`pnpm -r build`.


zechengz:
@codex review

PR #882, opened 2026-08-21
One stat struct at the runtime door, and the surfaces that project it

zechengz:
Every runtime surface asked the mounts for a stat and shaped the answer itself, four different ways. This makes the door the one place that builds it, and each guest a projection of that one struct.

## What was different, and is not any more

- **TS had a door-level stat type, python did not.** `VFSStat` was declared and validated in `runtime/vfs.ts`, while python's `RuntimeVFS.stat` returned the raw `FileStat` and every surface translated it. Both languages now declare the struct and build it at the door out of the mount's own row.
- **The field sets disagreed.** `VFSStat` carried `mode` and no link bit; python's `GuestStat` carried `is_link` and no mode. Both carry all of it now, and `GuestStat` (with `runtime/wasm/types.py`) is gone.
- **No-follow was asked three ways.** `WasmVFS.lstat` reached past the door with `stat(..., nofollow=True)`, TS monty read the mark off a readdir row, python monty probed with `readlink`. `stat` takes `nofollow` now, so lstat is one door question; the dispatcher consumes the flag and gates that read exactly as it gates `readlink`.
- **Even TS was not converged.** The `FileStat` to `VFSStat` translation lived in the workspace bridge, so the door validated a shape it did not build. The bridge answers `stat` with the raw row now, the same narrowing #881 did for `readdir`.

## What the row and the surfaces gained

`VFSEntry` carries the `mode` and stamp its stat already answered, and reports neither on the two slash-marked rows that never stat. A guest seeding a whole tree from one listing then needs no second stat per file.

- python monty's `path_stat` answered from its own tree, where a materialized file is a `MemoryFile`: `0o644` whatever the mount held, and an mtime of the moment it was fetched. It answers from the row now, and an unknown stamp stays epoch zero instead of becoming the host clock.
- pyodide's seed carried neither field, so every seeded node stamped `Date.now()` and a chmod the shell made was invisible. Both ride through the seed as `chmod` and `utime`, Emscripten's own spelling, and only a seeded node is restamped: a file the guest creates during the run really was modified now.
- wasi and quickjs needed no change.

## Two facts now written down rather than rediscovered

- **The mtime unit cannot be unified.** Epoch nanoseconds are past 2\*\*53, so a JS number cannot hold them exactly. python keeps `mtime_ns`, TS keeps `mtimeMs`.
- **`posix_mode` already sets the type bits**, so `is_dir` and `is_link` are mode's type bits spelled out. Mode is the authority, and a wire with no mode field of its own (preview1) reads the kind out of it.

## Verification

- python suite green, 0 failures
- TS: 13,107 tests across all 8 packages, 0 failures (core alone is 728 files / 9612 tests); `tsc --noEmit` clean; eslint clean; `pnpm -r build` green
- integ: runtime 92/0 python, 124/0 TS, plus the wasi suite 107/0 against a real CPython WASI guest
- pre-commit clean and stable on a second pass
- every new behavior A/B'd by reverting it: without the monty change a live guest prints `0o100644` and an mtime of now, and without the pyodide plumbing the seeded node reports the tree's defaults
- the first commit was verified green on its own, with the second commit's sources reverted, so the split bisects

## Not in scope

- **ts monty's `Path.stat` stays declined, and it is not a missing case.** That binding converts whatever its `os` callback returns structurally, so a stat reaches the guest as a dict and `st.st_size` raises `AttributeError`; probed with four shapes (plain object, class instance, tuple, Proxy) against @pydantic/monty 0.0.19. The JS package exports no `StatResult`, where python's binding hands over an `OSAccess` subclass and can build monty's own. Its conformance row now carries that reason.
- **python's quickjs cannot report a mode either**, because it runs over WASI preview1, whose filestat has no mode field. TS quickjs can, since it uses emscripten host functions. Same runtime name, two wires, one of which cannot carry the fact.
- `atime` is still not on the struct. Only `mtime` had a consumer.


zechengz:
@codex review

PR #891, opened 2026-08-23
test: add gws slides CLI coverage

sonhmai:
Add missing `gws slides` CLI coverage.

- integ: `presentations get/create/batchUpdate`
- unit: pin slides tree shape + write flags in python/typescript
- verified on `cli-gws` python target


zechengz:
@codex review

PR #892, opened 2026-08-23
ci(python): run main pytest suite with xdist

sonhmai:
Python CI time is mostly in the main pytest run.

Run the non-Redis main suite with `pytest-xdist` using `-n 4 --dist loadscope`, and keep the Redis-backed tests serial.


sonhmai:
@codex review

zechengz (approved):
Thanks :)

PR #895, opened 2026-08-23
fix: refuse a directory read with GNU's errno and exit code

zechengz:
## Summary

`base64`, `sort`, `xxd` and 20 more commands answered `No such file or directory` for a directory operand, because the read slots forwarded the backend's own error and a keyed backend has no key at a directory. Refused in the slot now, so no builder has to remember.

- `with_dir_guard` / `withDirGuard` on `read_bytes`/`read_stream`/`read_range`, applied where the factory already stamps the namespace fact. Post-hoc, so a successful read costs what it did.
- `READ_FAIL_EXIT` beside `USAGE_EXIT`. GNU's code for a failed read belongs to the command, not the errno (`sort` is 2 either way, `cat` is 1), with four exceptions that split by errno: `sed` 4/2, `gzip`/`gunzip`/`zcat` 2/1, `zgrep` 1/2. Gated on filesystem errors, so `sed 's/o/O/0'` keeps exiting 1.
- `sed` aborts on a directory and continues past a missing file, which is GNU's split between a read error and an open error. Every other command in the family continues, which already matched.
- A directory no longer routes to a filetype handler. The cascade picked one from the operand's name, so a directory named `dir.tally` ran a registered renderer and reported ENOENT.

Also fixes the missing-file exit code for the same commands, since GNU's code is per command rather than per errno.

GNU truth pinned on `debian:stable-slim`: coreutils 9.7, sed 4.9, gzip 1.13, jq 1.7, binutils 2.44, util-linux 2.41.5, bsdmainutils 12.1.8, xxd from vim-common, unzip 6.00, ripgrep 14.1.1. Correcting the original report: `xxd` does have upstream truth (it ships with vim), exit 2.

## Test plan

- `python/tests/commands/builtin/test_read_failure_exit.py` and its TS twin pin 36 command lines against the docker measurement, plus 9 multi-operand cases
- `integ/unix/readfail/dir.json`, 60 cases on `ram` and `disk`
- pytest 0 failures; TS core 734 files / 9812 passed; mypy clean; `pnpm typecheck` all packages; layout parity at baseline; pre-commit green and idempotent
- integ matching on both hosts: `ram` 2905, `disk` 2863, `ram-nested` 23

## Out of scope, found while tracing

- `cat /disk/noperm.txt` leaks the host path for any errno but ENOENT. `core/disk/errors.py::disk_errors` exists but is applied on the write side only.
- `echo x > /ram/dir` on a keyed backend exits 0, loses the bytes, and `find` then lists the path twice.
- `CommandOpts.filetype_fns` is set and read by nothing.
- Deleting the now-redundant `dir_aware_stream` (7 python call sites, 18 TypeScript) is left for its own change: it removes an eager refusal that the lazy-stream commands rely on for GNU's report-and-continue ordering.

`CLAUDE.md` was not edited. Nothing there describes the read chokepoint today; worth adding if you want it recorded.


zechengz:
@codex review

zechengz:
@codex review

Issue #888, opened 2026-08-22
Design: add NFS mount backend — userspace NFSv3 server over MountCore


Jackpkn:
> **v2.** Revised after maintainer feedback ([comment](https://github.com/strukto-ai/mirage/issues/888#issuecomment-5384434617)): use an existing NFS server library instead of implementing the protocol, no threading in mirage. Server library per maintainer decision: **huggingface/nfsserve**. The v1 from-scratch design is in this issue's edit history.

## Motivation

NFS provides a kernel mount alternative for environments where FUSE is unavailable or restricted. Primary target is **macOS** (avoids the macFUSE kext requirement); secondary is **Linux**. Same core/adapter split as the existing FUSE implementation.

**Licenses:** [nfsserve](https://crates.io/crates/nfsserve) is BSD-3-Clause (Apache-2.0-compatible dependency; ~490k downloads, maintained by Hugging Face, 0.11.0 Apr 2026). redis/agent-filesystem remains AGPL-3.0 — architecture reference only, no code derived from it. [Vaiz/nfs3](https://github.com/Vaiz/nfs3) (`nfs3_server`, BSD-3-Clause, same trait shape and lineage) is the documented fallback if nfsserve's gaps ever bite — a contained migration, not a rewrite.

## Architecture

```
┌─────────────────────────────────────────────┐
│         MountCore (fuse/core.py)            │
│    POSIX semantics — shared                 │
└─────────────────────┬───────────────────────┘
                      │
        ┌─────────────┼─────────────┐
        ▼             ▼             ▼
   ┌─────────┐   ┌─────────┐   ┌──────────┐
   │  FUSE   │   │  FSKIT  │   │   NFS    │
   │ Adapter │   │ Adapter │   │ Adapter  │
   │(libfuse)│   │(libfuse │   │(nfsserve │
   │         │   │+darwin) │   │  crate)  │
   └─────────┘   └─────────┘   └──────────┘
```

Same shape as FUSE: the library marshals the protocol, a thin native layer forwards callbacks, and the adapter logic lives in Python — `mfusepy` : `MirageFS` :: `mirage-nfs` wheel : `MirageNFS`.

### Python — adapter only, zero protocol code

```
python/mirage/
├── fuse/                      # Existing — unchanged
├── nfs/                       # NEW
│   ├── __init__.py
│   ├── types.py               # attr shapes crossing the Rust boundary
│   ├── errors.py              # STALE + errno→nfsstat3 helpers
[...]

PR #897, opened 2026-08-23
feat(nfs): NFS mount backend — NFSv3 via nfsserve, adapter over the op door

Jackpkn:
Implements #888 — the NFS mount backend, per the maintainer-confirmed design: an existing NFSv3 server library, an adapter with zero protocol code, and no threading in mirage.

## What this adds

A third kernel backend beside `fuse` and `fskit`. The tree is served by a loopback NFSv3 server ([nfsserve](https://github.com/huggingface/nfsserve), BSD-3-Clause) the kernel mounts — no filesystem driver, and on macOS **no admin rights**: loopback NFS mounts are unprivileged (verified live), where macFUSE needs a kext.

```
python/mirage/nfs/       adapter only: ids, writebuf, fs, backend, mount
python/mirage-nfs/       optional wheel (PyO3 + nfsserve): pure marshalling
workspace/nfs.py         NFSManager — one server, many mounts
integ/nfs/               battery in the check_json contract
```

Six commits in review order: wheel → enum → adapter → orchestration/manager → Windows guard → integ battery. Each message carries its full rationale.

## Design highlights (details in #888)

- **No threads in mirage.** The tokio runtime lives inside the extension; trait callbacks bridge onto the workspace loop (`run_coroutine_threadsafe`, ~59µs/call measured). `MirageNFS` is async-native over the `Ops` facade, so every NFS op walks the same door as a shell command — grants, policies, cache, namespace fire once, there.
- **One server, many mounts.** The MOUNT protocol takes a path, so a single server backs any number of kernel mountpoints (`127.0.0.1:/` and `:/docs` simultaneously, verified live) — the one-mount-per-process limit macOS FUSE lives under does not apply.
- **Write buffering** because nfsserve answers every WRITE as durable (`FILE_SYNC` hardcoded, COMMIT not dispatched) and mirage backends store whole objects: unbuffered, a copy is O(n²). Flush on idle, size cap, and teardown; READ/GETATTR overlay pending writes; the bounded durability window is documented, not hidden.
- **fileid ↔ path table**: monotonic, never reused, rename fixes the whole subtree, destination-inside-source refused with EINVAL before anything is touched. State holders are await-free and lock-free; an AST meta-test pins that invariant.
- **MountCore's hard-won rules mirrored and pinned**: link check precedes the ops stat (unguarded, `rm` of a link-to-directory rmdirs the target), macOS metadata names filtered, directory size 0, errors through the shared `classify_error`.

## Verified

- **macOS**: kernel-mount battery green; unprivileged mounts confirmed
- **Linux**: same battery 13/13 in a `--privileged` container with the real kernel NFS client (`mount -t nfs`)
- **Windows**: refused loudly at the guard (`mount.exe` grammar unimplemented; Pro-only client) — the WinFsp-advisory stance
- `integ/nfs/nfs.py` → `truth_nfs.json` via `integ/check_json.py` (18 keys), including the size-unknown limitation pinned as behavior: reads empty, stats 0, mount-time warning fires
- 116 new unit tests (written red-first); full Python suite 16.6k green; layout parity at baseline (python-first exceptions carry their removal condition)

## Deliberately deferred (named in #888)

- CI job wiring for the integ battery (ubuntu runner is the canonical Linux proof)
- TypeScript parity — `@context-fs/nfs` vs a napi binding of the same crate, decided by running this same battery
- The sync `KernelMounts.add` seam (needs an async route; `NFSManager` stands alone until then)
- Session-bound mounts (one server serves one delegate; per-session narrowing needs a second server)

Happy to split this into wheel/adapter PRs if that reviews easier — the commits are already cut along that line.


zechengz:
@codex review

python/mirage/nfs/backend.py
```diff
@@ -0,0 +1,171 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+
+import logging
+import socket
+import sys
+
+from mirage.fuse.backend import require_kernel_backend, resolve_backend
```
zechengz, line 19: Seems that we mix the nfs and the fuse, why we do this?

python/mirage/nfs/fs.py
```diff
@@ -0,0 +1,452 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+
+import errno
+import os
+import posixpath
+
+from mirage.fuse.platform.macos import is_macos_metadata
```
zechengz, line 19: Same here, why we mix the fuse and the nfs?

python/mirage-nfs/src/bridge.rs
```diff
@@ -0,0 +1,131 @@
...
+/// FUSE adapter uses, so a mirage exception is named once in one
+/// place: it classifies typed exceptions (FileNotFoundError and kin)
+/// as well as errno-carrying OSErrors. Anything it cannot name is a
+/// server fault rather than a lie about the file.
+fn to_status(py: Python<'_>, err: PyErr) -> nfsstat3 {
+    let name = err.get_type(py).name().map(|n| n.to_string());
+    if matches!(name.as_deref(), Ok("StaleHandleError")) {
+        return nfsstat3::NFS3ERR_STALE;
+    }
+    let classified: Option<i32> = PyModule::import(py, "mirage.fuse.errors")
```
zechengz, line 102: why fuse?

zechengz (changes requested):
Seems that we are mixing the fuse and nfs
also there are some errors such as every file over an NFS mount is dated 1970-01-01

you may need to test extensively in the integ etc

Jackpkn:
see  #911

PR #900, opened 2026-08-24
docs(permissions): document the permissions system, rename (approval <id>) to (ask <id>)

zechengz:
Adds docs/home/permissions.mdx covering profiles (allow/ask/deny rules, the two axes, paths hide/show, mounts narrowing, script and runtime policies), session binding, the ask lifecycle, the workspace list-asks/allow/deny CLI verbs, and the REST asks door. Wires it into docs.json and the CLI page.

Also renames the agent-facing pending marker from "(approval <id>)" to "(ask <id>)" in both languages so the stderr names the same noun the CLI and REST door use, with all pinned occurrences (tests, integ, examples) updated.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -75,6 +75,7 @@
               "home/cache",
               "home/snapshot",
               "home/observer",
+              "home/permissions",
               "home/policy-engine",
               "home/auth"
             ]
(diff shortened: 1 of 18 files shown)


zechengz:
@codex review

PR #901, opened 2026-08-24
fix(permissions): resolve mutation semantics below a hidden path

zechengz:
Settles what mutating commands do to content hidden by a session's paths.hide, under one rule: a session's mutation may destroy what it cannot see, but may never reveal it.

- mv and native dir-copy refuse with EACCES when relocating a directory would carry a hidden path out of hide coverage ("mv: cannot move 'a' to 'b': Permission denied"). A component pattern like *.tkn never refuses: it follows the name, so the file rides along still hidden.
- cp -r copies only the visible view, silently. The strategy fork now also triggers on hides intersecting an operand, so the native copy can no longer leak the hidden dir name or print the hidden path in stderr.
- rmdir on a directory the session sees as empty succeeds and removes the hidden remnants (a not-empty refusal would leak existence). rm -r keeps deleting hidden content: hide is visibility, a deny rule is protection.
- Enforced on both planes: the command-plane slot wrappers in generic_bind/adapter and the dispatcher for FUSE/guest runtimes, in both languages. Cross-mount mv gets the same refusal through a guard on the generic.

New move_reveals/moveReveals predicate in utils/hidden, unit twins, workspace behavior tests, and 18 battery cases (boxed/tagged sessions) green on both hosts.

### DIFF
--- a/integ/resources/ssh/hidden.json
+++ b/integ/resources/ssh/hidden.json
@@ -1,0 +1,44 @@
+{
+  "cases": [
+    {
+      "id": "ssh_subtree_prep",
+      "seq": 923360,
+      "targets": [
+        "ssh"
+      ],
+      "command": "mkdir -p /data/qsubh/sec && echo k > /data/qsubh/sec/k",
+      "expect": {
+        "exit": 0,
+        "stdout": "",
+        "stderr": ""
+      }
+    },
+    {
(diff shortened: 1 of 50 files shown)


zechengz:
@codex review

zechengz:
@codex review

zechengz:
@codex review

zechengz:
@codex review

zechengz:
@codex review

PR #910, opened 2026-08-25
fix: model /dev/null and /dev/zero as character devices

### LINKED ISSUE #570: Model /dev/null and /dev/zero as character devices
## Problem

Mirage currently models `/dev/null` and `/dev/zero` as RAM files rather than character devices.

`DevFiles.get('/zero')` returns one fixed 1 MiB `Uint8Array`. As a result:

- `cat /dev/zero` terminates after 1 MiB instead of producing data until interrupted.
- `head -c 2M /dev/zero` cannot return the requested 2 MiB.
- `find / -type f` includes `/dev/null` and `/dev/zero`, unlike Unix, because Mirage has no character-device file type and `WorkspaceFS.isFile()` treats every non-directory as a file.
- Example code that recursively reads every regular file can unexpectedly consume and print the synthetic zero buffer.

The fixed 1 MiB buffer is a useful safety limit, but it is observable filesystem behavior rather than an internal streaming chunk.

## Expected behavior

Match Unix device semantics as closely as Mirage's execution model allows:

- Reads from `/dev/null` return EOF; writes are discarded.
- Reads from `/dev/zero` repeatedly return zero-filled bytes according to the consumer's requested size.
- Device nodes are not classified as regular files.
- Unbounded commands such as `cat /dev/zero` are stopped by normal command safeguards, rather than by making the device finite.

Linux reference: https://man7.org/linux/man-pages/man4/zero.4.html

## Suggested direction

Introduce a character-device node type or equivalent device-aware abstraction, then make reads request-sized/streaming. Avoid solving this by increasing the fixed buffer.

## Acceptance criteria

- `find / -type f` excludes `/dev/null` and `/dev/zero`.
- If `find -type c` is supported, it returns both device nodes.
- `head -c 2M /dev/zero | wc -c` reports `2097152`.
- `head -c 17 /dev/zero` returns exactly 17 zero bytes.
- `cat /dev/zero` remains unbounded at the device layer and is terminated or truncated by configured command safeguards.
- Existing `/dev/null` read/write behavior remains covered.
- Python and TypeScript implementations retain parity.

sonhmai:
## Summary

- Model active `/dev/null` and `/dev/zero` entries as logical character devices with Linux-compatible `1:3` and `1:5` device numbers.
- Replace the fixed 1 MiB `/dev/zero` buffer with exact ranged reads and lazy streaming, bounded at the source for unqualified `cat` and `head` calls.
- Update `find`, `stat`, `ls`, `test`, `file`, `du`, Monty, WASM, and VFS projections to classify devices without whole-reading them.
- Preserve `/dev/null` empty reads and dropped writes, plus `rm /dev/null; echo x > /dev/null` recreation as a regular file.
- Ship matching Python and TypeScript behavior.

## Behavior

- `head -c 2M /dev/zero | wc -c` prints `2097152`.
- `find / -type f` excludes both devices; `find / -type c` includes them.
- `stat -c '%F %t %T' /dev/null` prints `character special file 1 3`.
- Whole-file consumers refuse endless `/dev/zero` reads instead of allocating indefinitely.
- Unqualified device streams stop at the 8 MiB safeguard with a truncation notice.

## Scope

- FUSE and FSKit intentionally continue exposing regular inodes so reads reach Mirage instead of being routed by the kernel to host device drivers.
- Archive and Git-specific device-member behavior remains follow-up work.

## Verification

- TypeScript build and core suite: 742 files, 10,122 tests passed.
- Full Python and pre-commit suites passed.

Fixes #570


sonhmai:
@claude review

sonhmai:
@codex review

sonhmai:
@codex review

zechengz (approved):
Thanks.

PR #911, opened 2026-08-25
feat(nfs): TypeScript NFS backend — napi binding of nfsserve, the workspace seam, and soft mounts

Jackpkn:
Implements the TypeScript half of #888 (PR 2 in its plan), plus the workspace seam and the PR 3 hardening items.

**Stacked on #897.** This branch is cut from `feat/nfs-backend`, and that branch lives on my fork rather than here, so GitHub cannot use it as a base — until #897 merges the diff below also carries its commits. Twenty-five commits are new relative to #897:

```
feat(nfs-ts): mirage-nfs-node -- nfsserve behind napi-rs      the addon
feat(nfs-ts): the adapter, mount orchestration, NFSManager    the TS backend
chore(nfs):   package both addon crates for install
fix(nfs):     mount soft, and bound every rung of teardown    <- the host-wedge fix
feat(nfs):    serve a declared backend=nfs mount from the workspace
feat(nfs):    session-scoped mounts, in both languages        <- last capability gap
fix(nfs-ts):  let stop() release node's event loop
test(nfs):    diff the fuse and nfs adapters case by case     adapter differential
test(nfs):    diff the two kernel mounts of one tree          kernel differential
test(nfs):    batteries that cannot leave a mount behind
ci(nfs):      both batteries on macOS (in the gate) and linux (advisory)
docs(nfs):    setup pages, examples, and the CLAUDE.md conventions
```

Each message carries its full rationale; `git log` reads as the story.

## What this adds

**The TypeScript backend.** `@context-fs/nfs` 0.1.4 was evaluated first and cannot reach parity: MKDIR, REMOVE, RMDIR, RENAME and SYMLINK all fall through to `PROC_UNAVAIL` and its provider interface carries no methods for them, so `mkdir`, `rm`, `mv` and `ln -s` through a mount are impossible and the existing battery fails it on five keys by construction. So the server is a napi-rs binding of the same nfsserve crate — one protocol implementation behind both languages, exactly as libfuse sits behind both FUSE adapters. Both `Cargo.lock`s pin `nfsserve 0.11.0` with the same checksum, so the two provably build the same crate.

```
typescript/packages/mirage-nfs/       napi addon: pure marshalling, mirrors the PyO3 one
typescript/packages/node/src/nfs/     adapter: ids, writebuf, fs, backend, mount
typescript/packages/node/src/workspace/nfs.ts   NFSManager — one server, many mounts
```

**The workspace seam, in both languages.** `NFSManager` stood alone; a declared `Mount(backend=nfs)` had nowhere to land, because the constructor that reads mount specs is synchronous and the server needs a running loop. The declaration is recorded there and mounted on the first await, `execute` awaits it, and `KernelMounts` now tracks which backend serves which prefix so the fuse and nfs views stay separate. `close()` unmounts the nfs tier first: unmounting makes the kernel client flush its dirty pages as final WRITEs, which need both a live server and the resources they land in.

**Session-scoped mounts.** The FUSE tier takes a session and runs every op under its mount grants; nfs refused one outright, and it was the last capability gap between the two kernel tiers. A scoped mount gets its own server — one server serves one delegate, so narrowing cannot reuse the unscoped one — and `KernelMounts` keeps a manager per session, so a second scoped prefix of the same session reuses it. The scoping wraps the delegate *entry points* rather than the op call sites inside the adapter: one that bound sixteen of seventeen call sites would still serve the seventeenth with the workspace's full reach, and nothing at the missing site would look wrong. A meta-test fails if the adapter grows a method the bound list does not name.

## The hard-mount fix (please read this one)

Running the batteries locally **froze the machine**, and the cause was in the library rather than the test. mirage mounted NFS **hard**, the kernel's default, where an I/O whose server stopped answering blocks forever and uninterruptibly — and since macOS walks the mount table for Finder, `df` and Spotlight, one wedged mount takes the desktop with it rather than just the caller. The server *is* the process that asked for the mount, so its own deadlock is exactly that case.

- Mounts are now `soft,timeo=50,retrans=3`, plus `intr` and `deadtimeout=60` on darwin, all off `NFSConfig`; `soft=False` still expressible.
- Teardown escalates in four bounded rungs — plain, a retry for EBUSY, `umount -f`, then `umount -l` on linux or `diskutil unmount force` on darwin — and names the command that clears the mountpoint if none of them do. It previously ran one plain unmount and waited forever, which is unrunnable on the one mount that needs it: a plain unmount asks the filesystem to flush, and the server that would answer is the one being torn down. There was no `umount -f` anywhere.
- The readiness probe had a deadline it could never reach: a stat of a wedged mount does not return, so a timeout checked only between probes never fired. Each probe is bounded now.
- Both batteries record every mountpoint, force-unmount from a SIGINT/SIGTERM handler and a top-level `finally`, run under an overall deadline, and no longer stat their own mountpoints — that call is answered by the server that just stopped if the unmount failed, so the one check that only ran after something went wrong was the one that hung.

## The differential the design asked for

#888's PR 1b scope names a differential-vs-FUSE run, and nothing compared the two adapters. There are now two, at different altitudes.

**At the adapter**, both are driven over identically seeded workspaces and diffed case by case — attrs, errno for a missing path, readdir names, whole and offset reads, a write read back before it is stored, growth past EOF, symlink, mkdir, rename, unlink, truncate — 16 cases in each language, as ordinary unit tests so they gate on every PR without a kernel or privileges.

**At the kernel**, `integ/nfs/differential.py` mounts one tree over FUSE *and* over NFS and runs the same sixteen shell lines against each mountpoint — reads whole and sliced, an md5 over 4 KiB, `ls`, `find`, stat size and type, a symlink read and followed, both error paths, and a write read back per side. They match on every line. It lives in the linux job because that is the only runner where both kernel tiers exist at once.

The adapter-level pair corrected the record on the size-unknown limitation: both adapters answer a READ with the real bytes. The divergence is that FUSE hydrates on OPEN and can restate the size afterwards, where NFSv3 has no OPEN, so the size stays 0 and the client never asks. Same observable behavior through a real mount, different mechanism — and the tests pin the divergence set, so if the two ever agree the mount-time warning and its docs should go with it.

## Verified

- **Both batteries, both platforms, in CI and locally**: 32 keys each against the shared `truth_nfs.json`, on a hosted `macos-15` runner and an `ubuntu-latest` one, with no mount and no scratch directory left behind. The TS battery runs through `NFSManager` over a real kernel mount (read, write, mkdir, mv, symlink, readlink, rm, ENOENT, ENOTEMPTY, size, a 1 MiB md5 round-trip).
- **Session scoping proven on the wire**, not at the adapter: a read-only session reads its file through the scoped mountpoint and is refused the write with `NFS3ERR_ACCES`, while the same write through the unscoped mountpoint of the same tree succeeds.
- **The kernel differential**: FUSE and NFS answer all sixteen lines identically.
- Full python suite green; 170 node tests across the NFS suites and 295 with the FUSE ones; `pnpm typecheck`, eslint, knip, mypy and `pre-commit run --all-files` clean.
- Layout parity at the committed baseline — the `nfs` python-only exception is gone, which was PR 2's stated condition.
- `integ-nfs-macos` is **in the merge gate**. `integ-nfs-linux` is advisory on purpose: it installs packages with apt, and a gate that fails on a mirror hiccup is worse than an honest advisory job.

## What is deliberately not here

- **Prebuilt artifacts.** The wheel is not on PyPI and the addon has no per-platform npm packages, so there is no `mirage-ai[nfs]` extra yet — an extra is resolved at lock time and would fail every `uv sync` in the repo for a wheel that does not exist. Today it is `pip install ./python/mirage-nfs`. This is the one item from your original comment not delivered, and it needs a publishing decision rather than more code: happy to add the napi platform matrix and a release workflow if you want it in this PR.

Everything else that was deferred earlier in this PR's life has landed: session-bound mounts, the kernel-level differential, the event-loop release on `stop()`, and the macOS job's promotion into the gate.

## Open questions from #888, answered

3. **Dependency pin** — crates.io, and both lockfiles already resolve `0.11.0` with the same checksum, so builds are reproducible and the two languages share one crate build.
4. **sudo per mount** — macOS loopback NFS needs **no** privileges (verified live); Linux does. #888 currently says both platforms need sudo; the docs here say what is true.


zechengz:
@codex review

zechengz (commented):
Ran the branch locally: 376 Python tests green, layout parity back at baseline 253, the `nfs` python-only exception gone. The big things from the last round are genuinely fixed — `KernelRoute` with an exhaustive table (a backend with no row raises instead of silently inheriting FUSE's route), `MountCore` moved to `mirage/mount/` and flipped async with zero `_run()` calls left, `mtime_epoch` a real field, and `integ-nfs-macos` in the merge gate as the only job in the repo that runs cargo. The soft-mount/teardown work is a serious catch. The two differentials are better than what I asked for.

Below is what I found that Codex's four comments don't cover. Items 1–3 are the Python twins of TS bugs Codex flagged; 4–9 are new.

---

### 1. `create_exclusive` destroys an existing file — Python too, not just TS

Codex flagged this on `typescript/packages/mirage-nfs/src/vfs.rs:215`. `python/mirage-nfs/src/vfs.rs:190` has the identical mapping, and I confirmed the data loss:

```
before: b'important data\n'
after : b''          # should have been NFS3ERR_EXIST, file untouched
```

**Flow:** `open("/mnt/keep.txt", O_CREAT|O_EXCL)` → NFSv3 CREATE mode EXCLUSIVE → `create_exclusive` → delegate `create` → `MountCore.create` truncates. Any lockfile idiom (`flock`-style `O_EXCL` sentinels, `pip`, `git index.lock`) silently wipes the file it was supposed to refuse to touch.

The comment in place says "the adapter's create is idempotent for an empty file" — that's true for an empty file and false for every other one, which is the case that matters.

### 2. Flush loses acknowledged writes when the store fails — Python too

Codex flagged `delegate.ts:350`. `python/mirage/nfs/delegate.py::_flush_one` is the same shape:

```python
async with self._flush_lock(fileid):
    base = await self._read_base(path)
    pending = self._writes.take(fileid)      # <- removed unconditionally
    if not pending:
        return
    await self._ops.write(path, WriteBuffer.merge(base, pending))   # <- may raise
```

The per-fileid lock is a good addition and closes the concurrent-flush race, but it doesn't help here.

**Flow:** client writes 4 MB → server answers `FILE_SYNC` (durable) for every chunk → idle timer fires → `take()` empties the buffer → `ops.write` fails (S3 throttle, policy denial, expired token) → the exception surfaces, the bytes are gone, and no later flush can recover them. The client was told those writes were durable.

Requeue on failure rather than taking up front.

### 3. `remove` drops the buffer before the removal succeeds — Python too

Codex flagged `delegate.ts:157`. Python `remove()` does the same, in this order:

```python
if fileid is not None:
    self._writes.drop(fileid)          # <- line 15
...
attrs = await self._core.attrs_for(path)   # <- may raise (denied, transient)
if S_ISDIR(attrs.mode): await self._core.rmdir(path)
else:                   await self._core.unlink(path)
```

**Flow:** client writes to `/x`, gets `FILE_SYNC` → client sends REMOVE → buffer dropped → `rmdir` fails with ENOTEMPTY (or unlink is denied by policy) → the file still exists, but now with its pre-write bytes, and the acknowledged writes are gone. A failed delete silently rolled back a successful write.

### 4. Reads are still 16× amplified, and the shared cache can't help

Unchanged from #897. Measured on this branch — 16 × 64 KiB READs of a 1 MiB file:

```
backend calls : {'read': 16}
bytes fetched : 16777216      # 16 MiB to serve 1 MiB
```

The interesting part is *why*, given `PrefetchCache` now lives in the shared `mirage/mount/`:

```
$ grep -n "_prefetch.put" mirage/mount/core.py
254:        self._prefetch.put(path, data)      # inside prefetch_read()
```

One fill site, inside `prefetch_read()`, called from `MountCore.open()`. **NFSv3 has no OPEN**, so NFS never reaches it:

```
delegate.read(fileid, offset, count)
  → _read_base(path)
  → core.read(path, _WHOLE_FILE, 0, fh=None)
        ctx            = self._ctx(None)      → None    (no handle)
        cached_data(p)                        → None    (never filled for NFS)
        await self._ops.read(...)             → WHOLE FILE, every single READ
```

So the cache was moved into shared code but stayed keyed to a lifecycle event only one protocol has. On a 1 GiB file that's ~16,000 full-file fetches; on an API-backed mount it's one request per 64 KiB.

Fix looks small: fill the cache in `core.read` on a `cached_data` miss, not only in `prefetch_read`. `core.read` already has the bytes in hand.

### 5. `readdir` silently truncates when the cookie entry is deleted

Unchanged from #897. Reproduced on this branch:

```
page1: [('a', 3), ('b', 4)]
page2 after removing 'b': []        # expected ['c', 'e', 'f']
```

**Flow:** resume scans the sorted names for `fileid == cookie` and `continue`s until it matches. If that entry is gone, `resuming` never flips false, the loop skips everything, and the empty list reads to the client as end-of-directory. So `ls` on a directory another writer is touching drops the tail of the listing with no error anywhere.

Suggest falling back to name order when the cookie's fileid isn't found.

### 6. `NFSManager` leaks a live kernel mount on a duplicate prefix

Unchanged from #897, and it matters more now that the teardown work exists.

```
two setups of the SAME prefix -> [('mount','/tmp/mp-a','/docs'), ('mount','/tmp/mp-b','/docs')]
registry: {'/docs': '/tmp/mp-b'}
unmounted on close: [('umount','/tmp/mp-b')]     # mp-a stays mounted forever
```

**Flow:** the collision check covers the mountpoint but not the prefix, then `self._mounts[prefix] = (...)` overwrites. `close()` iterates the registry, so `/tmp/mp-a` is never unmounted — leaving exactly the "live mount with no server behind it" that the soft-mount commit exists to prevent.

### 7. `lookup` accepts a name containing `/`, and ENOENTs `.` and `..`

```
lookup('..')        -> FileNotFoundError
lookup('.')         -> FileNotFoundError
lookup('sub/b.txt') -> OK, path='/docs/sub/b.txt'
```

`filename3` is a single component — a `/` in it is a protocol violation, and nfsserve doesn't filter it, so the delegate is the only guard. Traversal doesn't escape today only because nothing normalizes `..`, which is luck rather than a check. Worth rejecting `/`, `.` and `..` explicitly in `lookup`/`create`/`mkdir`/`symlink`/`remove`/`rename`.

Separately, `.` and `..` are the server's job over NFS (the kernel resolves them above the filesystem for FUSE, which is why `MountCore` never had to). Returning ENOENT is a latent gap for cold-cache `..` lookups.

### 8. `max_buffered_bytes` has no global ceiling

It's per-handle (16 MiB default), so N files being written concurrently is N × 16 MiB with nothing bounding the total. `flush_idle` only sweeps on its timer. A single `cp -r` of many large files can grow the process without limit.

### 9. `_read_base`: the TS twin is broader than the Python one

Complements Codex's `delegate.ts:358` comment. The two sides are not the same code — Python is narrow and correct, TS is a bare catch:

```python
# python/mirage/nfs/delegate.py
except (FileNotFoundError, IsADirectoryError):
    return b""
```
```ts
// typescript/packages/node/src/nfs/delegate.ts:353
  } catch {
    // missing file or a directory: the write starts from empty
    return Buffer.alloc(0)
  }
```

The TS comment names two conditions the catch doesn't actually restrict itself to, and a bare `catch` is the "never silently swallow exceptions" rule in CLAUDE.md. So the fix is to narrow TS to Python's set, not to invent a new policy — and a parity test on the two catch sets would keep them from drifting again.

---

### One doc nit

The "One departure from your sketch" section says `MirageNFS` is *"16 methods over the `Ops` facade rather than `MountCore`"*. That's stale — the delegate now makes 14 `self._core.*` calls against 1 direct `ops` call, so it **is** over `MountCore`. The paragraph asks me to confirm a departure that didn't end up happening; suggest dropping it.

### On the shape of 1–3

Codex found those three in TypeScript and they're all in Python as well, which is worth noting beyond the individual fixes: the parity rule means a bug found on one side should be checked on the other by default. A cheap guard here is to extend the existing adapter differential with failure cases — a backend whose `write` raises, a `remove` that's denied, an `O_EXCL` create on an existing path — since those are exactly the paths both languages got wrong and the current 16 cases are all happy-path.

PR #912, opened 2026-08-25
fix(shell): boot the tree-sitter engine once

sonhmai:
## Problem

`createShellParser` called `Parser.init` every time. That call boots one wasm
module for the whole process, so two callers that start together race it. The
second one reads the language out of a half-built module and throws:

```
Incompatible language version 0. Compatibility range 13 through 15.
```

Any program that creates two workspaces at the same time can hit this, because
each workspace builds its own parser.

The test fixture already worked around it with a memo in
`workspace_fixture.ts` (`parserPromise ??= createShellParser(...)`), which
shows the guard belongs in the function itself.

## Fix

Every caller awaits the same boot. A failed boot is not cached, so one bad
start does not poison every later parser.

Python does not have this bug. It builds `BASH_LANGUAGE` and `TS_PARSER` once
at import in `python/mirage/shell/parse.py`, so this change makes TypeScript
match it.

## Test

`parse_boot.test.ts` creates 8 parsers at once and parses with each. It is its
own file on purpose: the boot happens once per module graph, so a test sharing
a file with an already booted parser cannot see the race.

Checked it fails first:

```
FAIL src/shell/parse_boot.test.ts
 ❯ Parser.setLanguage web-tree-sitter/src/parser.ts:158:15
 ❯ createShellParser src/shell/parse.ts:168:10
Test Files  1 failed (1)
```

After the change:

```
Test Files  20 passed (20)
     Tests  528 passed (528)      # packages/core src/shell
```

Also run: full `packages/core` suite, prettier, eslint, `tsc --noEmit`.

### DIFF
--- a/typescript/packages/core/src/shell/parse.ts
+++ b/typescript/packages/core/src/shell/parse.ts
@@ -161,8 +161,21 @@ export function findUnterminatedBacktick(command: string): string | null {
   return opened !== null ? command.slice(opened) : null
 }
 
+// `Parser.init` boots one wasm module for the whole process, so two callers
+// that start at the same time used to race it: the second read the language
+// out of a half-built module and threw "Incompatible language version 0".
+// Every caller now awaits the same boot. A failed boot is not kept, or one bad
+// start would poison every later parser.
+let engineBoot: Promise<void> | null = null
+
 export async function createShellParser(config: ShellParserConfig): Promise<ShellParser> {
-  await Parser.init({ wasmBinary: toArrayBuffer(config.engineWasm) })
+  engineBoot ??= Parser.init({ wasmBinary: toArrayBuffer(config.engineWasm) }).catch(
+    (err: unknown) => {
+      engineBoot = null
+      throw err
+    },
+  )
+  await engineBoot
   const language = await Language.load(toUint8(config.grammarWasm))
   const parser = new Parser()
   parser.setLanguage(language)
--- a/typescript/packages/core/src/shell/parse_boot.test.ts
+++ b/typescript/packages/core/src/shell/parse_boot.test.ts
@@ -1,0 +1,39 @@
+// ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


sonhmai:
@codex review

zechengz:
Reproduced the race and benchmarked the fix. Numbers below are Node 24.15.0, medians of 3 runs, parsing `echo hello | grep x && cd /tmp`.

**The bug is real.** On `main`, 100 parsers created with `Promise.all` fail exactly as described:

```
{"mode":"concurrent","n":100,"ok":false,"ms":47.2,
 "error":"Incompatible language version 0. Compatibility range 13 through 15."}
```

With the patch applied, all 100 build and parse.

**Cost, with the fix applied:**

| shape | time | RSS | wasm |
| --- | --- | --- | --- |
| 100 parsers, concurrent | 47.7 ms | +259 MB | +116 MB |
| 100 parsers, sequential | 75.7 ms | +139 MB | +116 MB |
| 1 parser, 100 parses | 10.8 ms | +16 MB | +33 MB |

Scaling: 1 parser is 9.6 ms / +33 MB (essentially all fixed cost, the one-time engine boot), 10 parsers is 13.3 ms / +29 MB RSS, 100 is 47.7 ms / +259 MB. So the marginal parser is about **0.4 ms and 2.5 MB**.

**Worth being explicit about in the description:** the guard dedupes the engine boot only. `Language.load(grammarWasm)` and `new Parser()` still run per call, and that is where the memory goes, so N parsers still cost roughly N x 2.5 MB after this change. The fix makes concurrent creation *correct*, not *cheap*, which I think is the right scope, but the two are easy to conflate.

**One overstatement:** "Any program that creates two workspaces at the same time can hit this" is stronger than what the code does today. Both `packages/node/src/workspace.ts:36-47` and `packages/browser/src/workspace.ts:32-39` memoize with `cachedParser`, assigned synchronously before any await, so two workspaces built back to back in one process already share a single boot and are not exposed. The genuine exposure is code calling the exported `createShellParser` directly and concurrently, plus the test fixture. That is still worth fixing, and the fixture is the strongest evidence for it: `workspace/fixtures/workspace_fixture.ts:27-31` already carries a `parserPromise ??=` memo, which is this same guard written one layer too high. That workaround can be deleted once this lands.

The `.catch` that clears `engineBoot` is the detail most promise memoizations get wrong, good to see it here.

**Possible follow-up, not for this PR:** memoizing `Language.load` per grammar would take the 100-parser case from +259 MB toward the +16 MB row, since the language is immutable. I have not checked whether web-tree-sitter supports sharing one `Language` across `Parser` instances, so that needs verifying before anyone acts on it.

zechengz (approved):
Thanks! This is quite important

PR #914, opened 2026-08-25
perf(integ): run targets concurrently and report where the time goes

sonhmai:
Runs the declarative integ battery's targets concurrently, and reports where
its time actually goes.

## Concurrency

`--target-jobs N` runs targets in parallel. Targets own separate workspaces, so
they overlap safely, except when two speak to the same fake (`cli-gh` and
`github`, `cli-ntn` and `notion`, `qdrant` and `qdrant-window`). Those share a
lane and stay sequential; the rest are bound only by the overall width.

The default stays **strictly serial**, on the old loop. The lane scheduler
cannot stand in for it: a target waiting on a busy lane lets a later one from
another lane take the worker first, and with `s3` carrying three targets and
`gws` four, the default run would quietly reorder itself.

Each invocation gets its own result slot. Keyed by target id, `--target x
--target x --target-jobs 2` made both tasks write the second slot and merged it
twice, reporting 11700 cases where a serial run reports 5850.

## Profile

`--profile` prints three tables: per target, the slowest cases, and the
costliest commands. Several things had to be right for those numbers to mean
anything, and each one was wrong first:

- **Ranked by wall time**, the column the table is about. Ranking by summed
  case time buried a target whose cost is setup.
- **`wall` beside `in cases`**, because the gap between them is setup and
  teardown, and that is often where the time is.
- **A scenario is charged to itself**, not to its first step. A consistency
  case runs out-of-band remote writes and more than one invocation. Billing
  that to the first verb made `cat` and `find` carry time neither spent: on s3
  against local moto, `cat` read 63% costlier than it was and `find` 49%.
- **The sample is the whole case**, not the command alone. `run_case` freezes
  `elapsed` before awaiting `stat_check`, so the 34 cases carrying a `check`
  contributed a sample missing it, and that work fell into the setup gap.
  `elapsed` is still the command alone, because three cases assert
  `expect.elapsed` bounds against it.
- **Commands are named, not guessed at.** `out=$(cat ...)` used to label the
  row with a filename, and `(cd d && grep x)` with `(cd`. Rows whose verb is
  not a command name went from 254 corpus lines to 64, and 47 of those 64 are
  `[` and `[[`, which are the test builtin and already right.

Both hosts return identical verbs for all 5043 command lines in the corpus, and
pick the same percentile index.

## Scope

The typescript runner's option parser is **not** here. It grew out of one bug
in this PR's own new option and turned into an argparse-parity project, so it
moved to #952. What stays is the narrow fix: `--target-jobs` is recognised
before its operand is checked, so a trailing one is refused rather than running
serially without a word.

### DIFF
--- a/.github/workflows/test_integ.yml
+++ b/.github/workflows/test_integ.yml
@@ -279,7 +279,7 @@ jobs:
 
       - name: Run declarative battery on Nextcloud
         working-directory: integ
-        run: pnpm exec tsx runners/typescript/main.ts --target nextcloud --strict
+        run: pnpm exec tsx runners/typescript/main.ts --target nextcloud --strict --profile
 
       - name: Start MinIO
         run: |
@@ -434,7 +434,7 @@ jobs:
             -p qjs-wasi.wasm -O /tmp/quickjs/qjs-wasi.wasm
 
       - name: Run declarative battery (python hosts)
-        run: ./python/.venv/bin/python integ/runners/python/main.py --facet core --strict --allow-skip chroma,lancedb,nextcloud,notion,postgres,qdrant
+        run: ./python/.venv/bin/python integ/runners/python/main.py --facet core --strict --target-jobs 4 --profile --allow-skip chroma,lancedb,nextcloud,notion,postgres,qdrant
 
       # Moved here from the `integ` job when the HTTP fixture server became a
       # TypeScript kit fake: the adapter still starts it, but the interpreter is
@@ -518,7 +518,7 @@ jobs:
 
       - name: Run declarative battery (typescript hosts)
         working-directory: integ
-        run: pnpm exec tsx runners/typescript/main.ts --facet core --strict --allow-skip chroma,lancedb,nextcloud,notion,postgres,qdrant
+        run: pnpm exec tsx runners/typescript/main.ts --facet core --strict --target-jobs 4 --profile --allow-skip chroma,lancedb,nextcloud,notion,postgres,qdrant
 
       # The fixture server is python (started by the adapter through
       # startPythonServer), so this runs here rather than in integ-ts, which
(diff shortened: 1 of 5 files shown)


sonhmai:
@codex review

sonhmai:
Fixed. Real py/ts divergence and it would have made the profile numbers disagree between hosts for no reason.

Python's `round()` rounds a half to even, JS `Math.round` rounds it up. Python now uses `floor(x + 0.5)`, which is what `Math.round` does for a positive number. Checked the case named in the comment: the 2666-sample target picks index 1333 on both sides now, where python used to pick 1332.

@codex review

sonhmai:
Right, fixed in 8b9600d53 and thread resolved.

Both runners passed `elapsed=0` for a consistency scenario. That was harmless while the argument only fed the expectation check, but it is now also a profile sample, so those cases counted as free and dragged the totals and percentiles down. Both hosts measure the scenario now.

Checked before changing it: no consistency case declares elapsed bounds, so no assertion moves.

@codex review

sonhmai:
@codex review

sonhmai:
@codex review

sonhmai:
Real, and fixed as far as is proportionate. Both hosts now skip a leading assignment whose value is quoted or a parenthesised array, plus any separator after it, so `v='a b'; cat ...` reports `[` rather than the fragment `b';`.

Deliberately not shell-aware tokenisation. This string only labels a row in the profile, so a wrong guess costs a mislabelled row and never a wrong measurement. The change takes the corpus from 56 mislabelled lines to 39, and reading the remaining shapes would mean carrying a tokeniser in two languages for a label.

@codex review

sonhmai:
@codex review

sonhmai:
@codex review

sonhmai:
@codex review

sonhmai:
@codex review

sonhmai:
@codex review

sonhmai:
@codex review

sonhmai:
@codex review

sonhmai:
@codex review

sonhmai:
@codex review

sonhmai:
@codex review

sonhmai:
@codex review

zechengz:
@sonhmai It's ok to ignore some codex comments (for example if detected ones have no big issue)

sonhmai:
@codex review

sonhmai:
this has fulfilled its purpose so closing it now.

Issue #886, opened 2026-08-22
CI perf: cut duplicate setup/build work and shrink test wall time


sonhmai:
CI takes too much time in setup.

As of August 22, 2026, recent green runs are roughly:

- `Integ`: 30-33m
- `Test (TypeScript)`: ~17m
- `CLI exit codes`: ~17m
- `Test (Python)`: ~12m
- `Pre-commit`: ~9m

What looks wrong:

- same Python sync repeated everywhere
- same `pnpm install` repeated everywhere
- same TS package builds repeated everywhere
- `Integ` repeats full Python + Node + build bootstrap across a lot of jobs and the 6-way facet matrix
- TS CLI e2e is a whale: `typescript/packages/cli/src/e2e.test.ts` took about 217s by itself in a recent green TS run
- Python test lane does main pytest, truth checks, then resyncs for Camel and runs pytest again

Current shape:

- `uv sync --all-extras --no-extra camel` repeated in 12 places
- `pnpm install` repeated in 14 places
- TS builds repeated in 23 places

Some suggestion:

1. Bootstrap once per workflow and fan out artifacts in `Integ` and `CLI` instead of reinstalling and rebuilding in each job.
2. Split TS CLI e2e out of the normal TS unit lane. Gate it to CLI/daemon/workspace changes. Shard if still too slow.
3. Stop rebuilding TS packages in every workflow. Reuse built `dist/` or packed artifacts where possible.
4. Fix `integ-facets` matrix amplification. Either collapse low-cost facets or feed the matrix from prebuilt artifacts.
5. Move Camel into its own Python job with path filters, instead of mutating the main Python test env mid-job.
6. Use slimmer Python extras for non-integration jobs. `all-extras` is overkill for pre-commit and smoke lanes.

Non-goals:

- no test coverage cuts just to make CI green faster
- no behavior changes in Mirage itself

Success criteria:

- `Integ` down to <20m wall clock
- `Test (TypeScript)` down to <12m
- `CLI exit codes` down to <10m
- same coverage, same assertions, less duplicate bootstrap

zechengz:
sg, this is a real painpoint😢

sonhmai:
https://github.com/strukto-ai/mirage/pull/889 bootstrap TypeScript once and reuse dist artifacts in CLI + integ
https://github.com/strukto-ai/mirage/pull/892 run main Python pytest suite with xdist

sonhmai:
## PRs so far

Merged:

[...]

PR #917, opened 2026-08-26
perf(ci): cache mypy so the test job stops re-checking the package from cold

sonhmai:
Part of #886. Stacks on #918 (example concurrency + worksteal); once that
merges this reduces to the mypy cache alone.

## Result, measured on real CI

`Test (Python)` -> `test` job, against a 14-run baseline on main:

| | value |
|---|---|
| baseline median (n=14) | **737s** (12.28m) |
| baseline minimum | 497s |
| with these changes | **394s** (6.57m) |
| **delta** | **-343s, -46.5%** |

394s is **below every one of the 14 baseline runs**, so this is not a draw from
the same distribution.

## Where it comes from

| step | before | after | delta |
|---|---|---|---|
| Type check (mypy) | 100s | **3s** | -97s |
| pytest main suite | 392s | 282s | -110s |
| Examples | 63s | 32s | -31s |
| Initialize containers | 30s | 39s | +9s |

The mypy line is this PR: `.mypy_cache` was never persisted, so every run paid a
cold full type check of the package. Cached and keyed on `uv.lock` plus the
source tree, with restore-key fallbacks so a source change still reuses most of
the store. On a cold cache it saves nothing (measured: 87s); warm it is 3s.

pytest and Examples are #918.

## Coverage is not compromised

| | baseline run | this run |
|---|---|---|
| main suite | 17436 passed, 116 skipped, 1 xfailed | **17529 passed, 116 skipped, 1 xfailed** |
| redis serial | 273 passed | 273 passed |
| camel | 21 passed | 21 passed |

More tests pass than at baseline, because main moved forward. Nothing is
removed, skipped, deselected, or marked xfail. Every change here schedules the
same work differently or reuses a cache; none of them touch test selection.

## What was tried and rejected

- `-n 8` on the 4 vCPU runner: worse than `-n 4` (437s vs 398s). The runner is
  already saturated.
- `--dist loadfile`: worst of the four (472s median).
- A local laptop benchmark of the same options: produced the opposite ranking
  and was discarded. Run sequentially while machine load varied, the *same*
  config measured 445s and 206s. All numbers above come from concurrent sibling
  jobs at one commit instead.


zechengz:
@codex review

PR #925, opened 2026-08-27
Move slack, notion and mem0 onto the integ kit and scope /reset by tenant

zechengz:
Moves slack, notion and mem0 onto the integ kit, and makes /reset tenant-scoped
so two hosts can share one fake server.

## Tenant-scoped reset

/reset used to recreate the whole run SQLite file. No adapter sends a run id, so
both hosts land on run `default`: the second host's reset unlinked the database
the first was reading, and every later request on that run failed. A fake with a
tenant column can express the same reset as a delete restricted to the named
tenants, which is what it now does.

- `clear.ts` deletes by tenant, driven by the DMMF so a new model is cleared
  without anyone remembering to say so. `deleteOrder` puts FK holders first:
  `relationMode = "prisma"` moves referential integrity into the client, so
  declaration order hit `violate the required relation 'BoardToOwner'`.
- A model with no tenant column cannot be cleared per tenant, so `clearTenants`
  refuses that schema by name rather than leaving rows for the next seed.
- The clock and the minter moved from per-run to per-tenant. They had to: a
  second host's reset cleared the counter the first was minting from, and its
  next write failed on a duplicate id.
- The fixture is read before anything is deleted, so a bad fixture name is a 400
  that leaves the caller's rows alone.
- /reset reports `scoped`, so a caller can see which it got.

Proven both ways in the selftest (47 checks) and end to end: the same two-host
concurrent battery is 110/0 on each host with the scoped reset, and 9 failures
plus a dead TS run without it.

## slack, notion, mem0

- `server/slack.ts` (829 lines) and `server/notion_server.ts` (1884 lines)
  become module trees under `server/slack/` and `server/notion/`, each on its
  own prisma schema with `tenant` in every `@@id`. `prisma/schema.prisma`, the
  last shared schema, is gone.
- `tenantTokenPattern` lets one workspace be reached by two token spellings:
  slack uses `xoxb-` for most methods and `xoxp-` for search.*, and the raw
  bearer would have split one workspace across two tenants.
- Notion's `workspace_name` was the bearer token echoed back. It is seeded on a
  new `NotionMeta` model instead, which also absorbs the fixture's `defaults`
  block (those timestamps are read at create time, not only at seed time).
- mem0 moves off aiohttp; its memories and scores were module constants and are
  now a fixture. The kit router matches a route exactly, so the trailing-slash
  URLs the mem0 client sends are declared rather than absorbed by an rstrip.

## Verification

- notion: 30/30 responses byte-identical against the old server, across both API
  versions, writes, minted ids, markdown round-trips and delete
- mem0: 15/15 identical against the python fake
- batteries: slack 89+21, notion 119, mem0 56, all 0 failed on both hosts
- notion MCP/REST parity 33/33
- ntn conformance against the real 0.21.9 binary: 74 checked, 2 not comparable
  (the two cases whose line reads a /notion mount path)

## Notes

- The notion token stays fixed. It is the tenant channel AND observable: `ntn
  auth token` prints it without contacting the server, integ/cli/ntn.json pins
  that literal, and the conformance runner asserts the same line against the
  real binary with the same fixed token. Per-run tokens would make those two
  runs print different things with one golden between them.
- `http_server.py` stays python. Its CI job is python-only, and migrating a
  stateless fake would pull the node toolchain into it.
- CI: the slack and trello setup steps still ran `prisma generate --schema
  prisma/schema.prisma`, the file this PR deletes; that step would have failed
  on every run.


zechengz:
@codex review

PR #926, opened 2026-08-27
Fix orphaned $var expansion in redirect targets and forward record through the daemon

zechengz:
## Summary
- tree-sitter-bash 0.25.1 orphans a later unbraced `$var` when a name-terminating character follows it, so `echo hi > /api/$c/$id.json` wrote a file literally named `$` (bare words, for lists, arrays and assignments split the same way). `parse()` now rebraces the orphaned expansion (`$id` -> `${id}`) and reparses, in both languages.
- The daemon's execute route dropped `ExecuteOptions.record`, so an unrecorded run was unreachable over HTTP. Both routes now forward it.

## Test plan
- New parse-level and workspace tests for the repair (py + ts), including literal-dollar guards.
- New route tests: `record: false` leaves no history entry (py + ts), plus a CLI e2e over a live daemon.
- New integ corpus `integ/bash/redirect/vartarget.json`; ram+disk battery green on both hosts (5808 each).
- Full pytest, core/server/cli vitest, `pnpm -r typecheck`, pre-commit all green.

### DIFF
--- a/integ/bash/redirect/vartarget.json
+++ b/integ/bash/redirect/vartarget.json
@@ -1,0 +1,132 @@
+{
+  "cases": [
+    {
+      "id": "rv_second_var_suffix",
+      "seq": 603220,
+      "targets": [
+        "ram",
+        "disk",
+        "redis",
+        "opfs",
+        "s3",
+        "databricks",
+        "gridfs",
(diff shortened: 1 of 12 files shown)


zechengz:
@codex review

PR #927, opened 2026-08-27
Move the hf buckets fake onto the integ kit

zechengz:
Tier 4 of the integ mock server unification. `hf_server.py` was the last storage fake on the old aiohttp path; it becomes `integ/server/hf/` plus `integ/prisma/hf.prisma`, the same shape as box, dropbox and onedrive.

Both hosts share one external server, so the python host no longer starts its own in process. The env var is `HF_URL`, since the kit derives the announce token from the service name.

22 files changed, 1110 insertions, 601 deletions.

## Isolation

Each run takes its own account on one shared server, carried by the token the client already sends on every Hub call, replacing the old `integ/<runid>-<mount>` bucket name that isolated runs only as far as a name collision.

The Xet CAS is the exception and is deliberately **untenanted**. Content addressed by hash is global on the real Hub, and the client relies on that: it caches uploaded shards locally and, for content it finds there, commits without re-uploading the blocks. Both hosts run against one server on one machine and share that cache, so filing blocks per account left whichever host wrote a fixture second committing a file whose bytes the fake had filed under the other. Those tables carry a constant `tenant` rather than no column, because a model with no tenant blocks a scoped `/reset` outright.

## Xet, and why there is a new dependency

`server/hf/xet.ts` is the binary half: xorb decode, MDB shard parse, and the merkle hash hex spelling (four little-endian u64 words, so the raw bytes and the hex that names them differ in order).

OpenDAL's writer emits compression scheme 0 and scheme 1, probed, so a frame decoder is not optional. `lz4js` (ISC) is a new integ dependency, checked byte for byte against python's `lz4.frame.compress` on 12 B, 100 KB, 1 MB and 70 KB of random data. Python's `lz4` dev dependency goes away with the fake that needed it.

## Two bugs only a shared server can show

**The xet client caches per CAS endpoint url.** It keys `~/.cache/huggingface/xet/<origin>-<hash>/shard-cache/` on the url the fake hands out, and skips the xorb upload for content it finds there. The old fake ran on an ephemeral port per run, so every run got a fresh directory and this never fired. On a fixed port the fake answered `shard names an unknown xorb <hash>` and the write hung.

The untenanted CAS above is half the fix. The other half is that `casUrl` now carries a nonce minted once per process (`/cas/<BOOT>/...`): this fake's CAS is a SQLite file it recreates at startup, so a fixed url let a client cache outlive the content it described, and after any restart the next write committed bytes the fake had never received.

**The kit's `rangeReply` aborts opendal's hf read.** A start at or past EOF is a 416, which is what RFC 7233 says and what a real CDN does, and opendal's hf service reads that as a permanent read failure and gives up. The old fake clamped the window and answered a 206 carrying zero bytes, which is what `range_offset_past_eof_is_empty` is written against. `clampedReply` in `server/hf/wire.ts` keeps that, and is the only place a fake here does not use the kit helper.

This preserves the fake's behaviour; it does not fix the backend against a real Hub, which would return the 416 and hit the same abort. That is #810, and the docstring says so, so the fake is not mistaken for evidence the backend handles it.

Worth recording how the second one hid: the TypeScript battery exited 1 with zero `fail` lines, because the runner died on the aborted read rather than reporting a case. A `| tail` in the first invocation took tail's exit code and it read as green.

## Also

- The `integ-typecheck` hook's `files` filter named five fakes and had fallen behind `tsconfig.check.json` by five more, so editing box, dropbox, onedrive, slack, notion or mem0 skipped the only hook that type checks them. It now matches every server directory.
- The watch job starts this server and passes `HF_URL`. `build_hf` ran in process before, so gating it on the variable without that would have skipped hf watch coverage in CI in silence.
- `python/uv.lock` is hand-edited. Running `uv lock` locally bumps `revision` from 1 to 3 and adds an `upload-time` field to every entry: 10800 changed lines for one removed dev dependency. `uv lock --check` passes on the edit.

## Verified

hf battery 4303 passed, 0 failed on each host against one shared server, run in both orders. Watch battery 22 of 22. `pre-commit run --all-files` and the python suite green.


zechengz:
@codex review

PR #929, opened 2026-08-27
feat(commands): expose builtin command catalogs

### LINKED ISSUE #435: Design: public way to get a backend's builtin command (e.g. s3 cat)
## Problem

Users who want to extend or wrap a builtin command (the PR #433 s3 cost-model example overrides cat's provision) have no import path for it:

- `mirage.commands.builtin.generic.cat` is the pure stream transform, not a command.
- `mirage.commands.builtin.generic_bind.builders.cat.cat` is the command body but needs the backend's `CommandIO`, and each backend's bound ops (`_S3_CMD_OPS`) is module-private.
- The finished s3 cat only exists as an anonymous entry in `mirage.commands.builtin.s3.COMMANDS` (the factory removed the old per-backend `cat.py` wrappers, which is what made it importable before).

The only public handle today is a live mount: `ws.mount('/s3/').resolve_command('cat', None)`. That works (the s3 example uses it) but:

- it requires a constructed workspace, so nothing is available at import/decoration time;
- the returned fn is shared across all mounts of that backend, and `command(...)(builtin_fn)` silently contaminates it: `functools.wraps` copies `__dict__`, so the new registration appends into the builtin's shared `_registered_commands` list and every future s3 mount picks it up;
- integ/s3.py had to rebuild the estimator from `make_file_read_provision(s3_stat)` and delegate through a module-level holder dict, which is boilerplate nobody should have to discover.

## Options

**A. Lazy named exports per backend (PEP 562).** A module `__getattr__` on each backend package that resolves names from `COMMANDS`, so `from mirage.commands.builtin.s3 import cat` works again without reintroducing wrapper files. Each import returns the decorated fn; needs a story for the shared `_registered_commands` hazard (e.g. return a fresh copy).

**B. Public per-backend CommandIO.** Export the bound ops (`S3_COMMAND_IO`), so users bind any generic builder themselves: `functools.partial(builders.cat.cat, with_read_cache(S3_COMMAND_IO))`. Most flexible, most low-level.

**C. Registry-level helper.** Keep resolution as the story but make it safe and ergonomic: a `RegisteredCommand.with_(provision=..., fn=...)` copy helper plus `mount.register(...)", so an override is one line and never mutates shared state. Cheapest; does not help at import time.

TS mirror: TS has no decorators, users already rebuild `new RegisteredCommand({...})`; whichever option lands should keep the two stories aligned (a public named export of backend command arrays exists there, so A has a natural TS twin).

Came out of the PR #433 example work.

sonhmai:
## What changed

- Expose the S3 builtin command catalog from Python, Node.js, and browser packages.
- Add exact command lookup and copy-with-overrides helpers for handlers and provisioning.
- Keep custom behavior local to the resource where it is registered.
- Preserve the existing TypeScript array API and Python bulk registration path.
- Update the S3 docs and Python example with the new setup flow.

## Why

Applications previously had to inspect a live workspace or redeclare a builtin command before changing its behavior. Hosts can now select and customize a builtin command before creating a workspace, without changing the shared defaults or other mounts.

## Verification

- 10,193 TypeScript core tests passed.
- All affected Python tests passed.
- All repository formatting, lint, type, and dead-code checks passed.
- Integration mock-server type checking passed.

Closes #435


sonhmai:
@codex review

sonhmai:
@codex review

docs/python/resource/s3.mdx
```diff
@@ -119,6 +119,30 @@ The S3 resource supports the full set of shell commands since it operates
...
+
+cat = S3_COMMANDS.require("cat")
+
+
+async def logged_cat(accessor, paths, texts, opts):
+    print("reading", [path.virtual for path in paths])
+    return await cat.fn(accessor, paths, texts, opts)
+
+
+ws.mount("/s3/").register(cat.with_overrides(fn=logged_cat))
```
zechengz, line 139: I haven't taken a deeper look but there is a filetype specific registration.
And I will think about the interface for this.

sonhmai: Sure. `register_fns()` and still is there. This api is more like a nice to have. Let's look at it more.

zechengz: sg

PR #930, opened 2026-08-27
docs: say how a view reaches a handler, and delete the opt-in helper

bytecii:
View delivery to command handlers moved from parameter-name opt-in — where
`execute_cmd` inspected a handler's signature and injected `links`/`mounts`/`ns`
only into the handlers that named one — to unconditional field reads off
`CommandOpts`. Ten sites still described the mechanism that was removed, or
contradicted the code outright. This is documentation and dead code only; no
behavior changes.

### The opt-in story

`ops/types.py` (`SessionView`, `MountView`, `LinkView`, `NamespaceView`) and the
three builders in `workspace/executor/command/run.py` said a command "opts in by
naming a `links`/`mounts`/`ns`/`session_view` parameter". Every one now says what
happens: the view rides `CommandOpts` (and `CLIDoors` on the CLI tier) into every
invocation, and a command opts in by reading the field. The TypeScript twin in
`ops/types.ts` already stated this, so it is untouched.

The `Mount boundaries` bullet in `CLAUDE.md` carried the same dead claim, plus
`execute_cmd`/`executeCmd` "delivers it only to handlers that do", which
contradicted the file's own later rule ("A command consumes a fact by reading the
`opts` field, nothing else… no opt-in registry, spec field, or signature
convention"). It now says `MountView` rides `opts.ns.mounts`, and its field list
gains `visible_descendants` — the avoid-a-boundary / name-a-boundary split is
load-bearing and was missing. That later rule spelled the read `opts.links`,
but `CommandOpts` has no top-level `links` field (only `ns`), so it is now
`opts.ns.links`. `AGENTS.md` is a symlink to `CLAUDE.md` and follows.

### Claims that were simply wrong

- **`seed_var`** (`workspace/session/state.py`) said `execute_command` lands a
  prefix assignment (`FOO=bar cmd`) through the ungated door "so a `pre_session`
  rule never sees one", and that closing it was deferred work. It is closed:
  `workspace/node/command_dispatch.py` asks `ensure_var_visible` and then
  `pre_session`, with the value, before it seeds anything. The docstring now
  records the gate and keeps the GNU divergence (a refusal takes the whole
  statement, where bash prints its readonly refusal, runs the command anyway and
  exits 0) as settled rather than pending.
- **`workspace/cli/registry.py`** cited "the policy layer's `pre_execute`". No
  such hook exists — `VALIDITY` in `policy/types.py` names `pre_command`. The
  same wrong name appeared in `CLAUDE.md` and, as `preExecute`, in the
  TypeScript twin `workspace/cli/registry.ts`; all three are fixed.
- **`generic_bind/adapter.py`** said `glob_children` is stamped from
  `opts.child_mounts`. The real read is `opts.ns.child_mounts`
  (`factory.py`, `_run_with_namespace_globs`).
- **`CLIDoors.dispatch`** (`commands/cli/types.py`) said the `DispatchFn`
  Protocol "lives in `workspace.types`" and that the field is "typed loosely on
  purpose". It is declared in `runtime/types.py`, on the consumer side, and the
  field is annotated `DispatchFn | None`. (`workspace/types.py` holds only
  `ExecutionNode`.)

### Dead code

- `mirage/utils/params.py` is deleted. `accepts_kwarg` was the signature probe
  the old dispatcher used; it had no production caller left, and its docstring
  described dispatcher behavior that no longer exists. Its one importer,
  `tests/commands/test_read_range_optin.py`, now carries a local `_takes_window`
  predicate that asks the question that test actually has — does this reader
  declare `offset`/`size`? — with the same `**kwargs`-is-not-consent rule.
- Four bare no-op expression statements `doors.stat_path` in the `git`
  verbs `log`, `show`, `branch` and `diff` are removed. They were copy-paste
  remnants of the `stat_path = doors.stat_path` that `add`/`status` still need;
  `opened()` reads that field itself and owns the refusal, so the statements
  bound nothing and guarded nothing. Verified: with no doors, and with
  `dispatch` wired but the stat/name planes absent, all four verbs still exit
  128 with `fatal: this operation must be run in a work tree`.
- `CLIDoors` was the only one of the four record types in `commands/cli/types.py`
  missing from that package's `__all__`, while `CLIInvocation`, `CLISpec` and
  `WalkResult` are all re-exported; it is added.

### Two ratchets moved, both on an improvement

Deleting `mirage/utils/params.py` was at once a Python-only layout divergence and
an unmirrored source module, so it tripped two independent gates that fail when
the count falls:

- `spec/layout_exceptions.json` baseline 253 → 252
- `MIRROR_BASELINE` in `tests/test_tests_mirror_sources.py` 182 → 181

### Verification

- `pre-commit run` over the changeset: clean.
- `pytest`: green (the three `integ-*` pre-commit hooks fail under `--all-files`
  in a fresh worktree for environment reasons — `pnpm exec` run from `integ/`
  cannot see its workspace root at the sibling `typescript/`, and
  `--ignore-scripts` skips `prisma generate` — and they report
  `(no files to check)` on this changeset, which touches no `integ/` file).
- `pnpm -r typecheck`: all `typescript/` projects clean.
- `scripts/check_layout_parity.py --strict`: green at the new baseline.


zechengz:
@codex review

PR #932, opened 2026-08-27
fix(runtime): withhold the history view and root anchor from runtimes

bytecii:
## What

Python announced the **unfiltered** mount table to sandboxed runtimes; TypeScript filters it (`sandboxVisibleMounts`). This PR ports the filter and pins it in both languages, and writes the runtime tier's session-blindness doctrine down on `LanguageRuntime`.

## The divergence

The workspace hands every `LanguageRuntime` a resolver at `attach` time — the mount prefixes a guest may see (WASI preopens one directory per mount; `RuntimeVFS.prefixes()`/`mount_of` route on them; `PolicyContext.mounts` is stamped from the same resolver).

- **TypeScript** (`workspace.ts` `sandboxVisibleMounts`) withholds two prefixes, neither for being `/`: the **history view** (a shell surface, not a place to put files) and the **synthetic root anchor** (a mount nobody asked for — the workspace adds it so arg-less commands resolve). An explicit root mount is forwarded like any other prefix.
- **Python** passed `self._ops.mount_prefixes` unfiltered, so a WASI guest preopened `/.bash_history` and every runtime reported a claim on the anchor. `PrefixResolver`'s own docstring already promised "a sandbox-filtered list for the runtimes" — the docstring described TS, not Python.

## Changes

- `workspace.py`: new `Workspace._sandbox_visible_mounts()` (live read off the registry, mirroring TS), wired into the one `PrefixResolver` construction. `PolicyContext.mounts` follows the same resolver in both languages, so policies now see the same list too.
- `tests/workspace/test_runtime_binding.py`: a `ResolverProbe` runtime captures the attached resolver; three pins — history withheld, synthetic anchor withheld, **explicit root forwarded** (guards against over-filtering). The two "withholds" pins were verified red against the old wiring.
- `workspace.test.ts`: the same three pins. TS behavior was already correct but had no direct test.
- `language.py` / `language.ts`: doctrine paragraph — the doors are the data plane (dispatch) and the name plane (resolver), and the list is complete on purpose: there is **no session door**. A guest env write lands on the guest's own copy and dies with the run, so guest code can neither trip nor bypass a `pre_session` rule. A runtime that ever needs session state must take a gated `SessionView`.

## What this PR deliberately does not do

This came out of a design review of how state planes reach the runtime tier (D4: "runtimes have no doors record"). The survey falsified the record: `attach(dispatch, resolver)` already **is** the two doors — data plane and name plane — mirrored field-for-field across languages, with `MountResolver` documented as "the name-plane questions a runtime asks" and growth landing as protocol methods fed by injected sources, not signature changes. A `RuntimeDoors` record would additionally have Python-only payload (the session/recorder re-bind exists only where guest calls cross threads; TS has no ambient capture at all), violating the parity it was meant to serve. What the survey *did* surface is this filter divergence, which is the PR.

### DIFF
--- a/python/mirage/runtime/language.py
+++ b/python/mirage/runtime/language.py
@@ -40,6 +40,15 @@ class LanguageRuntime(Runtime):
     sandboxed interpreter bridges file I/O through the workspace
     dispatch attached here, while a host subprocess only sees the host
     filesystem and keeps the default no-op attach.
+
+    The doors are the data plane (dispatch) and the name plane
+    (resolver), and the list is complete on purpose: there is no
+    session door. A guest reads the frozen ``RunArgs.env`` snapshot; an
+    env write lands on the guest's own copy and dies with the run,
+    never reaching the live session, so guest code can neither trip nor
+    bypass a ``pre_session`` rule. That blindness is a security
+    boundary, not a gap — a runtime that ever needs session state must
+    take a gated ``SessionView``, never a bare session reference.
     """
 
     language: ClassVar[Language]
--- a/python/mirage/workspace/workspace/workspace.py
+++ b/python/mirage/workspace/workspace/workspace.py
@@ -233,7 +233,7 @@ class Workspace:
 
         self._runtimes, self._policy_router = wire_runtime_world(
             self._registry, self.dispatch,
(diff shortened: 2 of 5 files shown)


zechengz:
@codex review

PR #952, opened 2026-08-29
fix(integ): give the typescript runner argparse's option grammar

sonhmai:
Split out of #914, which had grown this while fixing one bug in its own new
option.

## Why

The two integ runners are held to the same command line, so a spelling one
accepts and the other drops is a difference in what CI accepts. The python
runner uses argparse. The typescript one matched option names by exact string
and ignored every word it did not recognise, so:

- `--target-jobs=4` ran the whole battery **serially, in silence**, where the
  python runner takes the same line
- a trailing `--target-jobs` did the same
- a typo'd flag ran with the defaults rather than failing
- there was no `-h`/`--help` at all, so the option list was written down
  nowhere

## What it now does

Rules copied off argparse's `_parse_optional`, not guessed:

- `--opt=value` as well as `--opt value`
- an unknown word is refused, space-joined the way argparse joins them
- a one-character token, a negative number and a token holding a space are
  values, not options. `NEGATIVE_NUMBER` is argparse's own
  `_negative_number_matcher`, so `-.5` and `-0.5` are values while `-5.` and
  `-1e5` are not
- `-h`/`--help`, rendering the same text argparse renders, exit 0
- a zero-argument option refuses an explicit value (`--strict=foo`)
- `--` ends the options
- `--target-jobs` parses as `int()` does: a sign, leading zeros, underscores
  between digits, and nothing else. `4.0`, `1e1` and `0x4` are refused
- **error order is observable.** An error raised while consuming an option
  beats a later `--help`, but `unrecognized arguments` is reported after the
  walk, so any help flag beats it. The `>= 1` floor is not argparse's and runs
  last, so `--target-jobs 0 -h` prints usage

## Verification

Differential fuzzing against a matching argparse: generate argv vectors, run
both, compare exit code and parsed values. **1823 vectors per seed, three
seeds, zero mismatches.**

That harness is what found four of these. Two of them no review had raised,
and one was a fix of mine that was wrong in the opposite direction: refusing
unknown words inline broke `main.ts x --help`, and refusing them in a pre-pass
broke `--strict=foo --help`. Neither is argparse's behaviour.

## Deliberate divergence

argparse defaults to `allow_abbrev=True`, so `--fac core` and `--prof` reach
the python runner as `--facet` and `--profile`. This does not, and refuses all
three as unrecognized.

Prefix matching plus ambiguity reporting is more machinery than a CI runner's
option list is worth, and the divergence errs strict: a line that works here
works there, never the reverse, so no invocation can silently mean two things.
Every line in `.github/workflows` spells its options in full. Documented in
place rather than left to be rediscovered.

### DIFF
--- a/integ/runners/typescript/main.ts
+++ b/integ/runners/typescript/main.ts
@@ -62,26 +62,158 @@ function parseArgs(): {
   let allowSkip = ''
   let targetJobs = 1
   let profile = false
-  const argv = process.argv.slice(2)
-  for (let i = 0; i < argv.length; i++) {
-    if (argv[i] === '--target' && i + 1 < argv.length) targets.push(argv[++i])
-    else if (argv[i] === '--facet' && i + 1 < argv.length) facet = argv[++i]
-    else if (argv[i] === '--emit' && i + 1 < argv.length) emit = argv[++i]
-    else if (argv[i] === '--strict') strict = true
-    else if (argv[i] === '--allow-skip' && i + 1 < argv.length) allowSkip = argv[++i]
-    else if (argv[i] === '--profile') profile = true
-    // Recognised before its operand is checked: gated on `i + 1 < argv.length`
-    // a trailing `--target-jobs` fell through every arm and ran serially in
-    // silence, where the argparse runner refuses the same line.
-    else if (argv[i] === '--target-jobs') {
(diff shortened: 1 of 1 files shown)


sonhmai:
nah

PR #963, opened 2026-08-30
fix(dropbox): stop stat and du laundering a listing 5xx into a false ENOENT

sonhmai:
## Why

`stat()`'s index-miss path populates the parent listing through `readdir` and
caught `(FileNotFoundError, DropboxApiError)`. `readdir` already maps a genuinely
missing path to ENOENT, so the only `DropboxApiError` that reaches `stat` is a
**non-409** — a 5xx or 429 while listing the parent. It was swallowed and
re-answered as ENOENT. A false ENOENT is the one an agent acts on destructively,
and it is exactly what `MISS_ERRORS` warns against: a transport error is not
absence.

Two twins carried the same bug:

- `stat.ts` had a bare `catch {}` that swallowed *everything*, ENOTDIR included,
  so a path under a file read back as "missing" on the TS side.
- `du/walk.ts` had the same bare `catch {}` one layer up, swallowing the very
  5xx `stat` should surface — so TS `du` silently under-counted a subtree it
  could not read, where Python `du` (which catches only `FileNotFoundError`)
  fails loud per the `du` contract.

And the tests reached `api.dropboxapi.com` on every run: the missing-path test
mocked only `readdir.list_folder`; when that 409'd, readdir's ancestor walk
called the unmocked `get_metadata` and hit the network. In most environments
Dropbox rejected the fake creds with a tolerated `DropboxApiError` (a false
green); under a TLS-intercepting egress proxy it surfaced as an uncaught SSL
error and a deterministic failure.

## What changed

- **Source (py + ts).** Narrow the populate catch to ENOENT only
  (`except FileNotFoundError` / `if (!isEnoent(err)) throw err`). In Python this
  changes only the 5xx path — `NotADirectoryError` was never in the tuple, so
  ENOTDIR already propagated; the TS twin swallowed both, so both now surface.
  Narrow the two `catch {}` in `du/walk.ts` the same way, mirroring
  `du/walk.py`, so TS `du` stops under-counting an unreadable subtree.
- **Tests, one seam.** Rework `test_stat` and add `stat.test.ts` (dropbox had
  none) around a single stable seam — the JSON-RPC transport (`dropbox_rpc` /
  `dropboxRpc`) via the shared `FakeDropboxRpc`, with the **real** `readdir`.
  `list_folder`, `get_metadata` and the token refresh all funnel through that
  one function, so the suite cannot touch the network and an unexpected endpoint
  fails loudly. readdir's error mapping is deterministic, so ENOENT / 5xx /
  ENOTDIR are driven end to end.
- **Coverage.** Add the untested `NULL_INDEX` `_stat_from_api` path (file,
  folder, 409→ENOENT, non-409 propagates), the index-hit caching contract,
  populate-then-still-absent ENOENT, the 5xx and ENOTDIR propagation that pin
  the fix, mount-prefix arithmetic, and the `_stat_from_entry` field fallbacks.
  Give `du.test.ts` realistic coded errors and pin `du`'s 5xx propagation in
  both languages. Drop the dead token seed and the local `make_accessor` for the
  conftest `dropbox_accessor` fixture.

## Verification

- python: `uv run pytest tests/core/dropbox` — all green (`test_stat` 4 → 19),
  no live network; pre-commit (format, ruff, isort, PathSpec) clean.
- typescript: `vitest run src/core/dropbox` — 72 passed; new `stat.test.ts` and
  the du 5xx pin included; `tsc --noEmit` and prettier clean on the changed
  files.
- Every new propagation test fails against the pre-fix source, so it pins the
  behaviour rather than restating it.

## Deliberate divergence

py and ts mirror in coverage and seam. One pre-existing structural divergence
the tests reflect rather than introduce: TS `FileType` is one flattened enum, so
`stat` renders a file's `type` as `guessType(name)` (e.g. `text`), while Python
carries `type=FILE` + a separate `content`; each language's assertions match its
own shape.

### DIFF
--- a/python/mirage/core/dropbox/stat.py
+++ b/python/mirage/core/dropbox/stat.py
@@ -95,10 +95,15 @@ async def stat(
                          resource_path=mount_key(parent_virtual, prefix)),
                 index=index,
             )
-        except (FileNotFoundError, DropboxApiError) as exc:
-            # Parent listing failed (missing dir surfaces as a 409 from
-            # the API): fall through to enoent, mirroring the TS stat.
-            logger.debug("stat populate failed for %s: %s", virtual, exc)
+        except FileNotFoundError as exc:
+            # readdir already maps a genuinely missing path to ENOENT, so
+            # catch only that. A non-409 DropboxApiError (a 5xx/429 while
+            # listing the parent) was previously swallowed here and
+            # re-answered as a destructively actionable false ENOENT; it now
+            # surfaces. NotADirectoryError (a path under a file) was never in
+            # this catch and already propagated.
+            logger.debug("stat found no parent listing for %s: %s", virtual,
+                         exc)
         result = await index.get(virtual_key)
         if result.entry is None:
             raise enoent(virtual)
--- a/python/tests/core/dropbox/du/test_size.py
+++ b/python/tests/core/dropbox/du/test_size.py
@@ -18,7 +18,7 @@ import pytest
 
 from mirage.accessor.dropbox import DropboxAccessor
 from mirage.cache.index.ram import RAMIndexCacheStore
-from mirage.core.dropbox.client import DropboxTokenManager
+from mirage.core.dropbox.client import DropboxApiError, DropboxTokenManager
 from mirage.core.dropbox.du import size
 from mirage.resource.dropbox.config import DropboxConfig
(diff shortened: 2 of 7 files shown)


sonhmai:
@codex review

zechengz (approved):

PR #968, opened 2026-08-31
refactor(observe): own op timing in the record path

sonhmai:
## Why

Time was read ambiently across the codebase. Around 45 Python modules under `mirage/core/<backend>/` repeated the same stopwatch line before their backend call and handed the result to `observe.context.record`, and `Ops` kept a second timing path of its own. TypeScript mirrored both. One number had roughly eighty definitions, so op latency meant slightly different things in different places.

## What

The stopwatch now belongs to observe/context, the one consumer of the number.

- `observe/context` gains `OpTimer` and `start_op` / `startOp`. A call site opens the timer where the backend work begins.
- `finish_record` / `finishRecord` is the single place an op's duration and wall-clock stamp are read. `Ops` builds its record through it too, so the facade's ledger and the recorder sink cannot disagree about what a duration measures.
- `record` takes the timer instead of a start stamp. Call sites stop reading a clock and mostly drop their `time` import.

Converted 53 sites in Python and 48 in TypeScript.

## Intent

- This is the enabling step, not the destination: every op's duration now sits behind one owner (`start_op` / `finish_record`).
- With that single seam, later work can change how time is read in one place instead of across ~80 call sites:
  - an injectable clock for deterministic tests
  - seeded or frozen time
  - a lint rule that keeps new call sites off the ambient clock
- Concretely, a follow-up swaps the clock source at `start_op` alone, and every call site inherits it:

```python
# now: the timer reads the clock itself
def start_op() -> OpTimer:
    return OpTimer()  # OpTimer() reads time.monotonic()

# follow-up (not this PR): source the clock from one seam
def start_op() -> OpTimer:
    return OpTimer(clock=current_clock())  # injected, frozen or seeded under test
```

- None of that ships here. This PR only moves who reads the clock.

## No output change

OpRecord fields, duration semantics, wall stamps, bytes and fingerprints are untouched. Each timed interval starts and ends exactly where it did (s3 excludes client acquisition, disk includes path resolution), so durations keep their per-site meaning. Streaming records are unaffected.

One deliberate change: TypeScript `Ops` measured with `Date.now()` while `record` used `performance.now()`. Folding the two puts `Ops` on the monotonic clock, matching Python's `Ops` and `record` in both languages. Same field, same units, better source.

## Verification

CI. The diff is wide but shallow (drop `import time`, open a timer, pass it) with no behavior to cover. Existing tests that call `record` directly were updated for the new signature; no assertions changed.

Part of #970 (item 1).

### DIFF
--- a/python/mirage/core/box/write.py
+++ b/python/mirage/core/box/write.py
@@ -12,13 +12,11 @@
 # limitations under the License.
 # ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
 
-import time
-
 from mirage.accessor.box import BoxAccessor
 from mirage.cache.context import invalidate_after_write
 from mirage.core.box.api import upload_file_version, upload_new_file
 from mirage.core.box.resolve import path_parts, resolve_item, resolve_parent_id
-from mirage.observe.context import record
+from mirage.observe.context import record, start_op
 from mirage.types import PathSpec
 from mirage.utils.errors import enoent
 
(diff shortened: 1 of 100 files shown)


sonhmai:
@codex review

sonhmai:
@codex review

zechengz (approved):
Thanks :)

PR #988, opened 2026-09-02
fix(sharepoint): scope the site/drive id cache to the accessor

sonhmai:
## Problem

Two SharePoint mounts in one process. Two tenants. One site name.
The second mount gets the first tenant's site id.

```
                one Python process running mirage
              ----------------------------------------------
                    resolve.py (imported once)
                 +--------------------------------+
                 | _site_cache   (module global)  |
                 | "Engineering" -> contoso...AAA |
                 +-------^---------------^--------+
                         |               |
              (1) writes |               | (2) reads -> gets A's id  X
                         |               |
              +----------+---------+  +--+-----------------+
              | /spA   accessorA   |  | /spB   accessorB   |
              | Tenant A - token A |  | Tenant B - token B |
              +--------------------+  +--------------------+
```

The name -> id cache is a module global:

```python
# core/sharepoint/resolve.py
_site_cache: dict[str, str] = {}   # keyed by site display name
```

`"Engineering"` is unique inside a tenant. The dict is shared by every
accessor in the process. Whoever resolves first writes the slot; everyone
after reads it.

## Current behavior

The two lines that leak, `core/sharepoint/resolve.py`:

```python
async def _resolve_site_id(accessor, site_name):
    if site_name in _site_cache:        # (a) hits on ANY tenant's entry
        return _site_cache[site_name]   # (b) returns it, no tenant check
    ...
```

Two mounts, `accessorA` (Contoso, token A) and `accessorB` (Fabrikam,
token B). Both tenants have a site named `Engineering`.

```
(1) cat /spA/Engineering/Documents/report.docx
    _resolve_site_id(accessorA, "Engineering")
    -> miss -> GET /sites with token A -> Contoso's sites
    -> _site_cache["Engineering"] = "contoso...AAA"        OK for A

(2) cat /spB/Engineering/Documents/report.docx
    _resolve_site_id(accessorB, "Engineering")
    -> (a) hit -> (b) return "contoso...AAA"               X
    Fabrikam is never called. accessorB holds Contoso's site id.

(3) _resolve_drive_id(accessorB, "contoso...AAA", "Documents")
    -> GET /sites/contoso...AAA/drives  with token B
    Best case 403/404: mount B looks broken for no reason.
    Worst case: B's session is pointed at A's resources.
```

Order decides who is wrong. Mount B first and A breaks instead.

## To-be

Give the cache the scope its data has. One accessor is built per
`SharePointResource` -- one per mount, one config, one tenant -- and it
lives as long as the mount. Put the dicts on it.

```diff
 # accessor/sharepoint.py
 class SharePointAccessor(SessionAccessor):
     def __init__(self, config: SharePointConfig) -> None:
         super().__init__(timeout=...)
         self.config = config
+        self.site_cache: dict[str, str] = {}
+        self.drive_cache: dict[tuple[str, str], str] = {}

 # core/sharepoint/resolve.py
-_site_cache: dict[str, str] = {}
-_drive_cache: dict[tuple[str, str], str] = {}

 async def _resolve_site_id(accessor, site_name):
-    if site_name in _site_cache:
-        return _site_cache[site_name]
+    cache = accessor.site_cache
+    if site_name in cache:
+        return cache[site_name]
     ...
```

Same key. Same lookup. Same populate. Only where the dict lives changed.

```
                one Python process running mirage
              ----------------------------------------------
                    resolve.py (no module global)

              +--------------------+  +--------------------+
              | /spA   accessorA   |  | /spB   accessorB   |
              | Tenant A - token A |  | Tenant B - token B |
              |                    |  |                    |
              | site_cache (own)   |  | site_cache (own)   |
              | "Engineering"      |  | "Engineering"      |
              |  -> contoso...AAA  |  |  -> fabrikam...BBB |
              +--------------------+  +--------------------+
                 two dicts, no shared slot, nothing to collide
```

Step (2) now misses, lists Fabrikam with token B, and resolves
`"Engineering"` to `fabrikam...BBB`. Hit rate on any one mount is
unchanged: the accessor outlives every op on it. The only lookups that now
miss are the ones that used to return another tenant's id.

While here: `site_entries` now also caches a site's internal `name`, as
the TypeScript `siteEntries` does. Before, a root `readdir` filled the
cache by display name only, so a later lookup by `name` relisted `/sites`.

## Tests

1 new test, 11 existing test files changed.

Before running, the existing tests put a known site id and drive id into
the two module-level dicts, so the code under test would not call `/sites`
and `/drives`. Those dicts are gone. The tests now put the same ids on the
accessor instead.

`test_resolve.py` does this inside each test rather than once for the
whole file, because some of its tests need the cache empty to check the
not-found path. The new test there checks that `site_entries` caches both
the display name and the internal name.

## Notes

No TypeScript change. `SharePointAccessor` there already keeps
`siteCache` / `driveCache` per instance and caches both names; this brings
python in line.

### DIFF
--- a/integ/runners/python/adapters/__init__.py
+++ b/integ/runners/python/adapters/__init__.py
@@ -48,7 +48,6 @@ from mirage.commands.cli.types import CLISpec
 from mirage.core.databricks_volume.path import configured_root
 from mirage.core.discord.config import DiscordConfig
 from mirage.core.email.config import EmailConfig
-from mirage.core.sharepoint import resolve as sharepoint_resolver
 from mirage.resource.aliyun import AliyunConfig, AliyunResource
 from mirage.resource.backblaze import BackblazeConfig, BackblazeResource
 from mirage.resource.box import BoxConfig, BoxResource
@@ -1500,13 +1499,6 @@ class LinearService:
         return None
 
 
-def _clear_sharepoint_caches() -> None:
-    # The resolver's site/drive id caches are module globals; a fresh
-    # fake tenant per run must not see ids from the previous one.
-    sharepoint_resolver._site_cache.clear()
-    sharepoint_resolver._drive_cache.clear()
-
-
 class JaegerService:
     """Points jaeger mounts at a real jaeger all-in-one container.
 
@@ -1583,7 +1575,6 @@ class SharePointService:
     async def create(cls, run_id: str, target: dict) -> "SharePointService":
         service = cls(f"{run_id}-{target['id']}",
                       os.environ["ONEDRIVE_URL"].rstrip("/"))
-        _clear_sharepoint_caches()
         for mount in target["mounts"]:
             await service.provision(mount)
         return service
@@ -1637,7 +1628,7 @@ class SharePointService:
                              key_prefix=mount.get("prefix")))
 
     async def teardown(self) -> None:
-        _clear_sharepoint_caches()
+        return None
 
 
 class NotionService:
--- a/python/mirage/accessor/sharepoint.py
+++ b/python/mirage/accessor/sharepoint.py
@@ -26,5 +26,24 @@ class SharePointAccessor(SessionAccessor):
     def __init__(self, config: SharePointConfig) -> None:
         super().__init__(timeout=aiohttp.ClientTimeout(total=config.timeout))
         self.config = config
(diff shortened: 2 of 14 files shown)


sonhmai:
@codex review

zechengz (approved):
Thanks!

Issue #970, opened 2026-08-31
Tracker: deterministic execution and workspace simulation


sonhmai:
## Goal

Make workspace execution reproducible under test: one seed replays one interleaving of commands, cache activity, and injected faults. This makes the open concurrency class (#411, #432) provable to fix and hard to reintroduce. It matters more as mirage becomes a workspace substrate for agent control planes: many concurrent sessions over shared state turn these invariants into someone else's production.

## How to read this plan

This is a set of suggested improvements based on how the codebase looks today (2026-08). It is a living list, not a commitment. Expect it to change: research may add items, evidence may reorder them, and any item whose effort stops matching its value should be dropped rather than done for completeness. When an item is dropped or reshaped, leave a short comment here saying why, so the reasoning stays visible to the next reader.

## Scope note

This tracker covers the determinism/concurrency slice of the testing strategy only. Sibling tracks get their own trackers: GNU/parity differential generation, adversarial policy fuzzing (isolation soundness), per-backend fake-drift conformance, and call-count perf budgets.

## Design anchors

- The dispatcher (`MountEntry.execute_op`) is the single waist where fake accessors swap in; `python/tests/e2e/s3_mock.py` is the fake template (single seam, in-memory dict).
- The integ fake-server kit's `integ/server/kit/typescript/clock.ts` (frozen base, tick per touch) is the Clock design to copy into product code.
- Prior art for this system shape: Dropbox's sync-engine testing (randomized deterministic simulation with mocked server, filesystem, and timers, seed replay, case minimization) and Coyote/shuttle-style controlled scheduling for single-process async cores. Not the FoundationDB/TigerBeetle cluster-simulator shape.

## Suggested sequence, one PR per item

- [x] 1. Consolidate op timing into `observe/context` (`OpTimer`/`start_op`/`finish_record`, fold `Ops._record`) — #968, merged 2026-09-01
- [ ] 2. `Clock` protocol with stdlib default, owned by Workspace, constructor-injected; migrate cache TTL and FUSE prefetch TTL — #990, in review
- [ ] 2b. Follow-up once #990 lands: move the index cache TTL (`cache/index/`, expires on `datetime.now()`) and `record_stream` timing onto the Clock, and decide how the ~100 backend `start_op()` timers reach the workspace clock (candidate: carry it on the existing recording scope). All three were deliberately left out of #990 to keep it small; the simulator (item 8) is the first consumer that needs them
[...]

PR #989, opened 2026-09-02
refactor(gdrive): route every Drive call through one client seam

### LINKED ISSUE #684: gdrive find is untestable: the fake Drive patches readdir's list_files but not resolve's or tree's
Found while verifying #270. `find` over a gdrive mount cannot be exercised against `tests/e2e/gdrive_mock.py` at all, so nothing covers it.

## What happens

```
$ find /gdrive/tree -type f
find: 401, message='Unauthorized', url="https://www.googleapis.com/drive/v3/files?q='root'+in+parents+and+name%3D'tree'..."
```

The call reaches the **real** Drive API mid-test.

## Cause

`_PATCH_TARGETS` patches `list_files` at one binding only:

```python
"list_files": [
    "mirage.core.gdrive.readdir.list_files",
],
```

but `find` goes through `mirage.core.gdrive.resolve` (`resolve_key`) and `mirage.core.gdrive.tree` (`iter_tree`), each of which imports `list_files` by value from `mirage.core.google.drive`. Those bindings stay live, so they hit the network.

## It needs more than extra targets

Adding `mirage.core.gdrive.resolve.list_files` and `mirage.core.gdrive.tree.list_files` gets further, then fails on the fake's signature:

```
find: _build_fakes.<locals>.fake_list_files() got an unexpected keyword argument 'name'
```

`resolve` passes a `name=` filter that `FakeGDrive` does not implement. So the fix is two parts: widen `_PATCH_TARGETS`, and teach the fake the `name` (and any other) query filter its callers use.

`list_shared_drives` has the same split (`resolve` imports it too).

## Why it matters

Anything reached only through `resolve`/`tree` is uncovered on gdrive: `find`, and the tree walk generally. A test that exercises them today either 401s or silently depends on live credentials.

I backed out a partial fix rather than destabilise a shared fixture inside an unrelated PR (#683).

sonhmai:
## Why

Tests could not fake Google Drive reliably. Each `core/gdrive` module imported Drive calls (`list_files`, `download_file`, ...) directly, so the e2e fake had to patch 11 places and still missed some. A missed one sent the test to the real Drive API, which failed with a 401 (#684). S3 does not have this problem: it has one connection seam, so one patch swaps the whole backend.

## What

- New `core/gdrive/api.py` and `api.ts`. `DriveApi` lists every Drive call the backend makes. `DriveClient` is the real one. `drive_api(token_manager)` builds it.
- `GDriveAccessor.drive` is the single door. Every gdrive op calls `accessor.drive.<call>()`. Helpers that took a `TokenManager` only to reach Drive now take a `DriveApi`.
- 25 direct imports removed across 13 Python modules, and the same set in TypeScript. Only `api.py` and `versions.py` (the Revisions wire module) still name a raw Drive call.
- The e2e fake patches one target. `FakeDriveApi` also covers writes, which the old mock never did.
- A guard test fails if a gdrive module imports a Drive call directly again.

## Intent

Before, the fake had to know every import site, and a new one was invisible:

```python
_PATCH_TARGETS = {
    "list_files": ["...readdir.list_files", "...resolve.list_files",
                   "...tree.list_files", "...rename.list_files"],
    "download_file": ["...read.download_file"],
    ...  # 11 targets
}
```

After, it names the door:

```python
patch("mirage.accessor.gdrive.drive_api", new=lambda tm: FakeDriveApi(...))
```

## Reviewer notes

- `GDriveAccessor.drive` builds a client per access instead of caching one in `__init__`. The resource builds its accessor before a test installs the fake, so a cached client would point at the real API. The client is a frozen dataclass; building one costs nothing next to the request.
- `test_cross_mount_matrix.py` now treats gdrive as writable. It was marked read-only only because the old mock did not fake writes, so `cp` into gdrive hit the real API and failed.
- `read.py` still imports the gdocs, gsheets and gslides renderers directly. Those backends need their own seam. Out of scope here.
- This deviates from #684, which proposed adding more patch targets. That keeps the fragility, so this PR builds the seam instead.

## Verification

- `tests/core/gdrive/test_api.py` and `core/gdrive/api.test.ts`: fail on any direct Drive import in a gdrive module. Each also plants a violation and checks its own detector catches it, so the guard is proven able to fail.
- `tests/e2e/gdrive_mock.py`: rewritten to install through the one seam, so every existing gdrive e2e test now runs through the same door as production code.
- `tests/e2e/test_cross_mount_matrix.py`: gdrive moved to the writable set, so cp, mv, cp -r and mv -r across gdrive run for real.
- TypeScript gdrive tests: 15 `vi.mock` blocks replaced by `makeGDriveAccessor(fake)`, for the same reason.

Closes #684

Part of #970 (item 6).

### DIFF
--- a/python/mirage/accessor/gdrive.py
+++ b/python/mirage/accessor/gdrive.py
@@ -13,6 +13,7 @@
 # ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
 
 from mirage.accessor.base import Accessor
+from mirage.core.gdrive.api import DriveApi, drive_api
 from mirage.core.google.client import TokenManager
 
 
@@ -26,3 +27,16 @@ class GDriveAccessor(Accessor):
     def __init__(self, config, token_manager: TokenManager) -> None:
         self.config = config
         self.token_manager = token_manager
+
+    @property
+    def drive(self) -> DriveApi:
+        """The single door every gdrive core op reaches Drive through.
+
+        Built per read rather than memoized in ``__init__``: an embedding
+        program constructs the resource (and with it this accessor) before
+        a test installs a fake, and a door built once at construction time
+        would already point at the live API by then. Building one is a
+        frozen dataclass holding the token manager, so it costs nothing
+        beside the request it is about to make.
+        """
+        return drive_api(self.token_manager)
--- a/python/mirage/core/gdrive/api.py
+++ b/python/mirage/core/gdrive/api.py
@@ -1,0 +1,237 @@
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+
+from dataclasses import dataclass
+from typing import Any, Protocol
+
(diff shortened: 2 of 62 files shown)


sonhmai:
@codex review

sonhmai:
@codex review

PR #990, opened 2026-09-02
feat(utils): add Clock interface and inject it into cache TTL, FUSE prefetch and OpTimer

sonhmai:
RAM file-cache TTL, FUSE prefetch expiry and op timing each read the system clock directly. A test that wants to sit on an expiry boundary has to sleep for real or monkeypatch `time` process-wide, and an embedding program cannot supply its own time. This adds one injectable clock and threads it to those three readers.

## What changes

- New `Clock` (`now()` for wall time, `monotonic()` for durations, both in seconds) with a `SystemClock` default and a `ManualClock` that stays fixed until `advance()` moves it. Python `mirage/utils/clock.py`, TypeScript `core/src/utils/clock.ts`.
- `Workspace` takes an optional `clock` and hands it to the RAM file cache, the op facade and, through the facade, a kernel mount's prefetch cache. Copies keep it; `load()` and `from_state()` accept one so restored cache entries age on the caller's clock.
- `CacheEntry.is_expired(now)` replaces the `expired` getter that read the system time itself. Entries stay plain data; the store passes the reading in.
- FUSE prefetch deadlines and op durations read `monotonic()`. RAM-cache stamps stay whole-second wall time because they survive a snapshot round trip.

Deployments that never pass a clock behave exactly as before.

```python
clock = ManualClock(start=1000)
ws = Workspace({"/data": RAMResource()}, clock=clock)
await ws.cache.set("/data/file.txt", b"hello", ttl=10)
clock.advance(9)
assert await ws.cache.exists("/data/file.txt")
clock.advance(1)
assert not await ws.cache.exists("/data/file.txt")
```

## Out of scope

Index cache TTLs, retry sleeps, backend-internal `start_op()` timers, the command-recording scope and `Observer` event timestamps, and ID generation are separate tracker items (#970 item 2b and following). Only the three readers above move in this PR, so a workspace on a `ManualClock` still stamps shell commands from the system clock until 2b lands. Redis TTLs stay server-side.

## Verification

- `tests/utils/test_clock.py`, `utils/clock.test.ts`: `ManualClock` does not move when read, `advance()` moves both readings together and refuses to go backwards.
- `tests/cache/file/test_ram.py`, `cache/file/ram.test.ts`: TTL boundary probed at `ttl-1` and `ttl` with no sleep; expired entries are dropped on read and replaced by `add`.
- `tests/fuse/test_core.py`, `node/fuse/core.test.ts`: prefetch expires on the monotonic boundary and survives a wall-clock jump, which is the bug a `Date.now()` deadline would have.
- `tests/observe/test_context.py`, `observe/context.test.ts`: `OpTimer` durations and record timestamps come from the injected clock.
- `tests/workspace/snapshot/test_state.py`, `workspace/snapshot.test.ts`: a restored entry ages on the restoring workspace's clock; a copy keeps its origin's clock.
- `tests/workspace/workspace/test_cache.py`, `workspace/cache.test.ts`: the workspace hands one clock to the cache and the op facade.

Part of #970 (item 2).

### DIFF
--- a/python/mirage/cache/file/entry.py
+++ b/python/mirage/cache/file/entry.py
@@ -12,8 +12,6 @@
 # limitations under the License.
 # ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
 
-import time
-
 from pydantic import BaseModel
 
 
@@ -23,8 +21,12 @@ class CacheEntry(BaseModel):
     fingerprint: str | None = None
     ttl: int | None = None
 
-    @property
-    def expired(self) -> bool:
+    def is_expired(self, now: int) -> bool:
+        """Check expiry using the RAMFileCacheStore's clock reading.
+
+        Args:
+            now (int): whole seconds since the unix epoch.
+        """
         if self.ttl is None:
             return False
-        return (int(time.time()) - self.cached_at) >= self.ttl
+        return (now - self.cached_at) >= self.ttl
--- a/python/mirage/utils/clock.py
+++ b/python/mirage/utils/clock.py
@@ -1,0 +1,81 @@
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
+
+import time
+from typing import Protocol
+
+
+class Clock(Protocol):
+    """Wall timestamps and monotonic durations, both in seconds."""
+
+    def now(self) -> float:
+        """Seconds since the unix epoch (``time.time``)."""
+        ...
+
+    def monotonic(self) -> float:
+        """Seconds from an arbitrary origin that never moves backwards."""
+        ...
+
+
+class SystemClock:
+    """Platform clock used when no clock is supplied."""
+
+    def now(self) -> float:
+        """Seconds since the unix epoch."""
+        return time.time()
+
+    def monotonic(self) -> float:
+        """Seconds from an arbitrary origin that never moves backwards."""
+        return time.monotonic()
+
+    def __repr__(self) -> str:
+        return "SystemClock()"
+
+
+class ManualClock:
+    """Clock advanced explicitly by the caller. Readings have no side effects.
+
+    Args:
+        start (float): initial wall time in seconds; monotonic starts at zero.
+    """
+
+    __slots__ = ("_monotonic", "_wall")
+
+    def __init__(self, start: float = 0.0) -> None:
+        self._wall = start
+        self._monotonic = 0.0
+
+    def now(self) -> float:
+        """Seconds on the virtual wall clock."""
+        return self._wall
+
+    def monotonic(self) -> float:
+        """Seconds on the virtual monotonic clock."""
+        return self._monotonic
+
+    def advance(self, seconds: float) -> None:
+        """Move the clock forward by ``seconds``.
+
+        Args:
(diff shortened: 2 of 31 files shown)


sonhmai:
@codex review

sonhmai:
@codex review

typescript/packages/node/src/fuse/core.ts
```diff
@@ -211,12 +221,18 @@ export class MountCore {
     return entry
   }
 
+  // The prefetch cache keeps its deadlines in milliseconds, so every
+  // reading of the mount's clock is expressed in the same unit.
+  private nowMs(): number {
+    return this.clock.now() * 1000
```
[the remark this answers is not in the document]
sonhmai: Fixed. Prefetch deadlines now read `this.clock.monotonic() * 1000`, matching the Python core.

Added a regression test on both sides: a clock whose wall reading jumps ten years while monotonic stands still. The prefetched entry stays cached.

typescript/packages/core/src/cache/file/ram.ts
```diff
@@ -14,27 +14,56 @@
...
+  ) {
     super()
     this.limit = parseLimit(options.limit ?? '512MB')
+    this.clockRef = options.clock ?? new SystemClock()
     this.maxDrainBytes = options.maxDrainBytes ?? null
   }
 
+  // The current whole second, as `cachedAt` records it.
+  private seconds(): number {
+    return Math.floor(this.clockRef.now())
```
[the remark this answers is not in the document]
sonhmai: Not changing this here. The reason is the entry's anchor.

`cachedAt` is written into a snapshot and read back in a different process. A monotonic reading has an arbitrary per-process origin, so it cannot survive that trip. Wall clock is the only reading that still means something once the entry outlives the process that cached it.

Doing this properly needs a hybrid: keep `cachedAt` on wall for the snapshot, hold an in-memory monotonic deadline for live entries, and reconcile restored entries from wall at load. That is a behaviour change with a new reconciliation path, not part of making time injectable, so it belongs in its own change.

The prefetch cache was different and was fixed. It is in-process only and never persisted, so monotonic is simply correct there.

The wall clock step you describe is real, but it predates this PR and is unchanged by it. `ManualClock` moves both readings together, so an injected clock cannot drift the two apart.

python/mirage/utils/clock.py
```diff
@@ -0,0 +1,119 @@
...
+    Two readings, because a deadline and a stamp are different
+    questions. ``monotonic`` measures a duration or a TTL and never
+    moves backwards, which wall clock cannot promise, since NTP can
+    step it mid-measurement. ``now`` answers what time it is, which is
+    the only thing a stored timestamp can mean.
+
+    Implemented by :class:`SystemClock` in production and
+    :class:`ManualClock` under a test or an embedding harness. It is a
+    ``Protocol``, so an embedder's own class satisfies it by shape and
+    needs no import of mirage.
```
sonhmai: simplify comments

python/mirage/cache/file/entry.py
```diff
@@ -25,6 +25,25 @@ class CacheEntry(BaseModel):
...
     @property
     def expired(self) -> bool:
+        """System-time expiry retained for the public v0.0.6 API.
+
+        Stores use ``is_expired`` with their own clock instead.
+        """
+        return self.is_expired(int(SystemClock().now()))
+
+    def is_expired(self, now: int) -> bool:
+        """Whether this entry's TTL has run out at ``now``.
```
sonhmai: who is gonna call and pass this in now?

python/mirage/cache/file/entry.py
```diff
@@ -25,6 +25,25 @@ class CacheEntry(BaseModel):
 
     @property
     def expired(self) -> bool:
```
sonhmai: remove this. no api uses this

CHANGELOG.md
sonhmai, line 1: remove

sonhmai:
@codex review

python/mirage/workspace/workspace/workspace.py
```diff
@@ -289,7 +298,8 @@ def __init__(
                         agent_id=agent_id or "",
                         session_id=session_id,
                         links=self._namespace,
-                        dispatch=self._dispatcher.dispatch)
+                        dispatch=self._dispatcher.dispatch,
+                        clock=self._clock)
```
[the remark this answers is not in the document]
sonhmai: will be addressed separately in #970 2b

PR #1004, opened 2026-09-05
Preserve Redis index entry metadata

sonhmai:
## Summary

- Before: TypeScript Redis index serialization omitted `IndexEntry.extra`, so metadata present on a cold entry disappeared after a Redis-backed warm lookup.
- After: Redis `put` and `setDir` persist `extra`, and `get` restores it. Existing records without the field remain compatible and default to an empty object.
- Scope is limited to index entry serialization; TTL, child cleanup, empty-directory representation, and cache identity are unchanged.

## Verification

- Reproduced on current `origin/main`: nested `extra` round-tripped as `{}`.
- Focused Redis suite: 12 tests passed using an isolated local Redis instance and unique per-test key prefixes.


sonhmai:
@codex review

zechengz (approved):

PR #1005, opened 2026-09-05
Fix zero-length Redis range reads

sonhmai:
## How error happens

1. A client asks Mirage to read a byte range from a Redis-mounted file, for example `read(path, offset=0, size=0)`.
2. A zero-byte request should return immediately with an empty payload. It must not transfer file content.
3. Before this change, a non-empty file could return its complete contents when the request started at offset zero. The same request at another offset or through ordinary in-memory slicing returned empty, so behavior depended on the execution path.
4. After this change, zero-byte reads return an empty payload for every offset and for both empty and non-empty files. A missing file still surfaces as ENOENT.

## System-level view

```text
Client / command / runtime
          |
          | read(path, offset, size)
          v
Redis resource read
          |
          | get_file_range / getFileRange
          v
+-----------------------------+
| Before                      |
| end = offset + size - 1     |
| offset=0, size=0 => 0..-1   |
+-----------------------------+
          |
          | GETRANGE key 0 -1
          v
Redis interprets -1 as the last byte
          |
          v
Entire file returned unexpectedly
```

The store cannot infer existence from `GETRANGE`: Redis returns the same empty value for a missing key, an empty file, and a range beyond EOF. That is why range reads pair content retrieval with `EXISTS`.

```text
After

get_file_range / getFileRange
          |
          +-- size == 0 --> EXISTS key
          |                    |
          |                    +-- exists  --> empty bytes
          |                    +-- missing --> None/null --> ENOENT
          |
          +-- size > 0 or unbounded
                               |
                               +-- existing GETRANGE path, unchanged
```

Python, Node, and browser stores implement the same branch. Redis resources declare `caches_reads` / `cachesReads` as false, so there is no Redis-resource warm-cache branch to redesign; this PR stays limited to server-side range semantics.

## Before / after contract

| Case | Before | After |
| --- | --- | --- |
| Existing non-empty file, offset 0, size 0 | Entire file | Empty bytes |
| Existing non-empty file, positive offset, size 0 | Empty bytes | Empty bytes |
| Existing empty file, size 0 | Empty bytes | Empty bytes |
| Missing file, size 0 | Missing | Missing / ENOENT |
| Positive-size or unbounded range | Normal range | Unchanged |

## Implementation

- Special-case `size=0` before constructing Redis's inclusive range.
- Issue only `EXISTS` for that case.
- Return empty bytes when the key exists; retain `None`/`null` when it does not.
- Apply identical semantics in Python, Node, and browser/Upstash stores.

## Verification

- `python/tests/core/redis/test_read.py`: covers the Python resource path for zero-byte reads at zero and positive offsets, including empty and missing files.
- `typescript/packages/node/src/resource/redis/store.test.ts`: covers the same range matrix against Redis from the Node store.
- `typescript/packages/browser/src/resource/redis/store.test.ts`: keeps browser/Upstash store semantics in parity without live credentials.
- `typescript/packages/browser/src/resource/redis/redis_live.test.ts`: verifies the zero-byte and missing-key behavior against live Upstash when credentials are available.

## CI status / known blocker

The changed Redis suites pass in CI. Current failures are unrelated existing policy/approval work: Python mypy at `python/mirage/policy/decisions.py:344`, all-files YAPF at `python/tests/policy/test_decisions.py`, and TypeScript approval behavior at `typescript/packages/dsh/src/approval.test.ts:185`. This PR does not alter those paths and should be rebased after their fixes land.

### DIFF
--- a/python/mirage/resource/redis/store.py
+++ b/python/mirage/resource/redis/store.py
@@ -124,6 +124,9 @@ class RedisStore:
             size (int | None): how many bytes, or None for the rest.
         """
         key = self._fk(path)
+        if size == 0:
+            # GETRANGE 0 -1 means the whole value, not an empty window.
+            return b"" if await self._client.exists(key) else None
         end = -1 if size is None else offset + size - 1
         async with self._client.pipeline(transaction=False) as pipe:
             pipe.exists(key)
--- a/python/tests/core/redis/test_read.py
+++ b/python/tests/core/redis/test_read.py
@@ -13,6 +13,7 @@
 # ========= Copyright 2026 @ Strukto.AI All Rights Reserved. =========
 
 import os
+import uuid
 
 import pytest
 import pytest_asyncio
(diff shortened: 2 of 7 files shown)


sonhmai:
@codex review

zechengz (approved):
Thanks!

PR #1007, opened 2026-09-05
fix: evict stale cache entries when unmounting

sonhmai:
## Summary

- evict the exact mount cache key and its slash-delimited descendants before detaching the mount
- preserve slash-adjacent cache entries while keeping nested mounts routable and shared resources open
- leave the mount attached when cache eviction fails, so unmount can be retried

## Before / after

Before, reading `/m/x` through a caching mount could leave `child-secret` in the workspace file cache. After unmounting `/m`, the same path fell through to the root resource but still returned the removed mount's cached bytes.

After, unmount evicts `/m` and `/m/*` before changing routing. A later `/m/x` read reaches the current owner, `/more/*` remains cached, and a surviving nested mount refills through its own route.

## Verification

- Python focused cache-manager and unmount tests: 10 passed
- TypeScript focused cache-manager and unmount tests: 10 passed
- TypeScript strict typecheck: passed
- targeted formatting, Ruff, Prettier, ESLint, and Knip: passed
- structured local review: no actionable findings

## Current base blockers

- Python and pre-commit mypy report the existing `python/mirage/policy/decisions.py:344` mismatch: `Sequence[str]` is passed where `tuple[str, ...]` is required.
- The TypeScript full suite reports an unrelated failure at `typescript/packages/dsh/src/approval.test.ts:185`, where the retry records two prompts instead of one.
- The full-tree pre-commit run also reports that YAPF would modify an unrelated file; all Python files changed here pass the targeted YAPF hook.

This branch deliberately does not modify or bypass those checks. It should be rebased after the separate maintainer fixes land.

## Scope

Concurrent operations that resolved the old mount before eviction can still fill the cache late. Preventing that race requires generation or cancellation coordination and is intentionally separate from this focused route-removal fix. Direct-op rename and rmdir invalidation are unchanged.

### DIFF
--- a/python/mirage/cache/manager.py
+++ b/python/mirage/cache/manager.py
@@ -180,7 +180,7 @@ class CacheManager:
             await self._evict_dir(parent or "/")
 
     async def drop_prefix(self) -> None:
-        """Drop every cached body under this mount, path unspecified.
+        """Drop the mount key and every cached body beneath it.
 
         For a mutation that names no path: an account CLI writes to its
         service by id, so nothing here can say which file changed, only
@@ -194,6 +194,7 @@ class CacheManager:
         """
         if not self._caches_reads or self._file_cache is None:
             return
+        await self._file_cache.remove(self._prefix or "/")
         await self._file_cache.evict_prefix(self._prefix + "/")
 
     async def _invalidate_parent(self, key: str) -> None:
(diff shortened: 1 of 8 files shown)


sonhmai:
@codex review

zechengz:
Maybe related to the https://github.com/strukto-ai/mirage/pull/1025? not sure though

sonhmai:
> Maybe related to the #1025? not sure though

Thanks @zechengz. Yes, this is covered by #1006 and #1025. Closing this now.

Issue #1010, opened 2026-09-05
Redis cache stores: redundant round trips per command


sonhmai:
Collected from a `redis-cli MONITOR` session with both the file cache and the index cache on Redis (minio mount at `/s3`). None is a correctness bug; together they are several extra RTTs per command, and one of them scales with directory size.

1. **`EXISTS` then `GET` on the file cache.** Every read does `EXISTS cache:data:<p>` and then `GET cache:data:<p>`. A single `GET` (`None` = miss) is enough.

2. **`EXISTS cache:data:<dir>` for directory operands.** After `ls -l /s3/`, the executor probes `cache:data:/s3/reports`. The index already says it is a folder, and the file cache never holds a directory. Skip paths the index reports as folders (`workspace/provision/command.py::_check_cache_hits`).

3. **`ls -l` issues one `GET index:entry:<child>` per row, sequentially.** Two files is two RTTs; a 1000-object prefix is 1000. `FileCacheMixin` already has `multi_get`; the index store needs the same (`MGET`) and the stat fast path in `ls -l` should use it.

4. **Duplicate `GET index:entry:<operand>` before each command.** Every trace shows the operand's entry read twice back to back (route classification, then the stat check) within one invocation. A per-invocation memo removes one.

Related: #1008 covers the `EXISTS`+`TTL`+`LRANGE` triple in `list_dir`, which is fixed together with the empty-listing bug there.

PR #1040, opened 2026-09-09
fix(cache): recover evicted metadata and verify live freshness

### LINKED ISSUE #1011: [Issue]: ALWAYS consistency is skipped when the index cache holds a stat entry

Bug

### Area

Python package, TypeScript packages

### Summary

Under `ConsistencyPolicy.ALWAYS`, a read is supposed to re-stat the backend and serve the cached bytes only when the fingerprint still matches. That check is skipped whenever the index cache holds an entry for the path.

Chain:

1. `Reconciler._probe` (`workspace/reconcile.py`) calls `mount.execute_op("stat", path)` and compares `remote_stat.fingerprint` with the file cache.
2. The object-store `stat` (`core/object_store/stat.py`) has a fast path: if the index has an entry (populated by any earlier `readdir`), it builds the `FileStat` from the entry and returns. That entry carries no fingerprint (there is a `TODO` at the return site).
3. `_probe` sees `fingerprint is None`, returns `Verdict.UNKNOWN`, and the caller treats UNKNOWN as "keep serving".

So after one `ls` in the directory, an out-of-band change to the object is invisible to `cat` for as long as the file-cache entry lives, which under ALWAYS is exactly the case the policy exists to catch. Both the RAM and the Redis index store show it, since the fast path is above the store. TypeScript has the same fast path without a fingerprint (`core/object_store/stat.ts`).

### Steps to reproduce

minio in docker (`quay.io/minio/minio`, bucket `mirage-demo`, `minioadmin`/`minioadmin`), Redis or RAM file cache, `consistency=ConsistencyPolicy.ALWAYS`.

```python
ws = Workspace({"/s3": MinIOResource(cfg)}, mode=MountMode.WRITE,
               consistency=ConsistencyPolicy.ALWAYS)
await ws.execute("echo v1 | tee /s3/o.txt")
await ws.execute("ls -l /s3/")          # populates the index entry for /s3/o.txt
await ws.execute("cat /s3/o.txt")       # v1, fills the file cache
# out of band: docker exec mirage-minio sh -c "printf 'v2\n' | mc pipe l/mirage-demo/o.txt"
await ws.execute("cat /s3/o.txt")       # prints v1
```

Drop the `ls -l` line and the second `cat` prints `v2`, because `stat` then goes to the backend (`HEAD`) and returns a fingerprint.

### Expected behavior

Under ALWAYS the second `cat` prints `v2` regardless of whether the directory was listed before.

### Actual behavior

```
always prime_index=False  cat2=v2 fresh
always prime_index=True   cat2=v1 STALE      (RAM index)
always prime_index=True   cat2=v1 STALE      (Redis index)
```

`redis-cli MONITOR` for the stale case shows `GET idx:entry:/s3/o.txt` followed by `GET cache:data:/s3/o.txt`, and no `GET cache:meta:...` (the fingerprint is never compared) and no request to minio.

### Mirage version or commit

main @ 67389cca5

### Additional context

Two ways to close it, not exclusive:

- Carry the fingerprint in the index. `ListObjectsV2` already returns `ETag` per key; `readdir` can store it in `IndexEntry.extra` and the fast path can return it. That also removes the `TODO`.
- Make the probe bypass the index: a stat issued for reconciliation must reach the backend, since answering it from a cache is circular.

Either way, a test that lists, reads, mutates out of band, and reads again under ALWAYS should be added for both languages; nothing pins this today.

sonhmai:
Closes #1011.

- Fix `ALWAYS` reads returning old content after a directory listing.
- Recover when a directory listing survives but a child's cached metadata is evicted.
- Apply both fixes in Python and TypeScript.
- Fully resolve #1011; address partial eviction from #1010. Broader cache performance work stays in #1010.

## Why #1011 happens

- **Index cache:** directory listings and file metadata.
- **File cache:** file contents and their fingerprints.
- **Bug:** validate cached content using cached metadata; accept an unknown fingerprint as fresh enough.
- Example: `/s3/o.txt`, contents `v1`, then an external overwrite to `v2`.
- `E1` and `E2` below stand for the two versions' backend fingerprints.

```text
SETUP
  Backend: o.txt = "v1", fingerprint E1
       |
       +-- ls /s3/ --------> index caches o.txt metadata
       |                    (no fingerprint)
       +-- cat /s3/o.txt --> file cache stores "v1", E1
       |
  External write: o.txt = "v2", fingerprint E2

CURRENT
  cat /s3/o.txt [ALWAYS]
       |
       v
  stat --> cached index --> no fingerprint --> UNKNOWN
                                                |
                                                v
                                      return cached "v1"  BUG
  Backend E2 is never checked.

AFTER THIS PR
  cat /s3/o.txt [ALWAYS]
       |
       v
  live backend stat --> E2
       |
       v
  compare with cached E1 --> different
       |
       v
  discard cached bytes + stale index metadata
       |
       v
  backend read --> "v2"
```

- Matching fingerprint: reuse cached bytes.
- Deleted object: report missing; do not return old bytes.
- Missing fingerprint: discard cached content and read the backend.
- Failed probe: shell reads retry through a fresh backend read; direct read checks propagate the error.
- Fresh, temporary probe index: preserve backend file-ID resolution without consulting stale metadata.
- Clear stale IDs too: `path -> old blob ID -> old bytes` must not survive a rejected freshness check.

## Partial metadata eviction

- Directory membership and child metadata are separate cache entries.
- Losing one child's metadata does not prove the child was deleted.
- Example: backend still contains `/s3/report.csv` with contents `sales,42`.

```text
1. List /s3/

   Backend                        Index cache
   report.csv = "sales,42"  --->  /s3/ children = [report.csv]
                                  /s3/report.csv metadata = FILE

2. Evict only the child's metadata

   /s3/ children = [report.csv]    still cached
   /s3/report.csv metadata = ?     evicted
   Backend report.csv             still exists

3. stat /s3/report.csv

   CURRENT
     metadata missing + parent listing cached
       --> treat child as absent                    BUG

   AFTER
     metadata missing --> parent lists report.csv
       --> backend stat --> FILE
       --> subsequent read returns "sales,42"
```

- Preserve negative lookups when a valid cached listing excludes the child:

```text
stat /s3/missing.csv
  --> child metadata missing
  --> cached /s3/ children = [report.csv]
  --> missing.csv is NOT listed
  --> report missing; no backend request
```

## Snapshot drift and revision pins

- **Unpinned snapshot:** compare the recorded fingerprint against live backend metadata.
- **Fix:** a warm or restored index must not hide a changed backend fingerprint.
- Example under strict drift policy:

```text
Read /s3/data.csv = "version 1 bytes" [E1]
  --> save snapshot with fingerprint E1
  --> external overwrite = "VERSION 2 DRIFTED" [E2]
  --> load snapshot; index still has old file metadata
  --> next read triggers live backend stat --> E2
  --> compare recorded E1 with live E2
  --> ContentDriftError; do not silently accept changed content
```

- **Revision-pinned snapshot:** keep reading the recorded version; behavior unchanged.
- Example with S3 versioning enabled (`V1`/`V2` stand for VersionIds):

```text
Read /s3/data.csv = "original" [VersionId V1]
  --> save snapshot with revision V1
  --> external overwrite = "mutated bytes" [VersionId V2]
  --> load snapshot; pin data.csv to V1
  --> backend read requests VersionId V1
  --> return "original", even though latest is V2
```

## Validation

- Focused tests: **115 Python + 158 TypeScript passed**.
- Coverage: RAM/Redis indexes; shell/filesystem reads; external overwrite/delete; partial eviction; unknown/failed probes; GitHub file IDs; snapshot drift and revision pins.
- Affected package typechecks and applicable pre-commit hooks passed.
- Local full TypeScript run: unrelated native jq 1.6/BSD sed conformance failures. Remote TypeScript workflow passed.
- [CI](https://github.com/strukto-ai/mirage/pull/1040/checks): **45 passed, 5 skipped**.
- Local and PR reviews: no actionable findings.

### DIFF
--- a/docs/home/cache.mdx
+++ b/docs/home/cache.mdx
@@ -11,6 +11,10 @@ Every `Workspace` ships with a **two-layer cache** so repeated work against remo
 - **Index cache.** Listings and metadata. The first directory walk hits the API; subsequent ones serve from the index until the TTL expires.
 - **File cache.** Object bytes. The first read streams from origin; later pipelines read from cache.
 
+Under `ConsistencyPolicy.ALWAYS`, cached file bytes are checked against live backend metadata, bypassing the index. A changed fingerprint triggers a fresh read; a deleted file is evicted. If the backend cannot supply a fingerprint or its probe fails, the read must fetch current bytes or report an error. `LAZY` continues to trust cached bytes.
+
+A fresh directory listing proves a child absent only when that child is not listed. If a listed child's metadata has been evicted independently, object-store stat fetches it from the backend.
+
 ## Stores
 
 Each layer is a pluggable store with two built-ins:
--- a/python/mirage/core/object_store/stat.py
+++ b/python/mirage/core/object_store/stat.py
@@ -77,11 +77,12 @@ def make_stat(driver: ObjectStoreDriver[A, C]) -> StatFn[A]:
         # probe speculatively (e.g. .git, HEAD, .hg during cd).
         parent = virtual_key.rsplit("/", 1)[0] or "/"
         parent_listing = await index.list_dir(parent)
-        if parent_listing.entries is not None:
+        if (parent_listing.entries is not None
+                and virtual_key not in parent_listing.entries):
             raise enoent(virtual)
 
-        # Slow path: no index cache available, or parent directory not
-        # yet listed. Hit the store.
+        # A listed child can lose its metadata to independent eviction.
+        # Missing cache information must fall back to the store.
         kpfx = driver.key_prefix_of(accessor)
         key = kp.apply(kpfx, path)
         async with driver.connect(accessor) as conn:
--- a/python/mirage/workspace/reconcile.py
+++ b/python/mirage/workspace/reconcile.py
@@ -16,6 +16,7 @@ import logging
 from enum import Enum
 
 from mirage.cache.file.mixin import FileCacheMixin
+from mirage.cache.index.ram import RAMIndexCacheStore
 from mirage.types import ConsistencyPolicy
 from mirage.workspace.mount.mount import MountEntry
 from mirage.workspace.mount.namespace import Namespace
@@ -65,15 +66,22 @@ class Reconciler:
             mount (MountEntry): the resolved mount for ``path``.
             path (str): absolute virtual path to probe.
         """
+        # Resolve backend IDs without reusing cached metadata.
         try:
-            remote_stat = await mount.execute_op("stat", path)
+            remote_stat = await mount.execute_op("stat",
+                                                 path,
+                                                 index=RAMIndexCacheStore())
         except FileNotFoundError:
             await self.on_missing(path)
+            await mount.index.clear()
             return Verdict.GONE
         if remote_stat is None or remote_stat.fingerprint is None:
+            await self._cache.remove(path)
+            await mount.index.clear()
             return Verdict.UNKNOWN
         if not await self._cache.is_fresh(path, remote_stat.fingerprint):
             await self._cache.remove(path)
+            await mount.index.clear()
             return Verdict.STALE
         return Verdict.FRESH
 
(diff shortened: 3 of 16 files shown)


sonhmai:
@codex review

zechengz (approved):
Thanks! :)

Issue #1016, opened 2026-09-05
[Issue]: Workspace.snapshot() raises AttributeError when the file cache is Redis


sonhmai:
### What type of report is this?

Bug

### Area

Python package

### Summary

`Workspace.snapshot()` raises when the workspace file cache is `RedisCacheConfig`. `to_state_dict` (`python/mirage/workspace/snapshot/state.py:159-168`) reads `cache._entries` and `cache._store.files` unconditionally; both exist only on `RAMFileCacheStore`.

TypeScript guards this: `workspace/snapshot/state.ts:98` uses `ws.cache instanceof RAMFileCacheStore ? ... : []`. The Python restore side already has the same guard (`_restore_cache`, `state.py:373-376`); only the capture side is missing it.

### Steps to reproduce

```python
ws = Workspace({"/r": RAMResource()}, mode=MountMode.WRITE,
               cache=RedisCacheConfig(url="redis://localhost:6379/8", key_prefix="c:"))
await ws.execute("echo hi | tee /r/a.txt")
await ws.snapshot(io.BytesIO())
```

### Expected behavior

A snapshot with no cache entries, as TypeScript produces.

### Actual behavior

```
AttributeError: 'RedisFileCacheStore' has no attribute '_entries'
```

### Mirage version or commit

main @ 67389cca5

bytecii:
Fixed in #1041 (merged as 9dd342862). `to_state_dict` now captures cache entries only for a RAM file cache, matching the TypeScript guard and the Python restore side; under `RedisCacheConfig` a snapshot carries no cache entries instead of raising. Pinned by the redis-file-cache round added to `integ/cross.sh` (`cross_cache.yaml`), which snapshots and loads in both directions.

Issue #1015, opened 2026-09-05
[Issue]: for-loop variable, pipeline assignment and $( ) assignment scope differ from bash


sonhmai:
### What type of report is this?

Bug

### Area

Python package, TypeScript packages

### Summary

Three shell scoping rules differ from bash. All three verified against bash 5.2 in `debian:stable-slim`; all three identical in Python and TypeScript.

1. **`for X in a b; do :; done` restores or deletes `X` after the loop.** `workspace/executor/control.py:231-236` (and `:587-590` for `select`) saves the previous value in a `finally` and re-seeds or pops it. In bash the loop variable is an ordinary variable and keeps its last value. TS: `executor/control.ts:236,591`.
2. **An assignment in a pipeline element persists.** `P=0; P=1 | cat; echo $P` prints `1`. bash forks every pipeline element, so it prints `0`. `executor/pipes.py:37` (`handle_pipe`) and `:137` (`handle_connection`) run elements against the parent session; only `handle_subshell` (`:238`) snapshots.
3. **An assignment inside `$( )` persists.** `Y=0; Z=$(Y=1; echo y); echo $Y` prints `1`; bash prints `0`. `workspace/expand/node.py:353` runs the body with `execute_fn(inner, session_id=session.session_id)`. Same for backticks and TS `expand/node.ts`. This one is the widest: `cd`, `unset`, `set -o`, `shopt` and function definitions inside a substitution all leak into the caller.

A fourth, smaller one: a `pre_session` refusal inside `$( )` is silent. `echo $(GATE_X=1; echo y)` under a refusing policy exits 0 with empty stdout and empty stderr; bash prints the refusal and keeps running the body. The `PolicyDenied` aborts the substitution and its stderr is dropped in `expand_node`.

### Steps to reproduce

```bash
docker run --rm debian:stable-slim bash -c 'for X in a b; do :; done; echo "for=[$X]"; P=0; P=1 | cat; echo "pipe=[$P]"; Y=0; Z=$(Y=1; echo y); echo "subst=[$Y]"'
```

Then the same three lines through `ws.execute`.

### Expected behavior

```
for=[b]
pipe=[0]
subst=[0]
```

### Actual behavior

```
for=[]
pipe=[1]
subst=[1]
```

### Mirage version or commit

main @ 67389cca5

### Additional context

[...]

Issue #1014, opened 2026-09-05
[Issue]: /.bash_history ignores history -c and leaks other sessions; empty line is recorded


sonhmai:
### What type of report is this?

Bug

### Area

Python package, TypeScript packages

### Summary

CLAUDE.md says `/.bash_history` and the `history` builtin are two views over the same events and never disagree. They do disagree, because they call different Observer methods.

- `history` builtin: `Observer.session_command_events(session)` (`python/mirage/commands/builtin/history/history.py:64,89`). Scoped to the session, applies `CLEAR` and `DELETE` events.
- `/.bash_history` view: `Observer.command_events()` (`python/mirage/core/history/read.py:40`; TS `core/history/read.ts:30` calls `commandEvents()`). Every session, raw order, no `CLEAR`/`DELETE` applied.

Second, an empty typed line is recorded. bash 5.2 never adds a blank line to history; whitespace-only and comment-only lines are added. `execute_line` (`workspace/workspace/execute.py`) gates recording only on `record and not provision`, with no emptiness check. Same in `execute.ts`.

### Steps to reproduce

Two sessions `s1`, `s2` in one workspace, RAM observer store.

```
s1: echo one ; echo two ; history -c ; echo three
s2: echo other-session-cmd
s1: history ; cat /.bash_history
```

### Expected behavior

`history` and `cat /.bash_history` in `s1` show the same two lines (`history -c`, `echo three`). `ws.execute('')` records nothing.

### Actual behavior

```
history:        1  history -c
                2  echo three
/.bash_history: #..\necho one\n#..\necho two\n#..\nhistory -c\n#..\necho three\n#..\nhistory\n#..\necho other-session-cmd
```

`echo one`/`echo two` survive in the file after `history -c`, and `s2`'s command appears in `s1`'s file. `ws.execute('')` lands as an `EVENT_COMMAND` with `command=''`.

### Mirage version or commit

main @ 67389cca5

### Additional context

The docstring in `history.py` justifies `-a/-n/-r/-w` being no-ops with "both are the same store and always in sync"; that premise is what this issue breaks. Fix: thread the caller's session id to the history read path and call `session_command_events`, in both languages; skip logging when `command.strip() == ""`.

Also noticed, not a bug: `HISTCONTROL`, `HISTIGNORE`, `HISTSIZE`, `HISTFILESIZE` are not honored anywhere (`render.py` `HISTSIZE = 500` is a display cap only). Worth an explicit decision in the History section.

[...]

PR #1041, opened 2026-09-09
for/select keep the loop variable, blank lines unrecorded, Redis-cache snapshot, TS stat parity

### LINKED ISSUE #1021: [Issue]: TS stat line differs from Python: mergeOverlayStat drops content, S3 modified keeps .000Z

Bug

### Area

TypeScript packages

### Summary

Two places where a TypeScript `stat` line differs from the Python one for the same object.

1. **`mergeOverlayStat` drops `content`.** `typescript/packages/core/src/workspace/mount/namespace/overlay.ts:53-65` ends with a hand-written `new FileStat({...})` that copies eleven fields and omits `content`. Python's `merge_overlay_stat` uses `stat.model_copy(update=update)` and keeps everything. `FileStat.with()` exists for this and its docstring (`types.ts:543-549`) says hand-written lists are how fields go missing. Visible after any `touch` (which writes an mtime into the overlay): `stat` prints `type=file` where Python prints `type=text`.
2. **S3 `modified` keeps `.000Z`.** `core/s3/driver.ts:150` uses `resp.LastModified.toISOString()`; `HeadObject` returns whole seconds, so it renders `2026-09-05T10:55:39.000Z`. Python's `to_iso_z` renders `2026-09-05T10:54:04Z`. The list path (`isoOf`, `driver.ts:55`) has the same gap, plus a digit-count split: Python six fraction digits, TS three, for the same millisecond value, and that one lands in the persisted `remote_time`.

### Steps to reproduce

```
tee /s3/c.txt <<< x ; touch /s3/c.txt ; stat /s3/c.txt
tee /s3/new.txt <<< n ; stat /s3/new.txt
```

### Expected behavior

```
name=c.txt size=2 modified=... type=text
name=new.txt size=2 modified=2026-09-05T10:54:04Z type=text
```

### Actual behavior

```
name=c.txt size=2 modified=... type=file
name=new.txt size=2 modified=2026-09-05T10:55:39.000Z type=text
```

### Mirage version or commit

main @ 67389cca5

### Additional context

Fix for (1) is `return stat.with(update)`. For (2), add a `toIsoZ(d: Date)` beside `epochToIso` in `utils/dates.ts` and use it in `core/s3/driver.ts`; the other bare `toISOString()` sites (`core/s3/watch.ts:34`, `core/databricks_volume/readdir.ts:115`, `node/src/core/ssh/watch.ts:57`) have the same shape. Pick one fraction-digit policy for the list path in both languages.

bytecii:
Four small parity/bash fixes from the 2026-09-05 issue batch, each pinned in both languages.

## `for`/`select` keep the loop variable (#1015, item 1)

bash 5.2 (`debian:stable-slim`): `Z=before; for Z in a b; do :; done; echo $Z` prints `b`, and `unset Y; for Y in ; do :; done` leaves `Y` unset. Both handlers saved the previous value and put it back (or popped the name) in a `finally`; the restore is gone, and the three tests that pinned the old behavior (`test_control.py`, `control.test.ts`, the node S3 `pipeline.test.ts`) now assert bash's. Items 2 and 3 of #1015 (pipeline element and `$( )` scope) were already fixed on main by the per-segment snapshot in `handle_pipe` and `child_line`.

## A blank line is not recorded (#1014, item 2)

bash adds a line to history only when it is non-empty (`shell_input_line[0]`): blank lines are skipped, whitespace-only and comment-only lines are kept (pinned with `printf 'echo one\n\n   \n# comment\n' | bash -i; history -w`). `execute_line`/`runParsedLine` gate `log_execution` on that. The other two claims in #1014 are not taken: `bash_history_survives_clear` in `integ/bash/history/clear.json` deliberately pins the file as the cross-session, clear-surviving view, which is also what a shared `~/.bash_history` does in GNU bash (`history -c` clears the list, not the file).

## `Workspace.snapshot()` with a Redis file cache (#1016)

`to_state_dict` read `cache._entries` and `cache._store.files` unconditionally; both exist only on `RAMFileCacheStore`, so a Redis cache raised `AttributeError`. It now carries no entries for a non-RAM cache, as `_restore_cache` already assumed and as `toStateDict` does.

## TS `stat` line parity (#1021)

- `mergeOverlayStat` rebuilt the `FileStat` by hand and dropped `content`, so after a `touch` TS printed `type=file` where Python printed `type=text`. It now goes through `FileStat.with`.
- A new `toIsoZ(date)` beside `epochToIso` renders the shape Python's `to_iso_z` renders, and replaces bare `toISOString()` in the S3 driver/watch, databricks readdir/stat, ssh entry/watch (these use `epochToIso`, the twin of the Python `epoch_to_iso` those sites call) and the node gridfs driver. `to_iso_z` itself now has one fraction policy both languages can express: none when zero, three digits otherwise (`isoformat()` wrote six), and reads a naive datetime as UTC (pymongo's `uploadDate` is naive UTC; `astimezone` read it as local time). Python's databricks `modified_to_iso` rendered `+00:00` and now goes through `to_iso_z` too.

Refs #1015, #1014, #1016, #1021.

## Integration coverage

Every fix above is now also pinned by the cross-host battery (run on both hosts locally: ram/ram-history and the 15 remote-backed targets the cases name, 0 failures):

- `integ/bash/control/scope.json`: a `for`/`select` loop leaves its variable at the last value (`break`, an empty list, nested loops, `select` fed by a here-string), on all 22 shell targets.
- `integ/bash/history/blank.json`: a blank typed line is never recorded while a whitespace-only and a comment-only line are, pinned through the builtin's numbering and the `/.bash_history` view.
- `integ/unix/meta_overlay/stat.json::ometa_stat_type_survives_touch`: `stat` keeps `type=text` after a `touch` on every overlay target (the `mergeOverlayStat` field drop).
- `integ/unix/stat/c.json::stat_y_never_prints_zero_millis`: `%y` never renders `.000Z` on the seven backends whose drivers stamp the time themselves (case-target baseline 2226 -> 2227, reason recorded in `target_exceptions.json`).
- `integ/cross.sh`: a redis-file-cache round in both directions (`integ/cross_cache.yaml`), i.e. the `workspace snapshot` that used to raise on a non-RAM cache store, followed by a load in the other language.

## After upstream #1037

Upstream merged strukto-ai/mirage#1037 while this was open, which fixed #1021 on its own terms: `mergeOverlayStat` goes through `FileStat.with`, S3 and Databricks stat timestamps render through a `toIsoZ` that pads a non-zero fraction to six digits (the pre-existing Python `to_iso_z` spelling), and the S3/SSH watch checkpoints deliberately keep the `toISOString()` spelling they persisted. The merge here takes upstream's versions of all of that and drops this PR's three-digit fraction policy (both `toIsoZ` and the `to_iso_z` change are gone). What this PR still adds on the stat side: the gridfs rows and the ssh readdir/stat rows render through the same helpers (upstream did not touch them), and the two integ pins above, which hold under upstream's policy as well: neither language can print `.000Z` any more.


zechengz:
@codex review

PR #1059, opened 2026-09-12
fix: use Dropbox list_folder/continue behind watch pull

### LINKED ISSUE #1053: Watch: nothing feeds Watcher.notify(), and every hook is a listing walk
Two gaps that share one seam. Both are about detection cost, and the code is already shaped for both.

## Where we are

All 13 watch implementations are the same thing:

| backend | hook |
|---|---|
| box, dropbox, gdrive, github, gridfs, hf_buckets, hf_hub, nextcloud, onedrive, opendal, s3, sharepoint, ssh | `ListingDeltaHook` |

That is walk the tree, fingerprint every entry, diff the listing. On an API tree it is one request per directory per poll, and it scales with the size of the tree rather than with the number of changes.

Meanwhile `Watcher.notify()` exists and nothing calls it:

> The single entry point for all detection: a consumer's webhook receiver, queue bridge, or poll loop maps its signal to a `FileEvent` and calls this.

So the push ingress is built and unused.

## 1. Native delta pull, behind the existing contract

`DeltaHook.pull(root, checkpoint)` already carries an opaque `checkpoint`, which is precisely a vendor cursor. No contract change is needed, only a second implementation beside `ListingDeltaHook`, with the walk kept as the reset path.

| backend | native feed |
|---|---|
| gdrive | `changes.list` + pageToken (its own docstring asks for this) |
| onedrive, sharepoint | Graph `/delta` + deltaLink |
| dropbox | `files/list_folder/continue` + cursor |
| box | events API |
| github | events / `since` |
| gmail (new) | `history.list` + startHistoryId, see the gmail issue |

Shared wrinkle: several of these are account-wide feeds that have to be filtered back to the watch root. Worth solving once rather than per backend.

## 2. Push, feeding notify()

Most of these vendors push over an ordinary webhook, which is much lighter than it sounds because `notify()` already does the cache invalidation and fan-out. What is missing is the bridge and the subscription lifecycle (register, renew, expire).

| backend | mechanism |
|---|---|
| gdrive | `changes.watch` -> HTTPS webhook |
| dropbox | webhook |
| box | webhooks |
| onedrive, sharepoint | Graph subscriptions -> webhook |
| nextcloud | webhooks (the `Watcher` docstring already names this case) |
| github | webhooks |
| s3 | event notifications -> SQS/SNS |
| gmail | Pub/Sub only, no plain webhook, so the heaviest of the set |

Renewal is the recurring cost: Graph subscriptions and Gmail watches both expire and need re-registering.

## Suggested order

Native delta first. It needs no deployment surface, no public endpoint and no per-tenant vendor configuration, and it removes most of the polling cost on its own. Push is worth it after that for latency, and is most defensible where a webhook already exists rather than where a whole Pub/Sub topic has to be provisioned.

modelpath-dev:
## Summary
- Dropbox `delta_hook` now pulls through `list_folder/continue` instead of walking the whole tree every poll.
- The listing walk stays as the reset path when Dropbox invalidates the cursor (`path/reset`). A listing-era snapshot still upgrades.
- `Watcher.notify()` stays consumer-owned. This does not call it from `pull`.
- The integ Dropbox fake treats a finished listing cursor as a delta watermark so later writes and deletes show up on continue.

Fixes #1053 for Dropbox. Other vendor cursors and push/webhooks are follow-up PRs.

## Test plan
- [x] `uv run pytest tests/core/dropbox/test_watch.py tests/core/dropbox/test_api.py` (15 passed)
- [x] vitest `src/core/dropbox/watch.test.ts` and `api.test.ts` (13 passed)
- [ ] CI


zechengz:
Reviewed and pushed the CI fix plus the findings below. Head is now `65ff7ddc3`.

## Why CI was red

Only `pre-commit` was this PR: `integ/server/dropbox/routes.ts` was not prettier-formatted.

The other five (`integ-data`, `integ-ts`, `integ-runtime`, `Cross-language snapshot interop`, and the two aggregate gates) all died on `docker pull minio/minio` with `pull access denied`, exit 125. That is the Docker Hub teardown, fixed on main in `90907dcae` (pull from quay); this branch forked before it. Merged main in.

Worth knowing: `integ-data` and `integ-ts` both reported `2832 passed, 0 failed` and *then* hit the docker step. The dropbox **watch** battery runs after that step, so this PR's own integ coverage never actually ran. I ran it locally against the fake: 24 dropbox cases green.

## Findings fixed

**1. The reset path could write an unusable cursor and wedge the watch.** `_snapshot` answers a root that 409s with `cursor=""`. The baseline branch guards that (`if not cursor: return await self._listing.pull(...)`), the reset branch did not, so it encoded:

```
after reset + missing root: {"_dbx": 1, "c": "", "s": {}}
RAISED 400 cursor sent: ''
```

`ListFolderContinueArg.cursor` is `String(min_length=1)`, so every later pull raises and the hook never re-lists, even once the folder is back. Both branches now route through one `_relist`/`relist` that falls back to a listing-format checkpoint when there is no cursor to keep. Regression tests in both languages; each fails on the old code with exactly `{_dbx: 1, c: '', s: {}} != {}`.

**2. The fake ordered every write before every delete.** A name deleted and then moved onto appears in both tables, so `changesSince` rendered it create-then-delete. Probed:

```
== continue ==            == ground truth (list_folder) ==
file    /a.txt            file /a.txt
deleted /a.txt
deleted /b.txt
```

A client applying that page in order ends up with `/a.txt` gone. `changesSince` now merges both tables by `seq`, and `copyTree` clears the destination tombstone the way `putFile`/`addFolder` do.

**3. The delta watermark was read after the rows.** A kit read waits for queued writes but does not join the queue, so an upload can land between `listChildren` and `maxItemSeq`; taking the watermark last swallows that row permanently. It is now read first, so a concurrent write is re-reported and the client's fingerprint diff discards it.

**4. Smaller ones.** `deletedEntry` was inserted between `entryFor` and the `content_hash` comment that documents it. TS `decode` cast a non-object checkpoint where python's `_decode` guards with `isinstance`. TS wrapped the whole snapshot loop in the reset `try` where python wraps only `continue_folder`. `remove()` called `minter.next('item')` in both arms of each tombstone upsert.

## Not covered

- Finding 2 has no CI regression test. It needs two mutations inside one poll window and `integ/watch/run.py` runs one `mutate` per case, so covering it would mean changing the runner.
- `DropboxDeltaHook` on the TypeScript side still has unit tests only; the watch battery is python-only. Pre-existing, not from this PR.

## Verified

pre-commit on every touched file (16 hooks pass) · `pnpm -r build` + `pnpm -r typecheck` clean · vitest `src/core/dropbox/` 82 passed · full `uv run pytest` 20496 passed, exit 0 · watch battery exit 0, 24 dropbox cases green.

zechengz (approved):
Modified and thanks for the contribution

modelpath-dev:
@zechengz 

Hi Zecheng, thanks for the review and the CI fix on this one. I enjoyed working through the Dropbox watch-pull issue and getting the conformance side right. I am currently looking for an internship, and I would love to keep contributing to Mirage in the meantime. I noticed the open issues around integration coverage and the TS CLI work, and that is exactly  the kind of thing I like digging into. 

If you have any intern or contract openings on the team, I would be glad to share my background and talk.

PR #1079, opened 2026-09-14
ci(integ): tune the nextcloud container, 31m42s to 6m24s

Mkhoa161:
## What

Tunes the Nextcloud service container before the declarative battery runs against it, in the two jobs that declare it (`integ-ts`, `integ-data`). No product code, no test changes, no cases dropped.

Refs #886 — its last comment names `integ-data` as Integ's tail and lists two unproven threads for it (parallel target lanes, fewer WebDAV round trips). This is a third one that turned out to be the actual bottleneck, and it is YAML only.

## Why the battery was slow

Not the storage path. **WebDAV is stateless, so every request re-verifies the Basic credentials**, and Nextcloud 30 hashes with Argon2id at 64 MiB / 4 passes. Measured inside `nextcloud:30-apache`:

| probe | stock |
|---|---:|
| `GET /core/img/favicon.ico` (static) | 0.6 ms |
| `GET /status.php` (minimal PHP, unauthenticated) | 9.5 ms |
| `PROPFIND /remote.php/dav/…` (authenticated) | **196.6 ms** |
| `password_hash(…, PASSWORD_ARGON2ID)` in-container | **177.2 ms** |

The hash is ~90% of every authenticated request, and it taxes reads exactly as hard as writes. The battery issues **9,422 requests** for its 3,044 cases; 9,422 × 197 ms = 30.9 min, against the 30.4 min this step actually takes in CI. That accounts for essentially the whole step.

Lowering `hashingMemoryCost`/`hashingTimeCost`/`hashingThreads` takes an authenticated PROPFIND from 197 ms to 23 ms. The storage-side knobs (file locking, APCu, `files_versions`, `files_trashbin`, `activity`) are real but an order of magnitude smaller — they were simply invisible underneath the hashing cost.

Nine-operation WebDAV round (PUT create, 2 overwrites, PROPFIND ×2, GET, MOVE, DELETE, MKCOL), 40 iterations each, same live container, one knob set at a time:

| config | round | vs stock |
|---|---:|---:|
| stock image | 1,996 ms | — |
| storage knobs only (locking, APCu, 3 apps) | 1,846 ms | −7% |
| hashing only, apps left on | 406 ms | −80% |
| **both** | **253 ms** | **−87%** |

Per-op, the storage knobs do land where expected once the hashing tax is gone: DELETE 71 → 29 ms, MOVE 59 → 32, PUT overwrite 54 → 30. Both sets are kept.

## Measured

Full battery, same machine, same image, back to back:

| arm | wall | requests | result |
|---|---:|---:|---|
| stock container | **31 min 42 s** | 9,422 | 3044 passed, 0 failed |
| tuned container | **6 min 24 s** | 9,422 | 3044 passed, 0 failed |

**4.95×, −79.8%.** Both arms issued **exactly the same 9,422 requests**, which is the control that matters: the tuning changed the cost of a request, not the work done or the cases exercised.

Tuning itself costs **2.6 s** for eleven `occ` invocations, and re-running it is a no-op (exit 0).

#886 is explicit that laptop timings in this work have been unreliable, so: both arms ran back to back on the same machine, same image, same battery, with only the container config differing. The stock arm came in at 31.7 min against the 30.4 min this step takes on a CI runner — within 4%, which is the reason to trust the ratio here. The number that settles it is still this PR's own Integ run against main's.

## Expected effect on CI

| job | now | after |
|---|---:|---:|
| `integ-data` | 37.0 | 12–14 |
| `integ-ts` | 33.3 | 9–11 |
| `integ-runtime` (untouched, becomes the ceiling) | 31.5 | 31.5 |

Integ wall clock 37.3 → ~31.5 min, where `integ-runtime` becomes the ceiling. That is what #1078 is splitting, and the two are independent — different jobs, no shared lines — so they compose: with both in, the ceiling drops to `integ-shared-ts` at 28.3 min and the next candidates become the shared-battery jobs.

About 47 runner-minutes come back immediately; the rest of this change's saving is banked behind that ceiling.

## Why coverage is unaffected

- The battery passes **3044/3044** against a fully tuned container.
- Both accessors are thin `opendal` WebDAV wrappers. Greps across both languages for `LOCK|423|Lock-Token|trashbin|versions|chunked|uploads/|quota` return nothing but licence headers — no code path reads any disabled app, and nothing issues a LOCK or asserts a 423, so `filelocking.enabled=false` has no contract to break. The Files Search `SEARCH` that `find` pushes down comes from the core `dav` app, which stays enabled.
- Disabling file locking is safe because there is no concurrency to lose: the runner, the executor's per-mount fan-out (which matters here — nextcloud mounts three prefixes on one server) and the watch battery's writes are all sequential.
- **The watch battery later in `integ-data` is immune by construction, not by measurement.** It never registers a webhook with Nextcloud: `PushTrigger` synthesises the payload itself and POSTs it to a local aiohttp receiver, and `webhook_listeners` is not enabled by any workflow. Pull mode is one recursive PROPFIND fingerprinted on the WebDAV ETag; event mode is disk-only. I did not run that battery locally — it needs six fake servers — so this is an argument from reading the code, and it is the weakest link in the coverage claim.
- The APCu change is self-detecting: every watch case asserts the new content present **and** the old content absent, so a server-side cache serving stale state goes red rather than passing quietly.
- An app a future image stops shipping will not turn this red: `core/Command/App/Disable.php` prints `No such app enabled: <id>` and leaves the exit code at 0, reserving exit 2 for an exception while disabling. The flip side, noted in a comment: that exit code verifies nothing, so a typo would be swallowed and show up only as the battery running slowly.

### One sharp edge for whoever tunes this container next

`python/mirage/core/nextcloud/find.py:219-221` falls back to a client-side walk if `SEARCH` answers 404/405/501, producing identical output. That is the one way a container-side change can delete the SearchDAV push-down without anything going red, and `--strict` would not catch it (it only fails on target-level skips). None of the disabled apps provides SearchDAV, so it does not fire here.

## Shape

One composite action, because two jobs had already duplicated the wait-for-install loop verbatim and `integ-battery-setup` sets the precedent ("so the setup stays defined once"). `${{ job.services.nextcloud.id }}` is the handle — a `services:` container gets an auto-generated name, so `docker exec nextcloud` does not resolve — and an empty one is guarded rather than left to surface as `docker exec ""` several lines later.

`admin-password` is required rather than defaulted, and read at each call site from the job env the battery itself authenticates with, so the reset and the battery cannot disagree. The poll carries `--connect-timeout`/`--max-time`, since the 60×5s bound covers the sleeps but not a server that accepts and then stalls, and the caller carries `timeout-minutes` because a composite step cannot.

The new action is added to the workflow's push paths and to the `ts` and `data` change filters, so editing it alone still triggers the jobs that use it.

## Verification

`actionlint`, `yamllint` and `shellcheck` are clean on the changed files. The tune step was executed verbatim out of the committed `action.yml` against a real container: exit 0 in ~2.5 s, idempotent on a re-run, and loud on each failure mode — empty container id trips the guard, empty password is refused by `--password-from-env` with exit 1, and re-enabling `password_policy` makes the reset fail with exit 1, so the ordering the speedup depends on cannot silently regress.

Two things I could **not** verify locally, stated plainly: `act` starts the service container but does not implement the `job.services` context, so the id lookup is confirmed only against GitHub's documented context table; and neither lint gate reads `action.yml` *as an action*, so its expressions are unchecked by CI. Both first run for real on this PR.

⚠️ The three hashing knobs are CI-container-only and must never reach a deployment config — they make a stolen hash cheap to crack. They are safe here because the container is ephemeral, listens on localhost, and its password is already a literal in this workflow. Called out in a comment at the point of use.


zechengz (approved):
Thanks!

